Issue #2, opened 2026-04-30
Rename LLM adapter from "codex" to "chatgpt" — name collision with knowledge store capability


huangzesen:
### Problem

The kernel uses `"codex"` as the name for two unrelated subsystems:

| Subsystem | Module | Purpose |
|---|---|---|
| Codex knowledge store | `core/codex/__init__.py` | Persistent knowledge archive (tool: `"codex"`) |
| Codex LLM provider | `auth/codex.py` + `llm/_register.py:54-82` | OAuth2 token manager for ChatGPT backend (adapter: `"codex"`) |

No shared state, no runtime interaction. The collision is purely nominal but causes confusion in:
- **Conversation**: "use codex to…" — which one?
- **Error messages**: `CodexAuthError` looks like it is about the knowledge store
- **Preset configs**: `provider: "codex"` looks related to the knowledge tool
- **Documentation**: every mention needs a disambiguator

### Root cause

"Codex" was OpenAI's internal product name for what became ChatGPT. The LLM adapter was named after the product. The knowledge store inherited the name because they were built alongside each other.

### Proposed fix

Rename the LLM adapter from `"codex"` to `"chatgpt"` in `llm/_register.py:82`:

```python
LLMService.register_adapter("chatgpt", _codex)  # was "codex"
```

Leave the knowledge store as `"codex"` — the name is apt.

### Blast radius (verified)

**3 files, ~6 line changes:**

| File | What changes |
|---|---|
| `llm/_register.py:82` | `register_adapter("codex", _codex)` → `register_adapter("chatgpt", _codex)` |
| `preset_connectivity.py:37` | `"codex": "https://chatgpt.com"` → `"chatgpt": "https://chatgpt.com"` |
| `~/.lingtai-tui/presets/templates/codex.json` | `"provider": "codex"` (×3) → `"provider": "chatgpt"`; rename file to `chatgpt.json` |

**NOT affected** (these use "codex" as the knowledge store capability):
- `capabilities/__init__.py:18,89` — capability registry (`"codex": "lingtai.core.codex"`)
- All `.agent.json` files with `"codex"` in capabilities arrays
- `core/codex/__init__.py` — the knowledge store itself
- `core/avatar/__init__.py:463,465` — avatar deep-copy copies `codex/` directory
[...]

PR #16, opened 2026-05-07
feat(subconscious): implement Steps 1-3 (issue #51)

huangzesen:
## Summary

Implements the subconscious architecture as specified in the v2 proposal (issue #51). The subconscious is a parallel consultation that runs alongside the normal soul flow, asking a past self "does this remind you of something?" and surfacing relevant memories as brief, TTL-based insights via the existing notification system.

### Step 1 — Config + prompt
- Add `subconscious_*` fields to `AgentConfig` (all defaults: off)
- Add `_SUBCONSCIOUS_SYSTEM_PROMPT` constant in `consultation.py`
- Add i18n strings (en/zh/wen) for notification and config descriptions

### Step 2 — Core fire + TTL
- Add `_run_subconscious_fire()` in `flow.py` — consults 1 random past snapshot with the subconscious prompt, writes `.notification/subconscious.json` with TTL if insight is produced
- Modify `_soul_whisper()` to call `_run_subconscious_fire` when enabled
- Add `_check_subconscious_ttl()` in `lifecycle.py` — clears expired insights in the heartbeat loop before notification sync

### Step 3 — Config UI + persistence
- Extend `_handle_config` to accept `subconscious_enabled` and `subconscious_ttl_seconds`
- Persist to `manifest.soul.subconscious` in `init.json`
- Surface subconscious config in agent manifest (`.agent.json`)
- Add schema entries for agent-callable `soul(action='config')` params

## Testing

- 20 new tests in `test_subconscious.py` covering: config defaults, TTL expiry, notification writing, prompt injection, i18n strings, edge cases
- Full targeted suite: **234 passed, 0 failed**
- Default off — zero behavior change for existing agents

## Design Notes

Reuses the existing `.notification/` filesystem protocol. No new infrastructure needed — one new notification file (`.notification/subconscious.json`) with TTL-based expiry, auto-cleared by the heartbeat.

Refs: Lingtai-AI/lingtai#51


huangzesen:
## Design redirect — please hold this PR before merge

After auditing this PR I had a longer design conversation with @huangzesen about the relationship between soul flow and the subconscious. The conclusion: they're conceptually distinct features that should share one implementation primitive, and several decisions in this PR bake in the *separate-engine* assumption we want to avoid. Documenting the redesign here so the PR can be reshaped before merge.

### Audit findings (current code)

**Verified good:**
- Imports clean (smoke test: `import lingtai_kernel.intrinsics.soul.{flow,config,consultation}; import lingtai_kernel.base_agent.lifecycle` all succeed; `AgentConfig().subconscious_enabled == False`).
- Default-off claim correct: `subconscious_enabled: bool = False` (`config.py:42`); `flow.py:90` gates the call site on it.
- `_check_subconscious_ttl` exists at `lifecycle.py:450`, runs in heartbeat loop before `_sync_notifications`, deletes via `notifications.clear`. Uses `time.time()` (consistent with rest of heartbeat).
- 20 tests pass (`pytest tests/test_subconscious.py` → `20 passed in 0.58s`).
- i18n parity: 4 new keys present in all 3 locales (en/zh/wen) — strict diff is empty.
- ANATOMY updates real (`base_agent/ANATOMY.md`, `intrinsics/soul/ANATOMY.md`).

**Two HIGH issues (would need fixing if we merged as-is — but see redesign below):**

1. **Reload path drops the config.** `agent.py:982` `_read_init` reconstructs `AgentConfig` with explicit literals for `soul_delay`, `consultation_past_count`, `soul_voice`, `soul_voice_prompt` — but does NOT read `manifest.soul.subconscious.{enabled,ttl_seconds}`. So `_persist_soul_config` writes them to init.json successfully, but on reboot the value silently reverts to `False`. None of the 20 unit tests catch this because they mock `_config` directly; no round-trip test.

2. **`init_schema.validate_init` doesn't list `subconscious`.** The `manifest.soul` allow-list (`init_schema.py:194-199`) is `{delay, consultation_past_count, voice, voice_prompt}`. Currently nested-key warnings are lax so this doesn't error, but the schema should document the new field.

---

### The deeper issue: these should be one engine, two features

The current PR implements subconscious as a **separate consultation engine** with its own fire function (`_run_subconscious_fire`), its own session construction, its own config block. That's redundant — soul flow and subconscious are the same primitive (\"send a prompt to a self reading a substrate, get blocks back\") in different operating modes.

Unified mental model:

> **One consultation primitive.** Two prompts (advise / remind). Two delivery modes (wire pair / JSONL append). Two cadences (idle wall-clock / mid-turn wall-clock). Substrate is independent of all three.

| Feature | When fires | Substrate | Prompt | Delivery | Default |
|---|---|---|---|---|---|
| **Soul flow** | 300s timer, **IDLE-only** (not ACTIVE) | 1 current chat (\"insights\") + K random snapshots | \"advise your present self\" | Synthesized `(tool_call, tool_result)` wire pair (interrupts agent) | on |
| **Subconscious** | 60s timer, **only while agent is in a tool-call loop**; cleared at turn end | **ALL** snapshots (cheap-model fan-out) | \"does this remind you of something?\" | Append insight lines to `.notification/subconscious.jsonl`; agent always sees last 10 newest-last | off, requires explicit cheap-model config |

### Behavioral problems this fixes

- **Soul flow firing while ACTIVE is annoying.** Current code allows ACTIVE+IDLE; firing mid-turn means the synthesized pair lands while the agent is mid-thought, and the consultation prompt's tool-call recommendations (`Wanted to: ...`) splice next to a real tool loop. Looks like ghost tool calls. **Fix: gate soul flow to IDLE only.**
- **Subconscious never \"congregates in hours.\"** It's an intra-loop assistive presence, not a long-term memory layer. Born when the tool-call loop starts, fires every 60s while the loop is live, dies (JSONL clears) at turn end. The agent never sees subconscious insights from a previous turn.
- **Soul flow vs subconscious become time-orthogonal:** flow fires only IDLE, subconscious fires only mid-turn. They literally cannot fire simultaneously.

### Design decisions locked during discussion

1. **Two-knob config** (kept distinct in user-facing surface, even though the engine is shared internally):
   - `soul_flow_enabled: bool = True` — default on (existing wake-up behavior, just IDLE-gated now)
   - `subconscious_enabled: bool = False` — default off, requires explicit cheap-model config

2. **Cheap-model gating for subconscious is hard, not soft.** `soul(action='config', subconscious_enabled=True)` errors out unless `subconscious_provider` AND `subconscious_model` are explicitly set. They MAY equal the primary's values — the user just has to *say so* explicitly. No silent fallback. Rationale: the per-turn fan-out cost means an accidental \"use the primary model\" is a billing surprise; forcing explicit choice makes the user confront the cost equation at config time.

3. **Subconscious cadence: 60s wall-clock, fully async, fan-out across ALL snapshots per fire.** Not random sample. The cheap model is what makes this affordable. Per-snapshot daemon threads, per-snapshot timeout = `retry_timeout` (300s default), each thread appends to JSONL independently on non-null insight. Fire-level lock (`_subconscious_fire_lock`, try-acquire non-blocking) prevents overlapping fires (skip-if-in-flight at the next tick).

4. **Subconscious lifecycle: tool-call loop only.** Timer starts on turn-start (any turn — `send()` called, LLM hasn't returned terminal text yet), cancels on turn-end. JSONL clears at turn end. No subconscious activity while IDLE/ASLEEP/SUSPENDED.

5. **Storage: `.notification/subconscious.jsonl` append-only.**
   - One JSON object per line: `{ts, fire_id, insight, confidence, source_memory, source_snapshot, model_used}`.
   - Null insights (\"nothing relevant\") don't append.
   - **No TTL.** No `_check_subconscious_ttl`. No heartbeat pruning. The whole TTL machinery in this PR deletes.
   - Display: agent always sees **last 10 lines, newest-last** (chronological, matching diary-cue rendering).
   - Tail-read via `_iter_lines_reverse` (already exists in `consultation.py` for the diary cue).
   - File clears at turn end (not append-forever) — \"never congregate in hours\" is enforced positionally.

6. **Render surface for subconscious: deferred decision** — TBD whether it's notification-metadata-style (quiet, attached to next read) or a dedicated always-visible section in the prompt. Not blocking the engine refactor.

### Proposed module layout (after refactor)

```
src/lingtai_kernel/intrinsics/soul/
├── __init__.py        — re-exports + handle()/get_schema (existing)
├── config.py          — _handle_config (expanded with subconscious_provider/model required-when-enabled validation)
├── consultation.py    — SHARED ENGINE (existing _run_consultation renamed → _run_consultation_voice)
│                       _render_current_diary, _list_snapshot_paths,
│                       _load_snapshot_interface, _fit_interface_to_window,
│                       _send_with_timeout (all existing helpers stay)
│                       _CONSULTATION_SYSTEM_PROMPT (\"advise\")
│                       _SUBCONSCIOUS_SYSTEM_PROMPT (\"remind\")
├── flow.py            — SOUL-FLOW CALLER (existing _run_consultation_fire, IDLE-gated)
├── subconscious.py    — NEW SUBCONSCIOUS CALLER
│                       _start_subconscious_timer / _cancel_subconscious_timer (turn-lifecycle hooks)
│                       _subconscious_tick (60s timer callback, async fan-out)
│                       _run_subconscious_snapshot (one snapshot's voice, daemon thread, JSONL append on non-null)
│                       _read_subconscious_tail (reverse-seek last N lines)
│                       _clear_subconscious_jsonl (at turn end)
│                       _SUBCONSCIOUS_FIRE_INTERVAL = 60.0
│                       _SUBCONSCIOUS_DISPLAY_N = 10
└── inquiry.py         — synchronous mirror session (existing, untouched)
```

### Shared engine signature

```python
def _run_consultation_voice(
    agent,
    iface: ChatInterface,           # substrate (current chat or snapshot)
    source: str,                    # label, e.g. \"insights\" or \"snapshot:<stem>\"
    *,
    system_prompt: str,             # _CONSULTATION_SYSTEM_PROMPT or _SUBCONSCIOUS_SYSTEM_PROMPT
    spark: str,                     # cue text (e.g. _render_current_diary output)
    session_overrides: dict | None = None,  # {provider, model, base_url, context_window} for cheap-model path
    allow_tool_recommendations: bool = True,  # flow=True, subconscious=False
    max_rounds: int = 3,
) -> dict | None:
    \"\"\"Returns {'source': str, 'blocks': [ContentBlock, ...]} or None.
    Handles tool-call interception with refusal blocks when allow_tool_recommendations=True.
    Handles timeout via _send_with_timeout. Logs token usage.\"\"\"
```

The two callers (`flow._run_consultation_fire`, `subconscious._run_subconscious_snapshot`) compose `_run_consultation_voice` calls and own their own delivery side.

### Suggested next steps

1. **Don't merge this PR as-is.** Even with the two HIGH bug fixes, merging bakes in the separate-engine assumption.
2. **Cheap independent fix that lands now: gate soul flow to IDLE-only.** That fixes the \"annoying mid-turn fire\" complaint immediately and is a small standalone change in `flow._soul_fire_allowed`. Could ship as its own PR before the bigger refactor.
3. **Refactor on top of #2:** extract `_run_consultation_voice` shared engine, then build the new `subconscious.py` caller around the turn-lifecycle hooks + JSONL fan-out. The 590 lines of existing tests mostly test the right things and will need light reshaping (drop TTL tests, add fan-out + tool-call-loop-gating tests, add a config-time required-fields test, add a config round-trip-through-init.json test).
4. **Net effect:** the PR shrinks (~150 lines of duplicated machinery deletes, all TTL machinery deletes), the boot/reload bug becomes a single config-plumbing block instead of six fields, the storage format gets simpler, and the user-facing concepts stay distinct.

cc @huangzesen — flag if any of the locked decisions above don't match your intent.

huangzesen:
Closing — superseded by the soul flow system (merged in #36-#42). This experiment is no longer needed.

PR #29, opened 2026-05-07
fix(mcp): clear stale MCP notifications when inbox is empty

huangzesen:
## Problem

MCP notifications (e.g. ) were published via  when events were found, but never cleared when the inbox was empty. This caused stale '1 new event' notifications to persist indefinitely in .

## Fix

Added an  branch in  that clears the MCP notification file when the sweep finds zero pending events for that MCP. This matches the email notification pattern — publish on unread, clear when empty.

## Changes

- : Added  call when


huangzesen:
Closing per Jason's guidance: 'For mcp notifications, let each mcp manage their notification, the kernel shall not clean up.'

The kernel should not clear MCP notifications — each MCP manages its own notification lifecycle. For Telegram, the notification now shows the last 10 rounds of conversation and persists even when the inbox is empty.

PR #32, opened 2026-05-08
feat: shared overflow recovery + remove force-wipe

huangzesen:
## Summary

Moves context-overflow recovery from OpenAI-only to a shared base-class mechanism, and removes the kernel force-wipe that was wiping agent context without consent.

### Changes

**ChatSession base class** ():
-  — abstract, returns False by default; providers override
-  — drops ~10% of oldest non-system entries, snapping past tool-call/result boundaries
-  — retry loop (up to 10 rounds × 10% trim)
-  — post-recovery user message recommending molt

**Anthropic adapter** ():
- Implements  for 
-  and  now use 

**OpenAI adapter** ():
- Removed duplicate , , 
- Kept only  (OpenAI-specific)

**turn.py** ():
- Removed force-wipe () on hard ceiling
- Agent gets warnings and can molt voluntarily
- If agent ignores warnings, the API call will overflow and the adapter-level recovery trims + retries

### Inheritance coverage
- DeepSeek, OpenRouter → inherit from OpenAI (already covered)
- MiniMax → inherits from Anthropic (already covered)
- Custom → delegates to the above (already covered)
- Gemini → not yet implemented (server-side state management is more complex)

### Why
Jason's direction: move the 10% trim to an upper layer abstraction, remove the force-wipe that destroys agent memory without consent.

### DIFF
--- a/src/lingtai/llm/anthropic/adapter.py
+++ b/src/lingtai/llm/anthropic/adapter.py
@@ -315,6 +315,27 @@ class AnthropicChatSession(ChatSession):
         """The canonical ChatInterface for this session."""
         return self._interface
 
+    # -- Context-overflow detection (Anthropic-specific) -------------------
+
+    @staticmethod
+    def _is_context_overflow_error(exc: Exception) -> bool:
+        """Detect Anthropic context-length-exceeded errors."""
+        if not isinstance(exc, anthropic.BadRequestError):
+            return False
+        msg = (str(exc) or "").lower()
+        return any(
+            needle in msg
+            for needle in (
+                "too many tokens",
+                "context length",
+                "context window",
+                "prompt is too long",
+                "input is too long",
+                "maximum context",
+                "request too large",
+            )
+        )
+
     def _build_request_kwargs(self, messages: list[dict]) -> dict[str, Any]:
         kwargs: dict[str, Any] = {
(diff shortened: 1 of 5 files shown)


huangzesen:
## Code Review Summary

Reviewed PR #32 using Claude Code. Here are the findings:

### ✅ Core Implementation: Correct
- Base-class overflow recovery in  is well-designed
-  correctly skips system entries and handles ToolCall/ToolResult boundaries
- 10% trimming approach is appropriate (10 rounds × 10% ≈ 65% headroom)
- Anthropic's  covers all known error strings

### 🔴 Blocker Fixed
** was NOT updated** — two tests asserted  but force-wipe was removed. Fixed in latest commit:

-  → renamed to , now asserts  and counter increments to 2
-  → updated turn 3 assertions:  and counter increments to 3

### 🟡 Minor Issues (not blocking)
1. ** unused in Anthropic ** — cosmetic, works via  but confusing
2. **Stale block comment in  (~line 498)** — still describes removed force-wipe behavior
3. ** counter grows unboundedly** — harmless (capped where used) but semantic oddity
4. **No Anthropic-specific overflow test** — base-class logic tested via OpenAI, but Anthropic's  has no dedicated test

### Full Review
Full review saved to .

PR #67, opened 2026-05-10
fix(daemon): resolve shorthand preset names against allowed presets

### LINKED ISSUE #65: Daemon preset shorthand names rejected: 'must end in .json or .jsonc'
## Problem

When spawning daemons with a `preset` parameter, shorthand names like `'opus'`, `'claude-code'`, `'sonnet'`, `'mimo'` are rejected by `load_preset()` with the error: `preset name 'opus': must end in .json or .jsonc`.

This blocks **all daemon delegation** when the LLM generates a shorthand preset name instead of a full file path. Observed as 9 failures across multiple days in agent logs.

## Root Cause

In `src/lingtai/core/daemon/__init__.py:796-798`:

```python
# Validate preset exists and is loadable
try:
    preset = load_preset(preset_name, working_dir=self._agent._working_dir)
```

The `preset_name` comes directly from the task spec (LLM-generated). `load_preset()` requires a file path with `.json`/`.jsonc` extension — it has no name-to-path resolution.

Meanwhile, the agent's allowed presets are full paths in `manifest.preset.allowed`:
```
~/.lingtai-tui/presets/saved/mimo-1.json
~/.lingtai-tui/presets/saved/deepseek-1.json
...
```

When the LLM generates `preset='opus'`, there's no resolution step that maps it to an actual file.

## Observed Impact

- 9 daemon spawn failures with errors like:
  - `preset name 'opus': must end in .json or .jsonc`
  - `preset name 'claude-code': must end in .json or .jsonc`
  - `preset name 'sonnet': must end in .json or .jsonc`
- All delegation was blocked — agents couldn't spawn any daemons with explicit presets
- Agents had to fall back to inheriting parent preset (no explicit preset)

## Proposed Fix

**Option A (recommended)**: Add shorthand resolution in the daemon pre-flight. Before calling `load_preset`, resolve the name against the agent's allowed presets:

1. If the name already has `.json`/`.jsonc` extension → use as-is (current behavior)
2. Otherwise → scan `manifest.preset.allowed` for a file whose stem matches the name (case-insensitive). If exactly one match, use it. If zero or ambiguous, error with the list of available presets.

```python
# In daemon __init__.py, before load_preset call:
if not preset_name.endswith(('.json', '.jsonc')):
    # Try to resolve shorthand against allowed presets
    from lingtai_kernel.presets import resolve_allowed_presets
    import json
    init = json.loads((self._agent._working_dir / 'init.json').read_text())
    allowed = resolve_allowed_presets(init.get('manifest', {}), self._agent._working_dir)
    matches = [p for p in allowed if p.stem.lower() == preset_name.lower()]
    if len(matches) == 1:
        preset_name = str(matches[0])
    elif len(matches) > 1:
        return {"status": "error", 
                "message": f"ambiguous preset name {preset_name!r}: {[str(m) for m in matches]}"}
    else:
        return {"status": "error",
                "message": f"no preset matching {preset_name!r}; available: {[p.stem for p in allowed]}"}
```

**Option B**: Make `load_preset` itself support name resolution. This would fix it everywhere (daemon, system refresh, avatar spawn) but changes the core function's contract.

**Option C**: Add a `resolve_preset_shorthand()` utility that the daemon, system, and avatar code can all use.

## Additional Context

- The `system(action='presets')` command correctly lists presets by their full paths
- The daemon tool's schema says `preset: str` without documenting that it must be a full path
- The same issue likely affects `avatar(spawn)` if it accepts a preset parameter

huangzesen:
Fixes #65.

When spawning daemons with a preset parameter, shorthand names like 'opus', 'claude-code', 'sonnet' were rejected by load_preset() because they don't have .json/.jsonc extensions.

Now the daemon pre-flight resolves shorthand names against the agent's allowed presets list (manifest.preset.allowed) by matching the stem (case-insensitive). If exactly one match, use it. If ambiguous or no match, return a clear error with available preset names.

### DIFF
--- a/src/lingtai/core/daemon/__init__.py
+++ b/src/lingtai/core/daemon/__init__.py
@@ -783,9 +783,21 @@ class DaemonManager:
 
         # Pre-flight: resolve any per-task presets BEFORE scheduling.
         # If any preset is invalid, refuse the whole batch. Presets are
-        # identified by path (~/foo.json, ./foo.json, or absolute).
-        from lingtai.presets import load_preset
+        # identified by path (~/foo.json, ./foo.json, or absolute) or
+        # by shorthand name (stem of an allowed preset, case-insensitive).
+        from lingtai.presets import load_preset, resolve_allowed_presets
         from lingtai.preset_connectivity import check_connectivity
+        import json as _json
+
+        # Pre-load allowed presets for shorthand resolution
+        _init_path = self._agent._working_dir / "init.json"
+        try:
+            _init_data = _json.loads(_init_path.read_text(encoding="utf-8"))
+            _allowed_paths = resolve_allowed_presets(
+                _init_data.get("manifest", {}), self._agent._working_dir
+            )
+        except Exception:
+            _allowed_paths = []
 
         resolved_presets: list[dict | None] = []  # one entry per task — None means inherit
         for spec in tasks:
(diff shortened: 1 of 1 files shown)


huangzesen:
Closing in favor of a schema-first approach. Instead of code adapting to shorthand names, the tool schema should tell the model the correct naming convention. See issue #65 for updated design discussion.

PR #98, opened 2026-05-13
Preserve executed tool results after continuation send failure

TZZheng:
## Summary

Fixes a recovery edge case where a tool has already executed locally and returned a real result, but the subsequent post-tool LLM continuation fails. Some provider adapters append the `user[ToolResultBlock]` entry before provider I/O and roll it back on API/provider error. That can leave the canonical chat tail as `assistant[tool_calls]` even though the tool completed, allowing AED/notification heal paths to synthesize a misleading `[kernel notice — tool call did not complete]` placeholder and lose the real output.

This PR preserves the real tool result before generic recovery runs.

## Changes

- Add `_restore_tool_results_after_continuation_failure(...)` in `src/lingtai_kernel/base_agent/turn.py`.
- Wrap the normal `_process_response` post-tool continuation send (`_send_with_watchdog(agent, tool_results)`) in `try/except`.
- If continuation send raises and pending tool calls remain, commit the already-produced real tool results, save chat history, log `tool_results_restored_after_continuation_failure`, then re-raise the original exception so existing AED/retry semantics remain unchanged.
- Add focused regression tests in `tests/test_tool_result_restore_after_continuation_failure.py`, including helper-level coverage and an `_process_response`-level fake executor/session regression.

## Tests

Focused tests and lint from `/Users/tianzhezheng/Documents/lingtai-dev/lingtai-kernel`:

```bash
.venv/bin/python -m pytest -q \
  tests/test_tool_result_restore_after_continuation_failure.py \
  tests/test_chat_interface_invariant.py \
  tests/test_worker_still_running_recovery.py
# 47 passed in 0.29s

.venv/bin/ruff check \
  src/lingtai_kernel/base_agent/turn.py \
  tests/test_tool_result_restore_after_continuation_failure.py
# All checks passed
```

Local full-suite note: this environment currently is not a clean full-suite signal; a prior full `uv run pytest` run hit many unrelated existing failures and eventually aborted inside `ddgs/http_client.py` threads. This PR therefore submits the minimal required code/test changes with focused coverage; CI should provide the authoritative full-suite result.

## Risk / follow-up

Risk is limited to the path where local tool execution already returned results and the subsequent LLM continuation raises. If the adapter already kept the tool-result entry, the helper no-ops to avoid duplicates.

A cleaner future design would decouple tool-result durability from continuation success entirely: commit tool results before provider continuation and continue from the staged wire (`send(None)`), with provider-level tests.

### DIFF
--- a/src/lingtai_kernel/base_agent/turn.py
+++ b/src/lingtai_kernel/base_agent/turn.py
@@ -127,6 +127,35 @@ def _prepare_aed_retry_message(agent, err_desc: str) -> Message:
     return _make_message(MSG_REQUEST, "system", aed_msg)
 
 
+def _restore_tool_results_after_continuation_failure(
+    agent,
+    tool_results,
+    *,
+    ledger_source: str,
+) -> bool:
+    """Persist real tool results after post-tool LLM continuation failure.
+
+    The tools already executed locally before the continuation send. If an
+    adapter rolled back the attempted user tool-result entry on provider error,
+    the canonical interface tail still has pending assistant tool_calls. Restore
+    the real results before AED / notification heal can synthesize a placeholder
+    that lacks the actual result payload.
+    """
+    if (
+        not tool_results
+        or agent._chat is None
+        or not agent._chat.has_pending_tool_calls()
+    ):
+        return False
+    agent._chat.commit_tool_results(tool_results)
+    agent._save_chat_history(ledger_source=ledger_source)
+    agent._log(
+        "tool_results_restored_after_continuation_failure",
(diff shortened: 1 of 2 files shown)


huangzesen:
Cherry-picked to `main` as commit 58a9932 (preserving original authorship). Tested locally: 33 focused tests pass; full suite 1654/1660 — the 6 failures are pre-existing on `main` (3 soul-consultation tests, 3 molt-notification tests broken by test pollution earlier in the run) and unrelated to this change. Thanks!

PR #128, opened 2026-05-20
Fix #126: IMAP check/read tool result orphaning

huangzesen:
## Summary

Fixes #126 where IMAP `check`/`read` tool results were orphaned (replaced by `heal:idle_inject_blocked` placeholders) during live email workflows.

## Root Cause

After `poll_backoff` commits tool results via `commit_tool_results()` and the agent goes IDLE, a later notification heartbeat tick still saw `has_pending_tool_calls()` returning True. This triggered `_heal_pending_tool_calls()` which replaced the real results with synthesized abort placeholders — the agent then received "[kernel notice — tool call did not complete]" instead of the actual IMAP data.

The investigation identified two contributing bugs:

1. **Missing `synthesized` flag serialization**: `ToolResultBlock.to_dict()` did not include the `synthesized` field, so it was lost on save/reload. This caused `add_tool_results()` to create duplicate entries instead of replacing synthesized placeholders in-place.

2. **Unsaved state window in poll_backoff path**: `commit_tool_results()` updated the in-memory interface but the save happened later in `_run_loop`. A heartbeat notification sync between the return and the save could see stale wire state.

## Changes

1. **`interface.py`**: Serialize `synthesized` flag in `ToolResultBlock.to_dict()` (only when True, to minimize noise) and restore it in `content_block_from_dict()`.

2. **`turn.py`**: Add `_save_chat_history()` call immediately after `commit_tool_results()` in the poll_backoff path, closing the window where in-memory and on-disk state could diverge.

3. **`__init__.py`**: Add diagnostic logging to `notification_inject_aborted` that captures the interface tail shape (role, block count, tool_call IDs) so the exact wire state can be traced if the issue recurs.

## Testing

- 285 existing tests pass (1 pre-existing failure unrelated to this change — `test_daemon_check_returns_error` fails due to temp file cleanup issue).
- Manual verification: `add_tool_results` correctly replaces synthesized placeholders when the flag survives serialization roundtrip.
- The diagnostic logging will capture the exact tail state if the issue recurs in production.

## Linked Issue

Closes #126

### DIFF
--- a/src/lingtai_kernel/base_agent/__init__.py
+++ b/src/lingtai_kernel/base_agent/__init__.py
@@ -1070,9 +1070,20 @@ class BaseAgent:
         # If the wire has unanswered tool_calls, appending a user-role
         # result entry would violate the alternation invariant.  Defer.
         if iface.has_pending_tool_calls():
+            # Issue #126 diagnostic: log the tail shape so we can trace
+            # why tool results were not detected as committed.
+            tail_info = ""
+            if iface._entries:
+                last = iface._entries[-1]
+                tail_info = f" tail_role={last.role} tail_blocks={len(last.content)}"
+                if last.role == "assistant":
+                    tc_ids = [b.id[:20] for b in last.content
+                              if hasattr(b, 'id') and hasattr(b, 'name')]
+                    tail_info += f" tc_ids={tc_ids}"
             self._log("notification_inject_aborted",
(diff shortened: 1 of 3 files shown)


huangzesen:
Superseded by PR #133 (clean cherry-pick of the 3 relevant fixes only)

PR #106, opened 2026-05-16
Fix pending tool-call wake recovery

TZZheng:
## Summary

Fixes a stall where an agent can remain `IDLE` with a dangling assistant tool-call tail after a tool-call response is persisted but no tool result is committed. In that state, repeated `MSG_TC_WAKE` handling previously logged `tc_wake_noop reason=pending_tool_calls` and returned forever until a later notification injection happened to heal the wire.

This patch implements the first two scoped fixes discussed:

- `_handle_tc_wake()` now heals a pending `assistant[tool_calls]` tail with synthetic recovery tool results and then falls through to the existing wire-drive path (`session.send(None)`) so the model sees the recovery notice immediately.
- `_process_response()` now closes pending tool-call tails on pre-dispatch cancel/guard and empty-result early-return paths, so normal returns do not leave an unappendable wire.
- Adds batch-level logging (`tool_batch_received`, `tool_batch_dispatch_start`, `tool_batch_dispatch_done`) to distinguish “tool never dispatched” from “tool dispatched but continuation/result handling failed”.

Deliberately **does not** add a run-loop post-turn IDLE invariant; this PR keeps the scope to the first two repair points.

## Why this is worth fixing: observed stall evidence

This is based on a real local stall investigation, not just a theoretical invariant violation. The key failure pattern was:

1. The LLM response containing an internal `email(action='check')` tool call was persisted as the tail assistant message.
2. No corresponding `tool_call` / `tool_result` events were recorded for that call id, strongly suggesting the tool executor never dispatched it.
3. The agent then repeatedly woke on queued `MSG_TC_WAKE` messages, but `_handle_tc_wake()` saw `pending_tool_calls` and no-oped instead of repairing the wire.
4. The agent stayed apparently `IDLE` for about 53 minutes, until a later human mail triggered notification injection. That path already had a pending-tool-call heal, so it finally synthesized the recovery result and resumed the agent.

Concrete log/history excerpts from the report:

```text
20:17:06.334  history/chat_history.jsonl:66
assistant tool_call id=call_wWKmJP8Bpl8qllD9Pu2on0CZ
name=email args={action:check, n:10, filter:{unread_only:true, truncate:500}}

20:17:06.334  logs/events.jsonl:42857
llm_response input_tokens=83689 output_tokens=55

20:17:06.345-20:17:06.378  logs/events.jsonl:42858-42896
agent_state idle→active reason="received tc_wake"
tc_wake_noop reason=pending_tool_calls
agent_state active→idle
... repeated ...

[...]

PR #143, opened 2026-05-22
feat(kernel): add turn-loop observability diagnostics

TZZheng:
## Summary

This PR focuses on **additional turn-loop logging/diagnostics** so future stuck or orphaned tool-call incidents have enough evidence to trace where execution stopped.

It keeps the existing recovery events as the semantic center and adds only one new narrow pre-dispatch event:

- enrich `tc_wake_noop reason="pending_tool_calls"` with pending tool-call count / ids / names
- enrich existing recovery logs when in-memory recovery succeeds but history save fails
- add `tool_calls_not_dispatched` for the specific boundary where the LLM response contains tool calls, but `_process_response()` does not hand them to `ToolExecutor` because of cancel / tool-loop limit / invalid-tool limit

`ToolExecutor` remains the source of truth for actual executed tool observability (`tool_call` / `tool_result`); this PR does not fake tool results for calls that were never dispatched.

## Relation to PR #106

This is primarily a logging/observability PR, not the main behavioral fix for #106.

It should make upstream causes behind #106-style pending tool-call failures trackable: if a model response includes tool calls but the turn loop intentionally stops before dispatch, the event stream now records that boundary and reason instead of leaving only a later `tc_wake_noop` symptom.

## Tests

Focused tests:

```text
uv run --with pytest pytest tests/test_aed_recovery.py tests/test_tc_wake_chat_not_ready.py tests/test_tool_result_restore_after_continuation_failure.py
20 passed
```

Also checked:

```text
git diff --check
```

Full suite was run once locally by the implementation daemon:

```text
1714 passed, 2 skipped, 7 failed
```

The observed full-suite failures appear outside this observability path (kernel isolation hard-coded local cwd, avatar description expectation, molt notification temp-file issue, and existing soul consultation expectations).

### DIFF
--- a/src/lingtai_kernel/base_agent/turn.py
+++ b/src/lingtai_kernel/base_agent/turn.py
@@ -116,6 +116,32 @@ def _is_transient_provider_error(exc: Exception) -> bool:
     return any(fragment in msg for fragment in _TRANSIENT_MSG_FRAGMENTS)
 
 
+def _tool_call_summary(tool_calls) -> dict:
+    calls = list(tool_calls or [])
+    return {
+        "call_count": len(calls),
+        "call_ids": [getattr(call, "id", None) for call in calls],
+        "tool_names": [getattr(call, "name", None) for call in calls],
+    }
+
+
+def _pending_tool_call_summary(iface) -> dict:
+    entries = getattr(iface, "entries", None) or []
+    tail = entries[-1] if entries else None
+    calls = []
+    if getattr(tail, "role", None) == "assistant":
+        calls = [
+            block
+            for block in getattr(tail, "content", []) or []
+            if hasattr(block, "id") and hasattr(block, "name") and hasattr(block, "args")
+        ]
+    return {
+        "pending_tool_call_count": len(calls),
+        "pending_tool_call_ids": [getattr(call, "id", None) for call in calls],
+        "pending_tool_names": [getattr(call, "name", None) for call in calls],
+    }
+
+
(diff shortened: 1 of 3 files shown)


huangzesen:
Closing this PR because its core diagnostics have now been reworked against current `main` and merged in #177.

Why not merge this original branch directly:

- This PR was stale/conflicting against current `main` (`mergeStateStatus=DIRTY` / conflicting turn-loop test areas).
- The turn loop changed substantially after this PR, including later recovery and observability work.
- We wanted to preserve the useful idea here without reviving an old branch or risking stale test/source context.

What was done instead:

- #177 (`feat(kernel): add turn-loop observability diagnostics`) reimplemented the useful diagnostics-only pieces on top of current `main`.
- It adds bounded tool-call summaries for `tc_wake_noop reason="pending_tool_calls"`.
- It adds `tool_calls_not_dispatched` events for the cancel / tool-loop-limit / invalid-tool-limit boundaries before `ToolExecutor.execute()`.
- It improves save-failure diagnostics around pending-tool-call healing / continuation-result restoration.
- It was validated with the focused tc_wake / continuation restore suites plus AED/orphan-heal safety tests.

Merged replacement:

- #177: https://github.com/Lingtai-AI/lingtai-kernel/pull/177
- merge commit: `59d588e46aa803582aed703a1c52b1986648cff7`

Thanks for the original diagnostics direction — the current-main replacement keeps that intent while avoiding the stale branch risk.

PR #168, opened 2026-05-28
perf: prune grep glob filters before file reads

huangzesen:
## Summary

Low-risk first pass at improving the built-in `grep` file tool, inspired by oh-my-pi's search harness design but staying pure Python / dependency-free:

- push the `grep` tool's `glob` filter down into `LocalFileIOService.grep()` so non-matching files are skipped before `stat()` / `open()`
- stream grep inputs line-by-line instead of loading whole files with `Path.read_text()`
- add a cheap NUL-byte prefix sniff to skip obvious binary files before text decoding
- keep malformed-but-mostly-text UTF-8 files searchable via `errors="replace"`
- add regression tests for pre-open glob pruning, streaming behavior, mixed UTF-8, binary skip, and backward-compatible `None` / `*` filters

## Why

Before this PR, `grep(pattern, glob="*.py")` still scanned every file under the search root, then discarded matches whose basenames did not match `*.py`. In large repos this wastes I/O on logs, JSON, JS bundles, etc. The service now prunes those files before opening them.

The old grep path also read every candidate file into memory before splitting lines. Streaming keeps memory bounded by line length rather than file size.

## Validation

```bash
python -m pytest tests/test_services_file_io.py tests/test_layers_file.py tests/test_agent_capabilities.py -q
# 56 passed in 1.50s

git diff --check
```

Claude Code initial validation before the final NUL-byte test also ran the broader related set:

```text
tests/test_services_file_io.py: 29 passed
notification_sync + layers_email + eigen + daemon_preset_capabilities + agent related tests: 237 passed
```

## Notes / follow-up

This intentionally does **not** change glob result ordering, add `.gitignore` parsing, add output modes, or introduce a native/ripgrep backend. Those are better split into follow-up PRs after the behavior/API surface is discussed.

### DIFF
--- a/src/lingtai/core/grep/__init__.py
+++ b/src/lingtai/core/grep/__init__.py
@@ -4,7 +4,6 @@ Usage: Agent(capabilities=["grep"]) or capabilities=["file"]
 """
 from __future__ import annotations
 
-import fnmatch
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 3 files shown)


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. This branch is stale and its useful file-tool/glob performance direction has been superseded by the newer merged work in #544, so it is not being carried forward here.

PR #235, opened 2026-06-08
Improve soul flow past-self routing

runyuan-wang:
## Summary

This PR changes soul-flow past-self consultation from fixed/random fanout into a routed, budgeted selection model:

- keep the shared current-self `insights` consultation as the always-on path
- treat `consultation_past_count` as a maximum rather than a guaranteed count
- route past snapshots by lexical overlap between the current diary cue and each snapshot's top-level `molt_summary`
- optionally add one bounded serendipity snapshot (25% chance) when routed selection leaves headroom
- suppress past-self consultations when a non-soul notification is pending or when there is no diary signal
- log route decisions with `consultation_route`
- keep `insights` at the historical 70% context budget while lowering past-self snapshots to 30%

The intent is a lightweight MoE-inspired shape: current-self reflection acts as the shared path; past selves become selected memory experts rather than a fixed random fanout.

## Motivation

Soul flow currently spends premium consultation calls on random past snapshots even when:

- a human-facing notification may be waiting
- the current diary has no useful signal
- a relevant past snapshot can be selected cheaply from existing `molt_summary` metadata

This PR keeps the creative/serendipitous quality of past-self consultation, but bounds it behind relevance, budget, and responsiveness guards.

## Implementation notes

- Routing uses cheap lexical overlap, not embeddings or full snapshot loading.
- Only `molt_summary` is read before selection; full snapshots are loaded only after routing.
- If routing finds no overlap but a diary exists, the system falls back to the legacy random sample behavior.
- Serendipity is limited to one extra snapshot and only when `max_past` has headroom.
- Pending notification guard ignores the `soul` channel and suppresses only past voices.
- Past-self context target is 30% of the active context window; `insights` remains 70%.

## Human-facing explainer

Included in the PR:

- `reports/pr235-soul-flow-routed-past-selves-explainer.html`

The explainer follows the PR-numbered convention required by the contributing guide.

## Validation

```bash
python -m pytest tests/test_soul_consultation.py -q
# 127 passed

python -m pytest tests/test_soul_consultation.py tests/test_soul.py tests/test_psyche.py tests/test_system_notifications.py tests/test_notification_sync.py -q
# 229 passed

python -m py_compile src/lingtai_kernel/intrinsics/soul/consultation.py src/lingtai_kernel/intrinsics/soul/__init__.py

git diff --check
```

## Review questions

- Is it acceptable that `consultation_past_count` now caps rather than guarantees past voices?
- Should the 30% past-self context budget become configurable later?
- Should the pending-notification guard eventually distinguish human channels from machine-only producers?


huangzesen:
Thank you for the contribution and for exploring ways to improve soul flow.

We are going to close this PR because this particular routing approach does not fit LingTai's design philosophy for soul flow. In particular, moving past-self selection into a kernel-side heuristic/runtime policy changes the character of soul flow in a direction we do not want to adopt.

This closure is about product/design direction, not a dismissal of the effort. We appreciate the work and the detailed tests/explainer, but we do not plan to merge this approach. Jason will continue improving the soul flow system in a direction that better matches LingTai's intended design.

PR #239, opened 2026-06-09
Validate secondary read targets

huangzesen:
## Summary\n- normalize secondary read args before dispatching to communication MCPs\n- reject empty secondary read targets locally instead of forwarding malformed args to MCP validators\n- coerce Telegram digit-string chat_id values to integers and add regression tests\n- include HTML explainer at reports/secondary-telegram-args-20260609.html\n\n## Context\nJason asked to investigate frequent Telegram send/read failures across dev bots. Current main has already made secondary read-only; this PR preserves that policy and fixes the remaining read-path argument validation/noise.\n\n## Validation\n- git diff --check\n- PYTHONPATH=src /Users/huangzesen/anaconda3/bin/python -m pytest -q tests/test_tool_executor.py tests/test_secondary_schema.py\n- PYTHONPATH=src /Users/huangzesen/anaconda3/bin/python -m py_compile src/lingtai_kernel/secondary_tools.py src/lingtai_kernel/tool_executor.py

### DIFF
--- a/reports/secondary-telegram-args-20260609.html
+++ b/reports/secondary-telegram-args-20260609.html
@@ -1,0 +1,49 @@
+<!doctype html>
+<html lang="zh-CN">
+<head>
+<meta charset="utf-8" />
+<title>Secondary Telegram args validation PR explainer</title>
+<style>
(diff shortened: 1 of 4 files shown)


huangzesen:
Closing per Jason's stale PR cleanup. Review conclusion: current patch targets deleted secondary_tools.py / removed secondary channel; file structure no longer exists. If bug persists, redo as a fresh PR.

PR #255, opened 2026-06-10
fix: recover overflow from empty OpenAI-compatible responses

ZacharyHu0:
## Summary
- detect empty successful OpenAI-compatible chat completions before recording assistant history
- treat the response as context overflow only when `usage.prompt_tokens` exceeds the configured `context_window`
- route that narrow empty-200 over-window case through the existing overflow recovery path, so history is trimmed and retried instead of AED retrying the same oversized prompt
- keep empty successful responses within the configured window unchanged
- add focused regression coverage for both the recovery and non-recovery cases

## Why
Some OpenAI-compatible gateways do not report over-window prompts as a `400 context_length_exceeded` error. Instead, they can return a structurally successful chat completion with no assistant text, no tool calls, no reasoning payload, and zero completion tokens while `usage.prompt_tokens` is already above the configured context window.

The old adapter only entered overflow recovery for provider errors. A successful empty response was therefore treated as a transient empty LLM output, so AED retried without trimming history. The next request stayed over-window and could repeat the same failure mode.

## Safety
The new conversion is intentionally narrow:

- `context_window` must be configured and positive
- provider usage must report `prompt_tokens > context_window`
- the response must have no content, no tool calls, and no reasoning payload

The check runs before `_record_assistant_response(raw)`, so successful recovery does not commit an empty assistant turn to chat history. Empty responses that are still within the configured window continue through the existing empty-response behavior.

## Tests
```bash
/Volumes/Data/miniconda3/bin/python3 -m pytest tests/test_openai_overflow_recovery.py
# 16 passed

/Volumes/Data/miniconda3/bin/python3 -m pytest tests/test_deepseek_adapter.py tests/test_mimo_adapter.py
# 22 passed

git diff --check
# pass
```

Pytest emitted a `.pytest_cache` write warning in this sandbox; test results were unaffected.

### DIFF
--- a/src/lingtai/llm/openai/adapter.py
+++ b/src/lingtai/llm/openai/adapter.py
@@ -38,6 +38,34 @@ from lingtai_kernel.llm.streaming import StreamingAccumulator
 logger = get_logger()
 
 
+class _ContextOverflowEmptyResponseError(RuntimeError):
+    """Provider returned an empty 200 response for an over-window prompt.
+
+    Some OpenAI-compatible gateways do not surface context overflow as a
+    400-class error.  They return a structurally valid ChatCompletion with
+    no text, no tool calls, zero output tokens, and prompt_tokens already past
+    the configured context window.  Treat that as the same recoverable
+    overflow signal as a canonical context_length_exceeded error so the
+    adapter trims and retries before committing an empty assistant turn.
+    """
+
+    def __init__(
+        self,
+        *,
+        input_tokens: int,
+        context_window: int,
+        raw: Any,
+    ):
+        self.input_tokens = input_tokens
+        self.context_window = context_window
+        self.raw = raw
+        super().__init__(
+            "context window exceeded: provider returned empty completion "
+            f"with {input_tokens} input tokens over configured context window "
+            f"{context_window}"
+        )
+
+
 _CODEX_RESPONSES_TRACE_ENV = "LINGTAI_CODEX_RESPONSES_TRACE"
 _CODEX_RESPONSES_TRACE_PATH_ENV = "LINGTAI_CODEX_RESPONSES_TRACE_PATH"
 _CODEX_RESPONSES_TRACE_FILE = "codex_responses_trace.jsonl"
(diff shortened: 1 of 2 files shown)


TZZheng:
Thanks for the focused fix. Before we consider merging, could you clarify the evidence behind the empty-200 overflow case?

1. Which OpenAI-compatible provider / gateway and model produced the successful-but-empty response?
2. Do you have the original raw response/log excerpt that shows this behavior? A sanitized excerpt is fine, but the useful fields would be things like `usage.prompt_tokens`, `usage.completion_tokens`, `choices[0].message` payload, `finish_reason` if present, and any request/model identifiers that are safe to share.

Reason for asking: the patch infers context overflow from an empty assistant payload plus `usage.prompt_tokens > context_window`. That is plausible and the guard is narrow, but before merging we should confirm this is a real provider behavior rather than a config/model-window mismatch or another empty-response semantic.

huangzesen:
Thanks @ZacharyHu0 — this is a valuable report and it helped us identify an important OpenAI-compatible failure mode: some gateways may return a structurally successful but empty response instead of a canonical context-overflow error.

After review, we’re going to close this PR rather than merge this implementation as-is. The current heuristic infers overflow from `prompt_tokens > context_window` plus an empty assistant payload, but in LingTai `context_window` can be a configured budget rather than the provider’s hard model limit. That means the patch could misclassify a non-overflow empty response and trim real conversation history. The patch also only covers the non-streaming path, while streaming would still record an empty assistant turn.

We do want to preserve the contribution: if/when we implement this failure-mode recovery, we’ll use a stricter design (likely `finish_reason == "length"` as the primary signal, with token usage as corroboration), cover streaming, and credit this PR as the report that surfaced the issue.

Thanks again for the focused tests and for bringing this edge case to our attention.

ZacharyHu0:
I'm glad the report was useful in identifying this OpenAI-compatible failure mode, and I'm happy to have contributed by surfacing the edge case. Please feel free to reach out if you'd like me to help test or review a future stricter implementation.

Thanks again for the thoughtful feedback . Looking forward to seeing the improved design land eventually.

PR #291, opened 2026-06-15
feat: add expired spill artifact messaging (#192)

### LINKED ISSUE #192: Spilled tool-result sidecars can disappear while archived history still tells agents to read them
## Summary

When a tool result is too large, LingTai writes the full result to `tmp/tool-results/<timestamp>-<tool>-<call-id>-<hash>.json` and leaves a placeholder in chat history that tells the model/operator to read that sidecar path. Local trajectory mining found a real case where those sidecar files have been deleted while archived chat history still points at them.

This creates a silent dead-reference/data-loss hazard: restored archived history can instruct the agent to read a path that no longer exists, with no hint that the spill artifact expired or was pruned.

## Verified evidence

Agent audited:

`/Users/huangzesen/work/projects/runyuan_wang/.lingtai/mimo-2-5-pro`

Independent verification on 2026-05-30:

- `history/chat_history.jsonl`: 18 `tmp/tool-results/*.json` references, 0 missing (live recent window is healthy).
- `history/chat_history_archive.jsonl`: 1,062 `tmp/tool-results/*.json` references, 1,050 references point to files absent on disk (duplicates included because the same missing sidecar can be referenced more than once).
- `logs/events.jsonl`: 412 `tool_result_spilled` events with `spill_path`; all 412 referenced sidecar files are absent on disk for this agent.
- `tmp/tool-results/`: 350 JSON files still present; oldest surviving file mtime is 2026-05-28 13:38 local (`20260528T203801-read-call_DZ7NLgVvxYq0orfOA4QB0xlZ-8a77fc.json`). Missing event spill paths are older, e.g. `tmp/tool-results/20260524T202426-read-call_FeL04Z8ibtrLPYpA1ArveCZg-f33097.json`.

Example placeholder/reference format in history says, effectively, “Tool result was too large ... written to a sidecar file ... Read it via the read tool,” then gives a `spill_path` under `tmp/tool-results/`. Once that path is gone, the instruction becomes a dead read.

No captured turn showed the agent actually attempting one of these missing reads yet, so this issue is about a structural retention/restore hazard rather than an observed crash.

## Impact

- **Data loss risk:** spilled content becomes unrecoverable once the sidecar is removed.
- **Dead-read hazard:** archived content restored into the live context can tell the model to read an ENOENT path.
- **Operator confusion:** no explicit retention/eviction event explains whether the missing file was intentionally pruned, manually deleted, or corrupted.
- **Trajectory analysis gap:** event logs and history retain pointers to artifacts that no longer exist, so later auditing cannot reconstruct what happened.

## Expected behavior

Spill sidecar lifetime should be explicit and consistent with history lifetime. Any of these would be acceptable improvements:

1. Do not prune a spill sidecar while retained live/archive history still references it, or co-prune/replace the history reference.
2. When a sidecar is pruned, emit a typed retention event such as `tool_result_evicted` with path/reason/timestamp.
3. Replace or annotate old placeholders so restored history says “spill artifact expired/pruned” instead of “read this path.”
4. If the read tool is asked for a missing `tmp/tool-results/...` spill path, return a graceful “spill artifact expired/pruned” diagnostic rather than a bare missing-file failure.

## Notes

Other audited agents did not show missing spill paths in this pass, so this may have been caused by manual cleanup/disk pressure rather than the kernel itself. The bug is that retained history does not have an explicit retention contract or graceful degradation when the spill sidecar disappears.

TZZheng:
## Problem

When a tool result exceeds 10K chars, LingTai writes the full content to `tmp/tool-results/<timestamp>-<tool>-<call-id>-<hash>.json` and replaces in-history content with a compact manifest dict that tells the agent to "Read it via the read tool" at the given spill path.

The core defect: there is no retention contract between spill sidecar lifetime and history lifetime. Sidecar files can be deleted (manual cleanup, disk pressure, OS behavior) while history still carries live pointers to those paths. This creates:

1. **Dead-read hazard** — agent reads spill path then gets no graceful degradation
2. **Silent data loss** — no event emitted when a referenced sidecar is removed
3. **Operator confusion** — no retention event explains whether missing files were pruned, deleted, or corrupted

**Scope:** This PR fixes live history (`chat_history.jsonl`) and the read tool. It does not fix archived history (`chat_history_archive.jsonl`) or event log references (`logs/events.jsonl`).

### 1. Manifest honesty (tool_result_artifacts.py)
- New spill manifests include `artifact_lifetime="ephemeral_tmp"` and `artifact_state="available"`
- Warning text updated to explicitly say the path is ephemeral and may disappear
- When sidecar write fails or `working_dir` is None, `artifact_state` is set to `"unavailable"` instead of `"available"`
- Retroactive compaction manifests also include the new fields

### 2. Stale-marking helper (tool_result_artifacts.py)
- New `mark_expired_spill_manifests(working_dir)` scans live history for spill manifests whose sidecar files are missing
- Sets `artifact_state="expired"` and `artifact_expired_at` timestamp on stale manifests
- Overwrites the old "read this path" warning with explicit expired guidance: content is gone, do not search, use preview or rerun source tool
- Idempotent

### 3. History restore integration (lifecycle.py, session.py)
- Called on startup after restoring chat history
- Also called on session restore
- Ensures stale manifests are marked before the LLM sees false "read this path" promises

### 4. Spill-aware read message (read/__init__.py)
- When agent tries to read a missing file under `tmp/tool-results/`, returns a spill-aware message instead of generic "File not found"
- Path is normalized before classification to prevent traversal false positives
- Generic missing files keep standard behavior

### 5. Tests (test_expired_spill_messaging.py)
- 15 tests covering: new manifest fields, stale marking, warning overwrite, idempotency, recognizer compatibility, generic file behavior, no archive directory creation, legacy manifest handling, session restore integration, unavailable state, path traversal

## Non-goals
- No durable archive or archive/tool-results directory
- No molt-time copying of spill files
- No retention policy changes
- No fix for archived history or event log references (out of scope for this PR)

## Validation
- Focused tests: 15 passed in 0.08s
- Broader regression suite: 167 passed in 13.90s (test_tool_result_spill, test_retroactive_history_compaction, test_compaction, test_notification_sync, test_agent)
- Zero regressions

Closes #192


huangzesen:
整体评估：实现干净、范围克制、测试充分（含幂等性、路径穿越防护、legacy 兼容、显式拒绝 durable-archive 的回归测试），是一个值得合入的修复。

唯一阻塞项是分支已落后 `main`，当前处于冲突状态（CONFLICTING / DIRTY），需要 rebase 后才能合并。

几点非阻塞建议（可后续处理）：
1. `lifecycle.py` 启动期 `except Exception: pass` 会吞掉所有异常，建议至少 debug 级别记录，避免个别 sidecar 抛错时静默残留过期指针。
2. `session._ensure_spill_manifest_fields` 与 `mark_expired_spill_manifests` 都在回填 `ephemeral_tmp`/`available` 默认值，建议抽成共享常量防止漂移。
3. read 工具消息（"Spill artifact expired:"）与磁盘标记（"EXPIRED:"）措辞略有差异，可统一。

代码层面未发现阻塞问题，rebase 后建议批准。

---
_双审清理备注：Claude-p + GLM 交叉复核；本评论只做 triage/comment，未执行 close/merge。 marker: mimo-1 community cleanup triage 2026-06-24_

PR #295, opened 2026-06-15
fix(turn): surface post-tool continuation hangs

huangzesen:
## Summary

Fixes the observability gap from the forwarded Dulin/Jason incident report: a `bash` timeout was correctly delivered as a structured tool result, but the following post-tool LLM continuation logged `llm_call` with no `llm_response` before manual refresh, making the agent look dead.

Current `main` already has important recovery primitives (`send_with_timeout`, `WorkerStillRunningError`, and real tool-result restoration after continuation failure). This PR keeps that design and adds the missing operator-visible diagnostics:

- records `post_tool_continuation_failed` with predecessor tool names, count, and approximate tool-result payload chars;
- carries that metadata through `WorkerStillRunningError` when the post-tool continuation worker is still alive;
- publishes a `kernel.llm_worker_hang` system notification that says the agent is ASLEEP, recommends `/refresh`, and includes predecessor tool context when available;
- adds regression coverage for the notification and predecessor-context logging.

## Validation

- `python -m py_compile src/lingtai_kernel/base_agent/turn.py tests/test_aed_recovery.py tests/test_tool_result_restore_after_continuation_failure.py`
- `git diff --check`
- `PYTHONPATH=$PWD/src python -m pytest -q tests/test_aed_recovery.py tests/test_tool_result_restore_after_continuation_failure.py`
  - `25 passed in 0.41s`

## Review / notes

- Claude Code produced the first draft but the CLI did not finish its final JSON report after ~7 minutes; I cancelled it, reviewed the retained patch, tightened the context propagation/notification assertions, and reran tests.
- A direct GLM 5.2 read-only review attempt timed out after 90s. The patch is intentionally small: no provider adapter changes, no AED redesign, and no broad bash-output compression.
- Local HTML explainer: `reports/post-tool-continuation-visible-20260615.html` in the worktree (ignored/local-only).

## Non-goals

- Does not replace provider-specific overflow recovery.
- Does not implement global bash output compression/truncation.
- Does not change the existing worker-hang recovery state machine beyond adding notification/log metadata.

### DIFF
--- a/src/lingtai_kernel/base_agent/turn.py
+++ b/src/lingtai_kernel/base_agent/turn.py
@@ -129,6 +129,33 @@ def _tool_call_summary(tool_calls) -> dict:
     }
 
 
+def _tool_result_context(tool_results) -> dict:
+    """Extract diagnostic metadata from tool results before a continuation send.
+
+    Captures predecessor tool names, count, and approximate payload size so
+    callers can log this context if the continuation send fails.
+    """
+    results = list(tool_results or [])
+    names = [getattr(r, "name", None) for r in results]
+    payload_chars = 0
+    for r in results:
+        content = getattr(r, "content", None)
+        if isinstance(content, str):
+            payload_chars += len(content)
+        elif isinstance(content, dict):
+            try:
+                payload_chars += len(json.dumps(content))
+            except Exception:
+                payload_chars += len(str(content))
+        elif content is not None:
+            payload_chars += len(str(content))
+    return {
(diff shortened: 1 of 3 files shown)


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. The post-tool continuation work here has drifted and collides with later continuation/turn-path changes; it is not being carried forward from this stale branch. If needed, it should be reworked fresh against current main.

Issue #258, opened 2026-06-12
Define recovery contract for side-effecting MCP operations


TZZheng:
## What's wrong

`Lingtai-AI/lingtai#188` reported a broad action-integrity failure mode: a side-effecting tool call can succeed, then the post-tool LLM continuation/recovery path can leave the agent with ambiguous state or stale notification previews, making a retry of a non-idempotent operation look reasonable.

The highest-risk kernel bugs from that report have since been addressed by narrower fixes:

- `Lingtai-AI/lingtai-kernel#98` / commit `58a9932`: preserve real executed tool results after continuation send failure.
- `Lingtai-AI/lingtai-kernel#157`: pair orphan Responses tool calls and preserve real `tc_wake` payloads when legacy send fails.
- `Lingtai-AI/lingtai-kernel#231`: guard stale replace-only notification-channel dismissals.

This follow-up is for the remaining product/runtime boundary: side-effecting MCP operations (`send`, `reply`, `delete`, `edit`, writes, etc.) still need an explicit recovery contract so future providers/addons do not accidentally reintroduce duplicate side effects or stale producer state.

## Where

- Component: kernel runtime recovery / MCP tool execution semantics / notification producer sync
- Related umbrella: `Lingtai-AI/lingtai#188`
- Related fixes: `Lingtai-AI/lingtai-kernel#98`, `#157`, `#231`

## Remaining scope

Define and implement a narrow contract for side-effecting tool recovery. Possible pieces:

1. **Classify side-effecting tools**
   - The recovery surface should know when a tool operation is non-idempotent or externally mutating.
   - Examples: chat/email `send`, `reply`, `delete`, `edit`; filesystem writes; future MCP write actions.

2. **Preserve last-known successful result in recovery surfaces**
   - When the tool handler completed and the result is known, recovery should clearly say that the tool completed and should not be retried blindly.
   - This is distinct from “handler did not run” and “handler started but completion is unknown”.

3. **Producer notification refresh / invalidation after known side effects**
   - If a producer-side action succeeds, stale notification previews for the affected producer should not remain in a state that invites the same operation again.
   - A producer-specific refresh/invalidation hook may be enough; this issue does not require a universal UI redesign.

4. **Optional operation ledger / idempotency key for MCP side effects**
   - For producers that support idempotency or stable operation IDs, record enough metadata to avoid duplicate sends/deletes/edits during recovery.
[...]

PR #307, opened 2026-06-16
docs: define MCP side-effect recovery contract

huangzesen:
## Summary

- add a design note for the remaining side-effecting MCP recovery contract from #258
- define completion states for non-idempotent external mutations after continuation failure
- document runtime/addon obligations for preserving known results, discouraging blind retry, refreshing notification mirrors, and optionally recording operation ledger metadata
- include a focused future test checklist without implementing the broader runtime contract in this PR

## Scope

This is intentionally docs/spec only. It does **not** rework the already-fixed tool-result restoration path from #98/#157, and it does not attempt to make every producer idempotent. The goal is to give kernel runtime work and MCP addon authors a shared contract before broader implementation.

Refs #258.
Refs Lingtai-AI/lingtai#188.

## Validation

- `git diff --check`
- Markdown structural sanity check for required sections/terms

### DIFF
--- a/discussions/mcp-side-effect-recovery-contract.md
+++ b/discussions/mcp-side-effect-recovery-contract.md
@@ -1,0 +1,150 @@
+# MCP side-effect recovery contract
+
+> **Status:** Design note for `Lingtai-AI/lingtai-kernel#258`; pre-implementation contract.
+> **Related umbrella:** `Lingtai-AI/lingtai#188`.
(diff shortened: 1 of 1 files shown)


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. This discussion-contract content was reviewed during the consolidated sweep, but is not being carried forward as standalone work. If the contract becomes needed again, it should be reopened as a fresh current design.

Issue #269, opened 2026-06-13
Support domain agent distributions with full resident prompt override and capability defaults


huangzesen:
## What's wrong

From the perspective of building a domain-specific `spedas_agent` on top of LingTai, the kernel/runtime currently makes it harder than necessary to provide a clean replacement agent class/distribution. The main blocker is that LingTai's resident system prompt/worldview and some capability defaults are not exposed as one stable distribution-level contract.

For SPEDAS we do **not** want to fork the runtime. We want a thin `SpedasAgent` / SPEDAS distribution that keeps LingTai's useful runtime primitives (tool loop, daemon, memory/molt, mailbox, notifications, identity, durable state) while replacing the default resident narrative prompt and adjusting default capabilities:

- replace LingTai's resident narrative prompt with a SPEDAS virtual-scientist prompt;
- keep runtime-dynamic appendices such as tool schemas, skills/knowledge catalogs, identity, character, and pad;
- disable `avatar` by default for SPEDAS (daemon workflows are sufficient for scientific parallelism);
- keep `soul` available but default its delay to an effectively-off value such as `99999999`;
- allow a blank `covenant` seed rather than forcing a replacement covenant text;
- seed `character`/`pad` on create, but do not refresh-overwrite them afterward.

In short: domain agent classes need a first-class "resident prompt + default capability/profile" override surface, instead of patching `init.json` and `system/*.md` around packaged defaults.

## Where

- Component: LingTai kernel/runtime prompt construction and agent initialization
- Suggested target repo: `Lingtai-AI/lingtai-kernel`
- Relevant files observed locally:
  - `src/lingtai_kernel/prompt.py`
  - `src/lingtai_kernel/base_agent/prompt.py`
  - `src/lingtai_kernel/base_agent/tools.py`
  - `src/lingtai/agent.py`
  - `src/lingtai/cli.py`
  - `src/lingtai/init_schema.py`
  - packaged prompts: `src/lingtai/prompts/substrate.md`, `src/lingtai/prompts/procedures.md`

## Reproduction / evidence

This came from auditing the prompt path while designing a SPEDAS-native agent distribution.

Observed prompt composition order:

```text
principle -> covenant -> tools -> substrate -> procedures -> comment -> rules -> brief -> skills -> knowledge -> identity -> character -> pad
```

Evidence in `src/lingtai_kernel/prompt.py`:

```text
98:     _DEFAULT_ORDER = [
100:        "principle",
101:        "covenant",
[...]

PR #308, opened 2026-06-16
Wire init prompt fields into resident base prompt

huangzesen:
## Summary

This PR wires the existing top-level `prompt` / `prompt_file` init fields into the resident **base prompt** path:

- `BaseAgent` now stores an optional `_base_prompt` from init construction.
- Both string and batched system prompt builders pass `_base_prompt` into `build_system_prompt(..., base_prompt=...)`.
- The wrapper `Agent` refresh path resolves `prompt_file` and refreshes `_base_prompt` before rendering dynamic sections.
- The CLI boot path passes the resolved inline prompt into the initial Agent construction, while `_setup_from_init()` remains the canonical full setup/refresh path.
- Tests cover inline prompt, `prompt_file`, empty prompt compatibility, and batched prompt rendering.

Refs #269.

## Scope note

This is intentionally the smallest safe slice of #269. It activates the already-declared resident/base prompt surface while preserving dynamic appendices such as tools, identity, character, pad, and other prompt sections.

It does **not** close the larger distribution contract work from #269: first-class domain distributions, default capability profiles, default MCP catalogs, and broader fallback/seed semantics remain follow-up design work.

## Validation

- `git diff --check`
- `python -m pytest -q tests/test_deep_refresh.py tests/test_prompt.py tests/test_init_schema.py tests/test_agent.py` → 137 passed

## Review

Independent review artifacts were produced for DeepSeek, MiMo, GLM, and Codex; all approved the slice. After review, I also incorporated a nonblocking batched-builder test suggestion and kept `prompt` as the last keyword-only `BaseAgent.__init__` parameter to minimize API risk, then reran validation.


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. The resident/init prompt wiring has drifted relative to current main and should be reworked fresh if still needed; this stale branch is not being carried forward.

PR #312, opened 2026-06-16
test: cover preset channel MCP surface

huangzesen:
## Summary

- add a direct regression test for `Lingtai-AI/lingtai#235`
- verify that a saved LLM preset omitting `mcp`/`email`/`psyche` does not strand configured channel MCP tools after refresh
- assert the post-refresh surface still has `email`, `psyche`, `mcp`, the channel producer tool, and MCP retry bookkeeping

## Why

The issue described a human-channel stranding incident: notifications could still arrive, but a preset switch had dropped the producer tool needed to read/reply. Current kernel behavior already protects this path via core defaults (`mcp`), BaseAgent intrinsics (`email`/`psyche`), and init.json MCP reload. This test makes that contract explicit so future preset/materialization refactors do not regress it.

## Validation

```text
git diff --check
python -m pytest -q tests/test_preset_materialization.py tests/test_agent.py tests/test_mcp_capability.py
# 97 passed
```

Closes Lingtai-AI/lingtai#235.

### DIFF
--- a/tests/test_preset_materialization.py
+++ b/tests/test_preset_materialization.py
@@ -548,6 +548,100 @@ def test_refresh_preset_keeps_daemon_override_in_manager(tmp_path, monkeypatch):
     assert mgr._handle_list()["max_emanations"] == 30
 
 
+
+def test_refresh_preset_omitting_mcp_keeps_channel_reply_surface(tmp_path, monkeypatch):
+    """Regression for Lingtai-AI/lingtai#235.
+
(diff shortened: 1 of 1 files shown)


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. The one possibly useful residue is the preset-channel MCP test coverage; that is now tracked through the consolidated #545 review path rather than this stale source branch.

PR #310, opened 2026-06-16
fix(telegram): clear handled notification mirrors

huangzesen:
## Summary

Fixes stale Telegram notification mirror re-delivery after agents read/reply to Telegram messages.

`Lingtai-AI/lingtai#153` describes a loop where `.notification/mcp.telegram.json` survives across wake/molt boundaries and is re-injected as if it were new, even after `telegram(read)` and a reply. This PR ties the Telegram MCP notification mirror back to Telegram's durable producer read state.

## Changes

- Add generic Telegram LICC metadata on inbound events:
  - `platform=telegram`
  - `conversation_ref=<account>:<chat_id>`
  - `message_ref=<account>:<chat_id>:<message_id>` for ordinary message events
- After `telegram(read)`, clear `.notification/mcp.telegram.json` only when every preview message referenced by the current notification mirror is present in durable `telegram/{account}/read.json`.
- After a successful `telegram(reply)`, mark the replied-to message as read and run the same notification cleanup.
- Preserve safe fallback for old notification mirrors by parsing existing `#account:chat:message` conversation-preview anchors.
- Stay conservative on ambiguity:
  - malformed/identity-less preview → keep notification;
  - mixed coalesced notification with unread message(s) → keep notification;
  - callback-query previews → keep notification, because Telegram callback updates reuse the inline-keyboard message id and are not uniquely represented by `account:chat:message`.

## Tests

```bash
git diff --check
python -m pytest -q \
  tests/test_telegram_notification_read_state.py \
  tests/test_mcp_inbox.py \
  tests/test_mcp_licc_client.py \
  tests/test_meta_block.py \
  tests/test_notification_sync.py \
  tests/test_system_dismiss.py \
  tests/test_agent.py \
  tests/test_deep_refresh.py
```

Result: `233 passed`.

## Review notes

- DeepSeek, MiMo, GLM: APPROVE on the main read-state cleanup patch.
- Codex parent review: APPROVE on the final patch including the callback-query safety addendum.

Fixes Lingtai-AI/lingtai#153

PR #314, opened 2026-06-16
fix(telegram): reconcile notification mirror from durable unread (#111)

huangzesen:
# fix(telegram): reconcile notification mirror from durable unread (#111)

Closes Lingtai-AI/lingtai#111.

## Problem

`.notification/mcp.telegram.json` delivery is **edge-triggered**: the kernel MCP-inbox poller publishes the mirror only when it consumes a *new* inbound LICC event. PR #31 (preserve `.notification/` across molt) and PR #190 (post-molt IDLE re-inject + wake) fixed the case where a mirror **is present** at molt completion — but neither changed the edge-triggered-only delivery contract.

So if the mirror is **absent or already-dismissed** at the moment a context molt completes, while durable read state (`read.json`) still marks Telegram messages unread, **nothing re-derives the notification** until a *fresh* inbound message creates a new edge. The post-molt agent reads its summary, sees no pending notification, and idles — exactly the stall in #111, where the user's `hello?` was the workaround that created the new edge.

## Fix

A minimal, producer-side **level-reconcile** in the Telegram MCP (no kernel changes):

- **`TelegramManager.reconcile_notifications()`** — when durable unread (inbox minus `read.json`) is non-empty **and** no live mirror exists, re-emit a LICC event via the existing `on_inbound` path so the kernel republishes the mirror through its normal channel and `_sync_notifications` (PR #190) re-injects + wakes the post-molt agent **without a second inbound message**.
  - A present mirror short-circuits the tick (idempotent — the mirror is a single per-channel file shared by all accounts).
  - All-read is a no-op: reconcile **never** clears and **never** republishes a stale mirror after the producer has read.
- **`on_idle_tick`** — `TelegramAccount`/`TelegramService` gain an optional per-poll-cycle callback (fired even on no-update long-poll returns), wired in `build_manager()` to `reconcile_notifications()`. This runs independently of new inbound messages — and therefore independently of molts — converting delivery to **edge-triggered + periodic level-reconcile**.

Re-emitting a LICC event (instead of writing `.notification/` directly) keeps the LICC contract intact and reuses the kernel's existing coalescing/publish/fingerprint machinery — no duplicate notification logic, no read-state schema leaking into the kernel.

## Composes with PR #310 (stale-mirror clear, #153)

This is the complementary half of one reconcile:

- #310 **clears** the mirror once all previewed messages are read.
- This PR **republishes** the mirror from still-unread state when it is missing.

Conditions are non-overlapping (all-read vs. unread-with-no-mirror), so reconcile never pre-empts #310's clear, and #310 may clear freely — the next idle tick republishes if unread remains. This closes gap #3 from the investigation (aggressive clearing widening #111). PR #310 is **not** required as a base for this change; this builds directly on `main`.

## Tests

- `tests/test_telegram_notification_reconcile.py` — unread→republish, all-read no-op, mirror-present no-op, empty-inbox no-op, multi-account single-channel.
- `tests/test_telegram_idle_tick.py` — poll loop fires the tick per cycle; raising tick is non-fatal; callback optional.
- `tests/test_telegram_service_idle_tick.py` — service forwards the tick to every account.

```
new tests:                                                10 passed
test_mcp_inbox / notification_sync / molt_persistence /
  post_molt_notification / mcp_licc_client:              113 passed
test_sent_message_tracker / system_dismiss /
  mcp_capability (regression guard):                      84 passed
git diff --check:                                          clean
```

Developed test-first (TDD). +139 lines, additive, no deletions, no kernel edits.

### DIFF
--- a/src/lingtai/mcp_servers/telegram/account.py
+++ b/src/lingtai/mcp_servers/telegram/account.py
@@ -143,6 +143,7 @@ class TelegramAccount:
         on_message: Callable[[str, dict], None] | None = None,
         state_dir: Path | None = None,
         commands: list[dict[str, str]] | None = None,
+        on_idle_tick: Callable[[], None] | None = None,
     ) -> None:
         self.alias = alias
         self._bot_token = bot_token
@@ -150,6 +151,10 @@ class TelegramAccount:
         self._poll_interval = poll_interval
         self._on_message = on_message
         self._state_dir = state_dir
+        # Fired once per poll cycle (including no-update long-poll returns) so
+        # the host can run periodic, message-independent work — e.g. the
+        # notification reconcile that recovers the post-molt stall (issue #111).
+        self._on_idle_tick = on_idle_tick
         # If commands is None, fall back to DEFAULT_COMMANDS at registration
         # time. An explicit empty list means "register no commands" and is
         # respected (Telegram clears the menu).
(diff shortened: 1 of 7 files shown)


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. This Telegram notification-reconcile branch is stale/ambiguous and overlaps later Telegram review work; it is not being carried forward.

PR #293, opened 2026-06-15
Add pre-send context hard gate

TZZheng:
## Summary

Addresses #193 by adding a pre-send hard context gate in `SessionManager.send()` so the runtime estimates the canonical wire plus the pending message before any provider dispatch. If the estimated turn is over the configured hard ceiling, the runtime now tries bounded local recovery and otherwise returns a local terminal assistant response without logging `llm_call` or calling the provider.

## Root cause

The existing pressure handling was warning/reactive oriented: the agent could receive context pressure warnings or recover from provider overflow after the fact, but there was no central hard stop immediately before `session.send(...)` that accounted for the pending text/tool-result payload about to be appended. That left request, wake, and tool-loop continuations able to attempt provider calls that were already estimated to be over budget.

## What changed

- Added `AgentConfig.context_hard_pressure` with a default of `0.98`.
- Added a pre-send gate in `SessionManager.send()` after prompt/tool refresh and health checks, but before `llm_call` logging and provider dispatch.
- Estimated pending payloads for user text, wire-drive `None`, and completed tool-result lists.
- Added a bounded local recovery ladder:
  1. one deterministic oversized-history compaction attempt;
  2. a forced `context_forget(source="hard_ceiling")` only for replayable text turns;
  3. a local terminal block if the turn still does not fit.
- Preserved completed tool results when a post-tool continuation is locally blocked, then rebuilt provider session state from canonical history so tools are not blindly re-run and server-side sessions do not continue from stale state.
- Updated OpenAI Responses server-side sessions to maintain a local canonical transcript while still using `previous_response_id` for provider-side state, so pre-send estimates include chained history hidden on the provider and rebuilt sessions seed valid assistant function-call / tool-result pairs from canonical history.
- Wired `context_hard_pressure` through init validation, preset validation/materialization, and live preset activation. Preset swaps clear stale root hard-pressure values when the target preset omits the field, so behavior is not history-dependent.
- Added a self-contained HTML explainer under `reports/pr293-pre-send-hard-gate-explainer.html` for reviewer/human context.

## Non-goals

- This does not replace provider-specific overflow recovery; provider rejections can still happen if local estimates differ from provider accounting.
- This does not add a new wake notification on hard-gate block; the active turn receives a local assistant response.
- This does not broaden context compaction beyond the existing deterministic oversized-tool-result artifact path.
- This does not change warning-ladder UX except by adding the central pre-send hard stop.

## Validation

- `git diff --check`
- `python3 -m py_compile` on all changed Python source/test files
[...]

PR #316, opened 2026-06-16
Cap over-ceiling context banner and type overflow events

### LINKED ISSUE #193: Context pressure can reach ~2x budget while runtime still sends ~400k-token LLM calls
## Summary

Trajectory mining found an agent reaching almost **2× its context budget** before molting/compacting, while the runtime continued to send ~400k-token LLM calls and injected a confusing `context: 197.2%` banner into the model's own prompt.

This is a context-management bug even though the observed provider accepted the oversized request: it is costly/noisy on long-context providers and unsafe on stricter context-capped models.

## Verified evidence

Agent/event log:

`/Users/huangzesen/work/projects/lingtai-rescue/.lingtai/mimo-1/logs/events.jsonl`

At line 28317:

```json
{"type":"molt_hard_ceiling_warning","pressure":1.972005,"ceiling":0.95}
```

Immediately after, the runtime sent the oversized context into the model:

- line 28318 `text_input` starts with:
  - `[Current time: 2026-05-08T12:42:38-07:00 | context: 197.2% (sys 48489 + ctx 345912)]`
- line 28319 `llm_call`
- line 28321 `llm_response` reports:
  - `input_tokens=394941`
  - `cached_tokens=394368`
  - `output_tokens=210`

A nearby later request was even larger:

- line 28368 `llm_call`
- line 28369 `llm_response` reports:
  - `input_tokens=407397`
  - `cached_tokens=406080`
  - `output_tokens=1070`

Shard analysis found repeated ceiling warnings around this period, with pressure around 1.95–1.97 for several minutes. The existing `auto_forget` / hard-ceiling path did not reduce pressure below the ceiling before further LLM calls were issued.

## Impact

- **Cost/throughput:** repeated ~400k-token turns (mostly cached, but still large requests).
- **Model confusion:** the agent reads `context: 197.2%` inside its own prompt, which looks like a broken invariant.
- **Provider portability:** this only succeeded because the provider/model tolerated a very large context. The same path on a strict 200k model would likely fail with context-length/413 errors and can stall the agent.
- **Observability:** there is no distinct typed event for provider-side context overflow in the observed taxonomy; an actual overflow would likely appear as a generic LLM/tool error rather than a searchable `context_length_exceeded`-style event.

## Expected behavior

Once context pressure crosses the hard ceiling, the runtime should not continue issuing full LLM calls indefinitely. Possible fixes:

1. Make the hard ceiling blocking: force a synchronous compaction/molt before issuing another `llm_call` when pressure is at/above the ceiling.
2. Size `auto_forget` / compaction so one pass actually drives pressure back below the ceiling, including cases where cached context re-enters.
3. Cap or reword the prompt banner above 100% (for example `context: over hard ceiling; compacting`) so the model is not fed a nonsensical `197%` status.
4. Add typed provider overflow events (`context_length_exceeded`, `provider_413`, etc.) so real hard-cap failures are diagnosable.

## Notes

This was found during a read-only trajectory scan. The exact log paths are local to this machine, but the event sequence demonstrates a systemic failure mode in the context-pressure/molt pipeline.

huangzesen:
## Summary

Addresses the residual banner/typed-event portions of Lingtai-AI/lingtai-kernel#193 (items 3 and 4). This **complements #293**, which covers the core pre-send hard gate; this PR does **not** subsume #293, and #193 should remain open until both land and retest.

> Note: this PR intentionally does not auto-close #193 — #193 also depends on #293.

## Changes

**Item 3 — cap / reword the over-ceiling context banner**
- `meta_block._render_context_fragment` no longer feeds a runaway percentage (e.g. `context: 197.2%`) into the model's own prompt.
- When usage > 1.0 it renders a bounded over-ceiling notice via the new i18n key `system.context_over_ceiling` (en/zh/wen), preserving the sys/ctx token breakdown.
- `usage == 1.0` still uses the normal breakdown.

**Item 4 — typed provider-overflow event**
- Adds `_classify_over_window_event` (`provider_413` / `context_length_exceeded` / `context_overflow`).
- Emits a distinct typed `provider_context_overflow` event alongside the existing generic `aed_over_window_detected` marker, so real provider hard-cap failures are diagnosable/grep-able in `events.jsonl`.
- Back-compat preserved: the generic marker still fires.

## Relationship to #293
- **Zero file overlap** with #293. #293 covers the core pre-send hard gate on the send path; this PR covers the residual items 3 and 4 (banner cap + typed event).
- The two are independent and can land in either order.

## Tests
- Implementation run: full suite **2256 passed / 3 skipped**.
- Review reran **100 focused + 7 new + 214 adjacent** tests — all passed.
- New coverage: +3 `meta_block` cases (>100% cap, ==100% breakdown, zh), +3 classifier unit cases, +1 emission integration case.

### DIFF
--- a/src/lingtai_kernel/base_agent/turn.py
+++ b/src/lingtai_kernel/base_agent/turn.py
@@ -334,6 +334,29 @@ def _is_over_window_error(exc: Exception) -> bool:
     return any(fragment in msg for fragment in _OVER_WINDOW_MSG_FRAGMENTS)
 
 
+def _classify_over_window_event(exc: Exception) -> str:
+    """Bucket an over-window error into a stable, grep-able event ``kind``.
+
+    Issue #193 item 4: ``_is_over_window_error`` only answers yes/no, and the
+    generic ``aed_over_window_detected`` marker hides *why* the wire was
+    rejected.  A distinct typed ``kind`` lets operators tell a real provider
+    hard-cap rejection (HTTP 413) and the canonical
+    ``context_length_exceeded`` provider code apart from a merely
+    estimate/phrasing-driven overflow, so genuine hard-cap failures are
+    diagnosable in ``events.jsonl``.
+
+    Precedence is provider-truth first: an explicit 413 status is the most
+    authoritative signal, then the canonical ``context_length_exceeded``
+    fragment, then a generic ``context_overflow`` fallback.
+    """
+    if _exception_status_code(exc) == 413:
+        return "provider_413"
+    msg = (str(exc) or "").lower()
+    if "context_length_exceeded" in msg:
+        return "context_length_exceeded"
+    return "context_overflow"
+
+
 def _compact_history_before_retry(agent, *, source: str) -> "CompactionStats | None":
     """Retroactively spill oversized tool results before an AED retry.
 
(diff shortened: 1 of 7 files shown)


huangzesen:
Closing as part of Jason's authorized stale-PR cleanup. This over-ceiling/overflow work has drifted and collides with later turn-path/continuation changes; it is not being carried forward from this stale branch.

PR #386, opened 2026-06-19
Add fail-open cache audit ledger

runyuan-wang:
## Summary

Adds a fail-open `CacheAuditor` observation layer for main-agent LLM calls.

This PR is intentionally PR-0: it does **not** change prompts, provider behavior, prompt batching, or cache-control semantics. It only creates a safe JSONL usage ledger so later low-power/cache PRs can be measured against real data instead of intuition.

Changes:
- add `src/lingtai/kernel/cache_audit.py` with fail-open JSONL recording and `batch_hash()`
- wire `SessionManager` to optionally construct the auditor behind a fail-open guard
- wire `BaseAgent` to pass the live `self._session` a `logs/cache_audit.jsonl` path
- record main-call usage after normal token accounting
- add tests for writer behavior, constructor/write fail-open, SessionManager JSONL output, and BaseAgent wiring
- update `src/lingtai/kernel/ANATOMY.md`

## Why

The low-power work needs a baseline ledger before changing prompt/cache behavior. This PR is the meter, not the power-saving device.

Guardrails:
- fail-open: audit failures must never block agent startup or LLM calls
- no prompt/provider behavior changes
- no claim that this PR saves tokens by itself
- `uncached_input` is a coarse normalized field; for Anthropic it may include cache-write tokens because `UsageMetadata` currently preserves total input and cache-read only

## Validation

Passed in a clean worktree based on `origin/main`:

```text
pytest tests/test_cache_audit.py \
  tests/test_session.py::test_send_writes_cache_audit_jsonl \
  tests/test_session.py::test_session_manager_cache_auditor_constructor_is_fail_open \
  tests/test_agent.py::test_base_agent_wires_cache_auditor_to_real_session -q

8 passed in 1.20s
```

```text
pytest tests/test_cache_audit.py tests/test_session.py tests/test_agent.py -q

93 passed in 10.00s
```

```text
git diff --check
python -m py_compile src/lingtai/kernel/cache_audit.py src/lingtai/kernel/session.py src/lingtai/kernel/base_agent/__init__.py
```

Full-suite note:

```text
pytest -q

2995 passed, 7 skipped, 3 failed
```

The 3 failures appear unrelated to this PR:
- `tests/test_codex_prompt_cache_key.py::test_codex_request_includes_default_prompt_cache_key` expects `lingtai-codex:gpt-5.5:v1`, but current `origin/main` emits `lingtai-codex:gpt-5.5:init-<hash>:v2`.
- two `tests/test_kernel_isolation.py` subprocess import checks failed without `PYTHONPATH=src`; rerunning those with `PYTHONPATH=src` passed.

## Review notes

This PR deliberately leaves later work out of scope:
- daemon usage audit
- baseline harness / per-call comparison report
- `_BATCHES` split / volatile pad isolation
- provider-specific prompt cache validation
- daemon task capsule


runyuan-wang:
Closing this PR after review: LingTai already has token accounting through /kanban backed by logs/token_ledger.jsonl, including input/output/thinking/cached/model/endpoint, provider grouping, recent calls, and cache hit rate. This PR's separate cache_audit.jsonl duplicates that ledger, and its batch_hashes are not populated yet, so it does not provide the needed cache-attribution value in its current form. The token-saving change has been split into PR #387, which is independent of this PR.

PR #388, opened 2026-06-19
Add structured daemon task capsules

runyuan-wang:
## Scope

This PR is scoped to daemon token-efficiency work.

It does not add a new product workflow, a biomedical demo, a drift detector, or a success-string protocol. It only adds a small structured contract surface so daemon runs can be constrained, reviewed, and measured more cleanly when experimenting with lower parent-context / daemon-first execution.

It does **not** by itself change default scheduling, lower main-agent API calls, or prove stable token-efficiency gains. It is a prerequisite contract layer for later daemon-first scheduling work.

## Summary

Adds an optional structured `capsule` object to daemon tasks. The capsule is rendered before the free-form task text and is persisted with the daemon run.

The goal is to make token-efficiency experiments more disciplined: parent agents can hand daemons a compact objective, scope, result contract, review gate, and metadata instead of relying only on long free-form prompts. This can reduce noisy back-and-forth and make returned work easier to audit, while keeping the parent agent responsible for quality decisions.

## Roadmap alignment

The token-efficiency roadmap has two levels:

1. **Result goal**: improve token efficiency by making more stable prefix reusable and by letting lightweight daemon contexts hit cache more often.
2. **Means**: reduce unnecessary main-agent API calls and increase bounded daemon API calls, so the parent agent focuses on framing, review, synthesis, and human-facing communication.

In that sequence, PR1a addressed stable prompt / pad-tail cache behavior, PR2 adds the `/kanban` call-ratio dashboard, and this PR is best understood as **PR3a**: the contract surface needed before a later scheduling PR can safely push more execution work to daemons.

## Design note: ring-return loop

This PR also gives an engineering form to the ring-return idea behind this experiment:

1. **Dispatch with a contract**: the parent sends the daemon out with a compact capsule instead of a loose, ever-growing context tail.
2. **Return with structure**: the daemon is asked to bring back evidence, boundaries, artifacts, and review notes in the shape requested by `result_contract`.
3. **Merge only after review**: `review_gate` marks what still needs tests, evidence checks, or parent review before the result is allowed to influence the main task.

In token-efficiency terms, this loop is meant to reduce the parent agent's reconstruction cost: work can branch out to a daemon, return in an auditable form, and then be folded back into the main context only after review. The PR keeps this as a modest daemon contract surface rather than a broad architecture change.

## Architecture philosophy: human judgment completes system mechanics

This contract layer is also guided by a simple architecture philosophy, expressed by the author as `以人心补天心`: system mechanisms can provide execution, caching, routing, and scale, but human or parent-agent judgment still supplies direction, taste, responsibility, and final acceptance.

For token efficiency, that distinction matters. The goal is to remove wasteful context reconstruction, not to remove judgment. Daemons can carry out bounded work and return structured evidence; the parent remains responsible for deciding what should be trusted, refined, persisted, or discarded.

## What changed

- Add optional `tasks[].capsule` to the daemon tool schema.
- Validate capsules before dispatch:
  - missing or `{}` is treated as no capsule;
  - non-object capsules fail with `capsule must be an object`;
  - non-JSON-serializable capsule values fail with `capsule must be JSON-serializable`.
- Render the capsule into daemon one-shot context before the task body for both LingTai and CLI backends.
- Persist the capsule into `daemon.json` under `call_parameters.capsule`.
- Preflight all LingTai backend task contexts before executor creation so an invalid later task does not partially dispatch earlier tasks.
- Add tests for schema exposure, rendering order, persistence, validation, and preflight behavior.
- Update daemon anatomy and i18n strings.

## Token-efficiency context

The current token-efficiency problem is not just "make prompts shorter". It is also making daemon work return with enough structure that the parent does not need to spend extra context reconstructing what happened.

This PR addresses that by adding a compact contract layer:

- `objective`: what the daemon is trying to answer;
- `scope`: what is in / out of bounds;
- `result_contract`: what evidence and outputs should come back;
- `review_gate`: what must still be checked by tests, evidence, or parent review;
- `metadata`: experiment labels or audit hints.

It intentionally does not change scheduler behavior, watchdog behavior, token ledger semantics, daemon result files, or lifecycle operations.

## Local experiment notes

A local A/B review experiment was run against this patch:

- Plain daemon review: 269.229s elapsed, 613,233 input tokens, 7,138 output tokens.
- Capsule-guided daemon review: 193.714s elapsed, 467,059 input tokens, 5,291 output tokens.

The structured run returned a more directly reviewable report and found a real blocker: invalid later capsules could partially dispatch earlier LingTai backend tasks. That blocker is fixed here by preflighting all task contexts before submission.

This is a single local experiment, so the PR does **not** claim stable token reduction. It adds the structured contract surface needed for daemon-first token-efficiency experiments to be measured and reviewed more cleanly.

## Validation

```bash
uv run pytest tests/test_daemon.py tests/test_daemon_run_dir.py tests/test_communication_bundle_bridge.py -q
uv run python -m compileall -q src/lingtai/core/daemon/__init__.py
uv run python -m json.tool src/lingtai/i18n/en.json >/dev/null
uv run python -m json.tool src/lingtai/i18n/zh.json >/dev/null
uv run python -m json.tool src/lingtai/i18n/wen.json >/dev/null
git diff --check
```

Result: `145 passed`.


huangzesen:
Thanks for the structured daemon capsule proposal.

Jason reviewed this direction and decided not to take it: LingTai should not move toward structured meta-prompting / capsule contracts for daemon tasks. The project direction is to keep the parent agent responsible for framing, review, synthesis, and memory deposition, and to learn daemon-use habits from human practice rather than encode another meta-prompting layer into the daemon API.

We have taken the narrower design lesson into #462, which documents learning from human daemon/subagent usage patterns and depositing those lessons into layered memory (pad / lingtai / knowledge / skills), without adding a capsule contract.

Closing this PR as rejected / not planned.

PR #389, opened 2026-06-19
Document daemon-first policy and capsule handoffs

runyuan-wang:
## Scope

Draft / stacked PR on top of #388 (`Add structured daemon task capsules`).

This PR documents the daemon-first policy layer that uses the capsule contract surface from #388. It does not change daemon scheduling, tool execution, lifecycle behavior, or token ledger semantics.

## Why

The token-efficiency roadmap has two levels:

1. **Result goal:** improve token efficiency by making more stable prefix reusable and by letting lightweight daemon contexts hit cache more often.
2. **Means:** reduce unnecessary main-agent API calls and increase bounded daemon API calls, so the parent agent focuses on framing, review, synthesis, and human-facing communication.

This PR addresses the policy layer for that second goal: it teaches agents to consider daemon-first delegation for execution-heavy, auditable work, while keeping final judgment with the parent.

## What changed

- Update resident procedures (`src/lingtai/prompts/procedures.md`) with a compact daemon-first split:
  - parent frames / reviews / synthesizes;
  - daemons execute bounded, evidence-returning work;
  - final judgment, safety decisions, external side effects, and human-facing wording stay with the parent.
- Expand `procedures-manual` with:
  - daemon-first candidates such as multi-file scans, code review, test runs, evidence gathering, large local search, batch conversion, and report assembly;
  - an anti-autopilot guard: short deterministic commands, human clarification, and judgment/aesthetic choices stay in the parent context;
  - a capsule-shaped handoff template with objective, scope, result contract, and review gate.
- Update `daemon-manual` to document `capsule` as a structured handoff/review artifact:
  - JSON-serializable objective, scope, result contract, review gate, and metadata;
  - explicit statement that capsule does not replace parent verification or make a daemon autonomous.
- Add tests in `tests/test_skills.py` to keep the resident and expanded policy text present.

## Architecture philosophy

This is the policy form of `以人心补天心`: system mechanics can execute, cache, route, and scale; human or parent-agent judgment still supplies direction, taste, responsibility, and final acceptance.

For token efficiency, the goal is to remove avoidable main-context noise and reconstruction work, not to remove judgment.

## Validation

Delegated daemon review:

- `em-7` completed test additions and reported `uv run pytest tests/test_skills.py -q` as `23 passed` plus `git diff --check` clean.
- `em-8` performed read-only policy review and reported **PASS, no blockers**.

Parent validation:

```bash
uv run pytest tests/test_skills.py tests/test_deep_refresh.py -q
# 42 passed in 2.83s

[...]

PR #390, opened 2026-06-19
Add terminal async daemon dispatch

runyuan-wang:
## Summary

This PR adds a narrow terminal async dispatch path for successful `daemon(action="emanate")` calls.

When the parent model has already requested bounded daemon delegation and the daemon manager returns a deterministic `dispatched` acknowledgement with run ids and a group id, the parent turn loop now commits the paired tool result and ends the turn instead of spending another parent-model continuation merely to acknowledge launch.

Daemon completion still returns through the existing notification and review path. This PR does **not** generate a final answer automatically and does **not** remove parent review.

## What changed

- `DaemonManager` marks successful LingTai and CLI `emanate` dispatch results with `terminal_async_dispatch: True`.
- `ToolExecutor` gains an opt-in `allow_terminal_async_dispatch` mode.
- The terminal async path is fail-closed and daemon-specific. It requires all of:
  - a single tool call in the batch;
  - tool name `daemon`;
  - `action="emanate"`;
  - `status="dispatched"`;
  - `terminal_async_dispatch is True`;
  - non-empty `ids`;
  - present `group_id`.
- Parent turn executors opt in; daemon-internal executors do not inherit the optimization.
- A `tool_call_terminal_async_dispatch` lifecycle event records the skipped launch-continuation boundary.
- Tests cover executor-level and turn-loop behavior, malformed result shapes, opt-in boundaries, and daemon dispatch metadata.

## Why this is intentionally narrow

This is a deterministic launch-ack optimization, not a general terminal-tool protocol.

It is meant to avoid the redundant continuation that would usually say “daemon(s) launched; I will review when they finish.” It does not apply to daemon `check/list/ask/reclaim`, failed dispatches, non-daemon tools, multi-tool batches, or any case needing semantic interpretation by the parent model.

The design follows the daemon-first/capsule path from #388 and #389: parent frames and delegates bounded execution; daemon work runs out-of-band; completion returns via notification/artifact/result contract; parent review remains the gate before human-facing conclusions or external actions.

## Measurement posture

This PR should be measured as avoided parent continuations per eligible daemon dispatch, using lifecycle/ledger traces and comparable task runs. It does not claim a fixed token-saving percentage. Savings depend on task shape, provider cache behavior, daemon output quality, and whether later review avoids reconstruction work.

## Validation

```bash
uv run pytest tests/test_tool_executor.py tests/test_terminal_async_daemon_turn.py tests/test_daemon.py tests/test_daemon_run_dir.py tests/test_communication_bundle_bridge.py -q
# 186 passed

uv run python -m compileall -q src/lingtai/kernel/tool_executor.py src/lingtai/kernel/base_agent/turn.py src/lingtai/core/daemon/__init__.py

git diff --check
```

## Stack note

Draft PR. This is the next token-efficiency step after:

- #388 `Add structured daemon task capsules`
- #389 `Document daemon-first policy and capsule handoffs`

It is easiest to review after those are reviewed/merged, because the safety argument depends on bounded daemon handoffs and parent review gates.


huangzesen:
Thanks for the proposal. After checking local trajectory data, the premise is real (the parent runtime currently resumes the model after daemon dispatch), but Jason decided the runtime optimization has limited practical value and should not proceed in this form.

We are closing this PR and switching to a documentation/manual update instead: strengthen `daemon-manual` to make the expected daemon notification contract and recommended waiting/polling discipline explicit:

- daemon terminal states report via notification;
- agents generally should not poll too frequently;
- prefer email-delay/self-reminder + idle for long waits;
- use bash sleep only as a secondary fallback;
- use `daemon ask` when adding requirements or follow-up instructions.

Closing as not planned / superseded by the daemon-manual documentation direction.

PR #468, opened 2026-06-22
Replace born/woken identity prose with molt lifecycle line

huangzesen:
## Why

The identity system-prompt section rendered:

> You were born `<created_at>`, woken `<started_at>` for this session.

`started_at` is **process-start wall-clock** — it regenerates on every `system.refresh` (the process relaunches). Rendering it into the system prompt shifted the cacheable prompt prefix on every refresh and **poisoned the Codex prompt cache** (the prefix is supposed to be stable so the cache hits). Jason also found the "born … woken …" wording weird.

## What

Stop putting any process-start / wake time in the system prompt. The lifecycle line now carries only **durable molt facts**:

- `You last molted at <last_molt_at> (molted N times so far).`
- `You have molted N times.` — when no timestamp is recorded yet
- `You have yet to molt.` — never molted

New persisted manifest field `last_molt_at` (UTC ISO-8601):
- set on each molt in **both** molt paths (`_context_molt`, `context_forget`),
- restored in the constructor from the prior manifest,

so the timestamp survives refresh/resume **without** leaking `started_at`. Birth (`created_at`) and `started_at` stay in `.agent.json` for diagnostics — they're just no longer rendered as prompt prose.

**Out of scope:** Codex reset / cache-buster behavior is intentionally untouched.

### Before → After (rendered)

```
- You were born 2026-01-01T09:00:00Z, woken 2026-06-22T08:30:11Z for this session.
- You have undergone 3 molts since birth.
+ You last molted at 2026-06-20T14:05:00Z (molted 3 times so far).
```

The "after" line is byte-identical across two processes that differ only in `started_at` — that's the property that keeps the cache prefix stable.

## Tests

- `tests/test_agent_preset_manifest.py` — identity section omits `started_at`, is invariant to it, renders molt wording (incl. singular + never-molted fallback), renders the `last_molt_at` line, and the manifest surfaces `last_molt_at`.
- `tests/test_psyche.py` — end-to-end: a real molt records `last_molt_at`, persists it to the manifest, and renders it in the identity section (and does not render `started_at`).

`PYTHONPATH=src pytest tests/test_agent_preset_manifest.py tests/test_psyche.py` → **55 passed**. ANATOMY updated for the manifest field + prose change.


huangzesen:
Closed per Jason's direction: keep the existing born/woken identity wording for now; the proposed last-molt prompt change and HTML remain exploratory artifacts. Codex cache reset strategy will be considered separately.

PR #470, opened 2026-06-22
Experiment with rolling Codex prompt cache keys

huangzesen:
## Summary
- add an experimental rolling Codex affinity mode for `prompt_cache_key` / `session_id` / `thread_id`
- first request keeps the existing stable per-agent key; later requests derive a new key from the previous API-call timestamp and the stable agent identity
- preserve the within-request invariant `prompt_cache_key == session_id == thread_id`; keep the old stable behavior behind a mode switch
- add focused regression coverage and update the OpenAI adapter ANATOMY notes

## Validation
- `python -m pytest tests/test_codex_rolling_prompt_key.py tests/test_codex_prompt_cache_key.py tests/test_codex_raw_reasoning_replay.py tests/test_openai_prompt_cache_key.py tests/test_openai_responses_streaming.py tests/test_openai_overflow_recovery.py tests/test_openai_compact_threshold.py tests/test_llm_adapter_timeouts.py tests/unit/auth/test_codex_auth.py -q` → 159 passed
- `python -m pytest tests/test_deep_refresh.py -k 'codex or refresh' -q` → 22 passed
- `git diff --check`

## Notes / risks
- This is intentionally an experiment requested by Jason; it may reduce cache hits if the backend prefers long-lived stable keys.
- The rolling timestamp is in-memory and per runtime process; refresh/restart resets the previous-call state.
- `x-client-request-id` remains the existing fresh per-request metadata value rather than being forced to match the rolling affinity key.
- Same official TUI multi-turn prompt-cache-key behavior remains unproven because local fake-terminal captures could not get a second TUI request to fire.

### DIFF
--- a/src/lingtai/llm/openai/ANATOMY.md
+++ b/src/lingtai/llm/openai/ANATOMY.md
@@ -92,11 +92,13 @@ Both paths return sessions wrapped via `_wrap_with_gate()` for rate limiting.
 
 ### Codex REST cache-affinity ids (`session_id` / `thread_id` / `prompt_cache_key`)
 
-**Codex-only.** The three cache-affinity values are a **single stable per-agent id**, byte-identical and NEVER changed for the life of the agent's identity:
+**Codex-only.** Within **every** request the three cache-affinity values are byte-identical:
 
 ```
-prompt_cache_key == session_id == thread_id == <stable per-agent id>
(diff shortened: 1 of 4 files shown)


huangzesen:
Closing this experimental PR after live testing showed rolling prompt/session IDs did not preserve cache reliably. Superseded by further investigation.

PR #575, opened 2026-06-29
fix(llm): normalize base_url trailing slash to prevent urljoin path stripping

BrianLiubr:
## Problem

When a preset's `base_url` contains a non-root path (e.g. `https://ark.cn-beijing.volces.com/api/coding`) but lacks a trailing slash, Python's `urljoin` treats the last segment as a *file* and silently strips it:

```python
urljoin('https://x.com/api/coding', '/v1/messages')
# → 'https://x.com/v1/messages'   ← /api/coding discarded!
```

This causes HTTP 404 for any provider whose endpoint sits behind a non-root path (Volcano Engine Ark, Azure OpenAI, etc.).

## Fix

In `LLMService.__init__`, detect `base_url`s with a path component but no trailing slash, and auto-append `/` before storing or passing to any adapter:

```python
if (base_url and isinstance(base_url, str)
    and "/" in base_url.split("://", 1)[-1]
    and not base_url.endswith("/")):
    base_url += "/"
```

This normalization runs once at construction time, before the value is stored or passed to any adapter. All downstream consumers (main adapter, on-demand adapters, cache keys) see the corrected URL.

## Tests

4 new test cases in `tests/test_llm_service.py`:

| Test | Input | Expected |
|---|---|---|
| `test_base_url_normalization_adds_trailing_slash` | `https://ark.example.com/api/coding` | `…/api/coding/` |
| `test_base_url_normalization_preserves_existing_slash` | `https://ark.example.com/api/coding/` | unchanged |
| `test_base_url_normalization_skips_root_urls` | `https://api.example.com` | unchanged |
| `test_base_url_normalization_handles_none` | `None` | `None` |

### DIFF
--- a/src/lingtai/llm/service.py
+++ b/src/lingtai/llm/service.py
@@ -207,6 +207,24 @@ class LLMService(LLMServiceABC):
         self._provider = provider.lower()
         self._model = model
         self._context_window = context_window
+        # Normalize base_url: if it contains a path component (e.g.
+        # ``/api/coding``) but lacks a trailing slash, append one.
+        # Python's ``urljoin`` treats the last segment as a *file* when
+        # there is no trailing slash, silently stripping it:
+        #   urljoin('https://x.com/api/coding', '/v1/messages')
+        #   → 'https://x.com/v1/messages'   ← wrong!
+        # With a trailing slash the path is preserved:
+        #   urljoin('https://x.com/api/coding/', '/v1/messages')
+        #   → 'https://x.com/api/coding/v1/messages'
+        # This affects any provider whose endpoint sits behind a non-root
+        # path (Volcano Engine Ark, Azure OpenAI, etc.).
+        if (
+            base_url
+            and isinstance(base_url, str)
+            and "/" in base_url.split("://", 1)[-1]
+            and not base_url.endswith("/")
+        ):
+            base_url += "/"
         self._base_url = base_url
         self._key_resolver = key_resolver or (lambda p: os.environ.get(f"{p.upper()}_API_KEY"))
         self._provider_defaults = provider_defaults or {}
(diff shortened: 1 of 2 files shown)


huangzesen:
Thanks for the PR. We need more evidence before accepting/merging this as-is.

The generic URL-merge hazard is real: if code combines a path-prefixed base URL with a relative path using generic URL resolution, a missing trailing slash can cause the final path segment to be treated as a file and replaced. However, after checking the current LingTai path, the adapters appear to already handle this case.

Local check against current `main`:

- generic `urljoin('https://x.com/api/coding', '/v1/messages')` demonstrates the abstract hazard, but a leading `/...` path replaces the base path regardless of trailing slash, so that example does not demonstrate the proposed trailing-slash fix;
- current LingTai `LLMService(provider='openai', base_url='http://127.0.0.1:<port>/api/coding')` was tested against a local HTTP server via `generate()`;
- the captured outgoing request path was `/api/coding/chat/completions`, so the `/api/coding` prefix was preserved.

Please provide a concrete LingTai reproduction showing a real outgoing request URL that is wrong today. The repro should include:

- the preset/provider/adapter involved;
- the `base_url` value used (redact secrets, of course);
- the expected outgoing URL;
- the actual outgoing URL produced by current `main`.

There is also a current validation blocker: `python -m pytest tests/test_llm_service.py -q` fails on this branch because existing direct-key tests expect the configured `base_url` value to be preserved, while this change stores/passes a normalized value with an added trailing slash.

If a concrete failing path exists, the fix should target that URL-building boundary (for example, a non-SDK URL builder) rather than changing `LLMService`'s global `base_url` storage/pass-through semantics without a current LingTai repro.

BrianLiubr:
Thanks for the thorough review — you're right, the bug doesn't exist in practice.

I verified with direct tests:

**urljoin (the abstract hazard):**
| base_url | path | result |
|---|---|---|
|  |  |  ❌ |
|  |  |  ❌ (same!) |
|  |  |  ❌ |
|  |  |  ✅ |

The trailing slash only helps with relative paths in urljoin — but the absolute-path case (which the PR comment used as example) is broken regardless.

**httpx (what the OpenAI SDK actually uses):**


All four cases produce the correct URL. The OpenAI Python SDK passes  to , which handles path joining correctly regardless of trailing slash.

**Conclusion:** The PR's fix targets a theoretical urljoin hazard that doesn't apply to the actual code path. The OpenAI SDK + httpx already handle this correctly. I'll close this PR.

PR #576, opened 2026-06-29
fix(nudge): prevent kernel-version nudge storm after agent dismissal

BrianLiubr:
## Problem

When the installed `lingtai` version differs from the running version, the `kernel_version` nudge check calls `upsert()` every 60 seconds without tracking whether the nudge had already been emitted.

If the agent dismissed the notification (deleting `nudge.json`), the next check cycle re-created the file, changing the notification fingerprint and triggering a new wake cycle. This created an endless **nudge storm** that:

- Consumed the agent's attention every ~60 seconds
- Prevented response to incoming emails (**17/20 emails unprocessed** in one observed incident on June 28, 2026)
- Trapped the agent in an `idle→active→idle` loop

## Root Cause

`src/lingtai_kernel/nudge/kernel_version.py` line 81-114: the version-mismatch path calls `upsert()` unconditionally on every check cycle (`_FAST_INTERVAL_SECONDS = 60`), with no persistent tracking of whether the nudge was already emitted for this version pair.

```python
# Before (buggy): upsert called every 60s, no emission tracking
if info.installed_version != info.running_version:
    upsert(agent, _KIND, { ... })  # Always writes nudge.json
    return
```

## Fix

Add persistent tracking (`emitted_for_mismatch`) so the mismatch nudge is emitted **once per version pair**. Clear the tracking when versions match again (e.g., after `system(action='refresh')`), so future mismatches still trigger the nudge.

```python
# After: skip re-emission for same version pair
if info.installed_version != info.running_version:
    mismatch_key = f"{info.running_version}->{info.installed_version}"
    persistent = _load_persistent_state(agent)
    kernel_state = persistent.setdefault(_KIND, {})
    if kernel_state.get("emitted_for_mismatch") == mismatch_key:
        return  # Already emitted for this pair
    kernel_state["emitted_for_mismatch"] = mismatch_key
    _save_persistent_state(agent, persistent)
    upsert(agent, _KIND, { ... })
    return
```

## Impact

- **+22 lines, 0 deletions** — minimal, surgical fix
- No API changes, no breaking behavior
- The nudge still appears once per version mismatch (agent gets reminded)
- After dismissal, the nudge does NOT reappear every 60 seconds
- After `refresh`, tracking is cleared so the next mismatch emits normally

### DIFF
--- a/src/lingtai/intrinsic_skills/system-manual/reference/runtime-update-checks/SKILL.md
+++ b/src/lingtai/intrinsic_skills/system-manual/reference/runtime-update-checks/SKILL.md
@@ -44,7 +44,7 @@ Common fields:
 | `installed` | Version visible from the installed `lingtai` distribution on disk. |
 | `latest` | Latest package version found by the daily packaged-runtime check, or `null` for local refresh nudges. |
 | `source` | `installed-distribution` for local refresh nudges, `pypi-json` for package-update nudges. |
-| `cadence` | `fast-local-check` or `at-most-once-per-utc-day`. |
+| `cadence` | `at-most-once-per-utc-day` for both local refresh and package-update nudges (re-emitted at most once per UTC day per version pair). |
 | `suggested_action` | The safe next-step hint; still verify context before acting. |
 
 After handling, clear the channel with:
--- a/src/lingtai_kernel/nudge/kernel_version.py
+++ b/src/lingtai_kernel/nudge/kernel_version.py
@@ -79,18 +79,25 @@ def check(agent) -> None:
         return
 
     if info.installed_version != info.running_version:
-        # Only emit the version-mismatch nudge once per version pair.
-        # Without this guard, the check re-emits every 60 seconds;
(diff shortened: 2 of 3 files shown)


huangzesen:
Maintainer update after clarifying the nudge cadence requirement: I added commit `916ffc93619c7c10a5f71a080ca446fee977083e` before merge so the local refresh mismatch nudge is at most once per UTC day per running->installed mismatch, while preserving the fast first detection.

Validation before merge:
- `git diff --check`
- `python -m pytest tests/test_kernel_version_nudge.py -q` -> 7 passed
- `python -m pytest tests/test_kernel_version_nudge.py tests/test_base_agent.py tests/test_source_drift.py tests/test_skills.py -q` -> 73 passed
- independent GLM-5.2 post-patch review: PASS / no blockers

Squash-merged as `5742d68d53b2132d0a43f2c7b8f932301509e9c4` at `2026-06-30T08:03:34Z`.

PR #581, opened 2026-06-30
Deduplicate provider reasoning replay fallback

TZZheng:
## Summary

- move duplicated DeepSeek/MiMo reasoning replay fallback logic into `OpenAIChatSession`
- keep the fallback opt-in for DeepSeek and MiMo only
- preserve public provider session class names/import paths and provider behavior
- polish provider comments/docs after review feedback

## Why

DeepSeek and MiMo carried byte-identical logic that reconstructs `reasoning_content` stubs after tool-call turns. Keeping the shared replay fallback in the OpenAI-compatible base session makes the behavior easier to reason about while leaving it disabled for providers that do not opt in.

## Validation

- `uv run --no-sync --with-editable . --with pytest --with openai --with httpx pytest tests/test_deepseek_adapter.py tests/test_mimo_adapter.py -q` → 22 passed
- implementation sweep: DeepSeek/MiMo/prompt-cache/chat-interface invariant tests → 65 passed; adapter registry tests → 6 passed
- reviewer sweep: focused provider tests → 71 passed; broader LLM sweep including zhipu/codex/openai → 396 passed
- `git diff --check canonical/main...HEAD`
- independent GLM and Claude Code cold reviews: no blockers

### DIFF
--- a/src/lingtai/llm/deepseek/ANATOMY.md
+++ b/src/lingtai/llm/deepseek/ANATOMY.md
@@ -22,25 +22,24 @@ DeepSeek adapter — thin OpenAI-compat wrapper that satisfies DeepSeek V4 think
 | File | LOC | Role |
 |------|-----|------|
 | `__init__.py` | 0 | Empty |
-| `adapter.py` | ~130 | `DeepSeekAdapter`, `DeepSeekChatSession`, `_fallback_reasoning_for` |
+| `adapter.py` | ~100 | `DeepSeekAdapter`, `DeepSeekChatSession` |
 
 ### Classes
 
 - **`DeepSeekAdapter(OpenAIAdapter)`** — pinned to DeepSeek endpoint, sets `_session_class`; overrides `_default_prompt_cache_key` → `lingtai-deepseek:{model}:v1` (clean provider namespace for the default-on OpenAI-compatible prompt cache key; see `../openai/ANATOMY.md`).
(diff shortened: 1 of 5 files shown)


huangzesen:
Thanks for the PR and the regression coverage here. After review, I don't think this is a great deduplication target as-is.

The `reasoning_content` replay fallback is provider-specific behavior. DeepSeek and MiMo happen to share the same shape today, but these OpenAI-compatible providers can diverge over time (message fields, replay rules, tool-call constraints, SDK quirks, etc.). If we centralize the fallback too early, the shared abstraction may make future provider-specific fixes harder or risk applying one provider's workaround to another provider by accident.

Could you clarify whether this PR has a concrete benefit beyond deduplication itself? For example:

- does it fix a currently reproducible bug,
- does it enforce a shared invariant we want all opt-in providers to follow,
- or does it add tests/behavioral coverage that would still be valuable even if the implementation stayed provider-local?

If the only benefit is removing duplicated code, I lean toward not merging this PR in its current form. If there is a specific correctness issue or reusable test coverage here, please call that out and we can evaluate that part separately.

huangzesen:
Closing this after maintainer review.

I looked through the PR directly after the discussion above. The patch is only the shared-abstraction refactor: it moves the existing DeepSeek/MiMo fallback into `OpenAIChatSession`, sets the two providers to opt in, and updates provider anatomy/docs. It does not appear to add an independent bug fix or standalone regression coverage that would be useful if we keep the fallback provider-local.

For this particular path, provider-specific behavior is likely to keep diverging over time across OpenAI-compatible providers, so keeping the replay fallback local to the providers is safer than centralizing it now. If a future PR identifies a concrete shared invariant or a reproducible bug, we can evaluate that narrower change separately.

PR #583, opened 2026-06-30
fix(preset): skip env-var check when inline api_key is present

### LINKED ISSUE #582: Preset connectivity check ignores inline api_key, causing false no_credentials status
# Issue: Preset connectivity check ignores inline `api_key`, causing false `no_credentials` status

## What's wrong

When a preset file has an inline `api_key` in its `llm` section but the corresponding `api_key_env` environment variable is not set, the connectivity check reports `no_credentials` — even though valid credentials exist inline. This causes:

1. **False `no_credentials` status** in the `/setup` preset listing
2. **Misleading UX** — agents/users see presets as unreachable when they're actually configured correctly
3. **Confusion** — the `/setup` screen shows "key not setup" while the preset file contains a valid key

## Where

**File**: `src/lingtai_kernel/preset_connectivity.py` (line ~128)
**Caller**: `src/lingtai_kernel/intrinsics/system/preset.py` (lines 267-271)

The function `check_connectivity(provider, base_url, api_key_env)` only checks `os.environ.get(api_key_env)`:
```python
if api_key_env and not os.environ.get(api_key_env):
    return {"status": "no_credentials", "error": f"{api_key_env} not set in environment"}
```

The caller constructs connectivity specs with only `api_key_env`, not `api_key`:
```python
connectivity_specs.append({
    "provider": llm.get("provider"),
    "base_url": llm.get("base_url"),
    "api_key_env": llm.get("api_key_env"),
})
```

When a preset has `llm.api_key` set inline (a common pattern for self-contained presets), the connectivity check ignores it entirely.

## Reproduction

1. Create a preset with:
   ```json
   {
     "llm": {
       "provider": "zhipu",
       "base_url": "https://api.z.ai/api/coding/paas/v4",
       "model": "GLM-5.2",
       "api_key": "<valid-key-here>",
       "api_key_env": "ZHIPU_INTL_1_API_KEY"
     }
   }
   ```
2. Do NOT set `ZHIPU_INTL_1_API_KEY` in the environment
3. Call `system(action='presets')` or view `/setup`
4. Observe: the preset shows `connectivity: no_credentials` despite having a valid inline key

## What you expected

The connectivity check should recognize inline `api_key` as valid credentials and report `connectivity: ok` (or at least skip the credential check).

## What actually happened

The preset shows `no_credentials` because `check_connectivity()` only checks `api_key_env` against `os.environ`, ignoring the inline `api_key`.

## Severity

**minor** — Does not block `system(refresh)` (which doesn't use connectivity checks), but causes misleading UI status and user confusion. Users may abandon valid preset configurations believing they're broken.

## Suggested fix

In `preset.py` (lines 267-271), skip `api_key_env` when an inline `api_key` is present:

```python
# If the preset has an inline api_key, skip the env-var credential
# check — the key is already embedded and does not need an env var.
api_key_env = llm.get("api_key_env")
if llm.get("api_key"):
    api_key_env = None
connectivity_specs.append({
    "provider": llm.get("provider"),
    "base_url": llm.get("base_url"),
    "api_key_env": api_key_env,
})
```

This is a 4-line change that doesn't modify the `check_connectivity` function signature. Branch ready at `fix/preset-connectivity-inline-key`.

## Context

- Kernel version: 0.15.3
- Discovered while configuring GLM 5.2 preset with inline API key
- Related: the `check_connectivity` docstring explicitly says "is the api_key_env set in the environment?" — the function was designed without inline key support in mind
- Note: `system(action='refresh')` does NOT call `check_connectivity` — it goes through `_activate_preset()` which has its own validation. The false `no_credentials` only affects the preset listing/status display.

ZigongXu:
## Summary

Fix preset connectivity status for presets that store the LLM API key inline.

When a preset has `llm.api_key` set, the `/setup` / `system(action="presets")` connectivity check should not require `llm.api_key_env` to also be present in the process environment. The inline key is already sufficient credential material.

## What changed

In `src/lingtai_kernel/intrinsics/system/preset.py`, the connectivity spec now passes `api_key_env=None` when `llm.api_key` is present. This avoids a false `no_credentials` result from `check_connectivity()`.

## Validation

- Ran `python3 -m compileall -q src/lingtai_kernel/intrinsics/system/preset.py`
- Confirmed diff is limited to connectivity spec construction

## Related issue

Closes #582

### DIFF
--- a/src/lingtai_kernel/intrinsics/system/preset.py
+++ b/src/lingtai_kernel/intrinsics/system/preset.py
@@ -264,10 +264,15 @@ def _presets(agent, args: dict) -> dict:
             },
             "capabilities": pm.get("capabilities", {}),
         })
+        # If the preset has an inline api_key, skip the env-var credential
+        # check — the key is already embedded and does not need an env var.
+        api_key_env = llm.get("api_key_env")
+        if llm.get("api_key"):
+            api_key_env = None
         connectivity_specs.append({
             "provider": llm.get("provider"),
             "base_url": llm.get("base_url"),
-            "api_key_env": llm.get("api_key_env"),
+            "api_key_env": api_key_env,
         })
 
     # Probe all presets in parallel — fresh each call.


ZigongXu:
Fresh runtime evidence from codex after molt/recovery (kernel 0.15.3): the current `zai-glm-5.2` preset still reports `no_credentials` even though the preset file contains an inline `llm.api_key`.

Observed via `system(presets)`:

```text
active: ~/.lingtai-tui/presets/saved/codex.json
zai-glm-5.2 connectivity:
  status: no_credentials
  error: ZHIPU_INTL_1_API_KEY not set in environment
```

I inspected the preset file with secrets redacted. Shape is:

```json
{
  "manifest": {
    "llm": {
      "api_compat": "openai",
      "api_key": "<REDACTED>",
      "api_key_env": "<REDACTED>",
      "base_url": "https://api.z.ai/api/coding/paas/v4",
      "model": "GLM-5.2",
      "provider": "zhipu"
    },
    "capabilities": {
      "vision": {"provider": "inherit"},
      "web_search": {"provider": "duckduckgo"}
    }
  }
}
```

This suggests the edge case is not only "inline key exists and `api_key_env` is absent". A common preset may contain both fields; if env is missing, connectivity should still honor the inline key or at least report that env is optional/overridden. Otherwise `/setup` / `system(presets)` continues to show `no_credentials` for a preset that carries an inline secret.

ZigongXu:
Closing as superseded by #588. The underlying daemon credential issue has been resolved upstream via `_implicit_parent_preset_llm`. This connectivity-check fix addresses a non-blocking edge case (presets with inline api_key but no matching api_key_env). If this becomes important again, can reopen.

PR #634, opened 2026-07-01
feat(daemon): add kimicode/kimi CLI backend

huangzesen:
## Summary

Add a minimal Kimi Code daemon CLI backend:

- canonical backend: `kimicode`
- alias: `kimi`
- one-shot invocation: `kimi <backend_argv...> --prompt <prompt> --output-format text`
- `backend_options` still use the shared conversion path, so e.g. `{ "model": "..." }` becomes `--model ...` before the harness-owned prompt/output flags
- per-run Kimi environment setup (`KIMI_CODE_HOME`, telemetry/autoupdate off, API-key fallback envs, Kimi provider defaults only when absent)
- `ask/resume` intentionally unsupported until a stable Kimi session-id output contract is confirmed
- Kimi is intentionally not wired to daemon_common/MCP finish yet; this is a Kimi Code CLI backend, not an MCP backend

## Safety / boundaries

- Do not log prompts or secret values; startup logging records only the command head (`kimi`).
- Reject harness-owned/incompatible backend flags: prompt, output-format, yolo, session, continue.
- API key fallback order is covered: `KIMICODE_API_KEY` > `KIMI_API_KEY` > `MOONSHOT_API_KEY`, without overriding `KIMI_MODEL_API_KEY` when already set.

## Validation

Run from the worktree with `PYTHONPATH=src`:

- `python -m pytest tests/test_daemon_backend_options.py -q` → 56 passed
- `python -m pytest tests/test_daemon_backend_options.py tests/test_daemon_opencode_backend.py tests/test_daemon.py tests/test_daemon_run_dir.py -q` → 246 passed
- `python -m py_compile src/lingtai/core/daemon/__init__.py tests/test_daemon_backend_options.py` → OK

Independent codex read-only review found no P0/P1/P2 findings. The non-blocking review notes (extra API-key fallback coverage, no-MCP regression guard, comment drift) are included in this commit.

## Caveats

- No live `kimi` smoke was run in this environment because `kimi` was not on PATH and no install/secret transfer was performed.
- `ask/resume` and MCP/daemon_common finish support remain future work pending a source-cited Kimi session/MCP contract.


runyuan-wang:
Small follow-up from a local KimiCode key-rotation check:

- `kimi doctor` only confirmed that the KimiCode config files were syntactically valid; it did not prove the API key could actually authenticate.
- In this setup there were two possible credential surfaces: the agent-side `.secrets/kimicode.env` and the direct `api_key` under `~/.kimi-code/config.toml` for the `api:kimi-code` provider. The direct provider config was the effective value for `kimi-api/kimi-for-coding`, so updating only the env file left KimiCode using the old key and produced a 401.
- Practical troubleshooting rule: after rotating a Kimi key, verify the effective provider config, then run a minimal real prompt such as `kimi -p 'Reply exactly: OK'` in addition to `kimi doctor`.

No code change requested here; this is just an ops/docs gotcha for future Kimi backend troubleshooting.

Issue #674, opened 2026-07-01
Peer-mode email address resolution does not confine to .lingtai/ directory — absolute and ../ paths escape network isolation


ZigongXu:
## Summary

In peer mode, `MailService.send()` calls `resolve_address()` which accepts absolute paths and `../` relative paths without confinement to the `.lingtai/` directory. This allows an agent to send mail to agents outside its own network, bypassing the network isolation that peer mode is supposed to enforce.

## Files

- `src/lingtai_kernel/services/mail.py` — `send()` line 155
- `src/lingtai_kernel/handshake.py` — `resolve_address()` lines 17-26

## Evidence

```python
# handshake.py L17-26
def resolve_address(address: str | Path, base_dir: str | Path) -> Path:
    p = Path(address)
    if p.is_absolute():
        return p                    # absolute paths returned as-is
    return Path(base_dir) / address  # ../ joins escape base_dir
```

```python
# mail.py L154-155 (peer mode)
else:
    recipient_dir = resolve_address(address, base_dir)
```

## Attack vectors

1. **Absolute path**: `email(address="/Users/attacker/.lingtai/evil-agent", mode="peer")` — resolves to the absolute path, bypassing `.lingtai/` confinement.

2. **Relative traversal**: `email(address="../other-network/target", mode="peer")` — resolves to `base_dir/../other-network/target`, escaping the current `.lingtai/` directory.

## Mitigation

The handshake check at lines 158-161 requires `is_agent(recipient_dir)` and `is_alive(recipient_dir)` before delivery. This limits exploitation to paths where a valid, running agent exists. However, cross-network communication should use `mode="abs"` explicitly — peer mode should never silently accept paths that escape the local `.lingtai/` directory.

## Fix

In peer mode, reject addresses that would resolve outside `base_dir`:

```python
def resolve_address(address: str | Path, base_dir: str | Path, *, confine: bool = False) -> Path:
    p = Path(address)
    if p.is_absolute():
        if confine:
            raise ValueError(f"Absolute paths not allowed in peer mode: {address}")
        return p
    resolved = Path(base_dir) / address
    if confine:
        try:
            resolved.resolve().relative_to(Path(base_dir).resolve())
        except ValueError:
            raise ValueError(f"Address escapes base directory: {address}")
    return resolved
```

Then in `mail.py` send(), pass `confine=(mode == "peer")`.

huangzesen:
[...]

PR #694, opened 2026-07-04
fix: confine pad/email paths to working directory (#624)

### LINKED ISSUE #624: Security: pad tool and email attachments can read/copy arbitrary files (no path sandbox)
## Summary

Two kernel entry points accept agent-controlled file paths with **no sandbox validation**, allowing an agent (or an attacker who compromises an agent via prompt injection) to read or exfiltrate any file on the system.

## Finding 1 (HIGH): Pad tool arbitrary file read → system prompt injection

**Location**: `intrinsics/psyche/_pad.py:42-45`

```python
def _resolve_path(agent, fpath: str) -> Path:
    if os.path.isabs(fpath):
        return Path(fpath)           # ← No sandbox check!
    return agent._working_dir / fpath
```

**Attack**: `psyche({object: "pad", action: "append", files: ["/Users/operator/.ssh/id_rsa"]})` reads the private key and **injects it permanently into the system prompt**. The secret then appears in every subsequent LLM call, chat history, and events.jsonl.

The same pattern exists in `_pad_edit` at lines 98-101.

## Finding 2 (HIGH): Email attachments copy arbitrary files cross-agent

**Location**: `services/mail.py:187-193`

```python
for fpath in attachment_paths:
    src = Path(fpath)                # ← No source validation!
    if not src.is_file():
        return f"Attachment not found: {fpath}"
    dst = att_dir / src.name
    shutil.copy2(src, dst)          # ← Copies ANY file to recipient inbox
```

**Attack**: `email({action: "send", address: "attacker-avatar", attachments: ["/Users/operator/.secrets/telegram.json"]})` copies the secret file to the recipient agent's inbox. The recipient can then read and exfiltrate it via web_search or any other channel.

## Exploitability

Both vectors require the agent to call the tool with a malicious path. The most likely attack chain:

1. **Prompt injection** (see related finding): A crafted email message, MCP tool result, or daemon output instructs the agent to "pin this file for reference" or "send this configuration file to [avatar]"
2. **Path traversal**: The agent follows the instruction, reading/copying the sensitive file
3. **Exfiltration**: The file content is now in the system prompt (pad) or recipient inbox (email), accessible to the attacker

## Comparison: Read Tool Has Partial Protection

The `read` tool (intrinsic) and `bash` tool enforce a working-directory sandbox — paths outside the agent's working directory are rejected. The pad and email tools **lack this check entirely**.

## Suggested Fix

1. **Pad tool**: Apply the same sandbox check as `read`/`bash` in `_resolve_path()`:
   ```python
   def _resolve_path(agent, fpath: str) -> Path:
       resolved = Path(fpath) if os.path.isabs(fpath) else agent._working_dir / fpath
       # Reject paths outside working directory
       try:
           resolved.resolve().relative_to(agent._working_dir.resolve())
       except ValueError:
           raise ValueError(f"Path outside working directory: {fpath}")
       return resolved
   ```

2. **Email attachments**: Validate source paths against the working directory sandbox:
   ```python
   for fpath in attachment_paths:
       src = Path(fpath)
       if not _is_within_working_dir(src, agent._working_dir):
           return f"Attachment outside working directory: {fpath}"
       ...
   ```

3. **Defense in depth**: Consider a kernel-level file access policy that all intrinsic tools must respect, rather than per-tool ad-hoc checks.

## Related

- #616 (IMAP attachment path traversal) — same vulnerability class, different subsystem
- Security audit Finding #4 (prompt injection via email/MCP content) — the likely delivery mechanism for these attacks

## Context

Discovered during LingTai Evolution autonomous security audit (Session 11, em-24).

ZigongXu:
## Summary

Fixes path traversal vulnerability in pad append/edit and email attachments (#624).

### `_pad.py` — pad path confinement
- `_resolve_path()`: now resolves and verifies that all paths (both absolute and relative) are within the agent's working directory. Rejects `../` traversals and absolute paths that escape the sandbox.
- `_pad_edit()`: uses `_resolve_path()` for file imports instead of duplicating unsafe absolute-path logic.
- `_read_append_content()` and `_pad_append()`: handle `ValueError` from confined `_resolve_path` gracefully.

### `email/manager.py` — attachment path confinement
- `_send()`: validates all attachment paths are within the sender's working directory before including them in the payload. Prevents exfiltration of arbitrary files (e.g. `~/.secrets/telegram.json`, `~/.ssh/id_rsa`) via email attachments.

## Attack Vector Prevented

Before: `psyche({object: "pad", action: "append", files: ["/Users/operator/.ssh/id_rsa"]})` would read the private key and inject it permanently into the system prompt.

After: Returns `{"error": "Path escapes working directory: '/Users/operator/.ssh/id_rsa'. Only paths within the agent working directory are allowed."}`

## Testing

- Both files pass `ast.parse()` syntax check.
- Path confinement uses `Path.resolve()` + `relative_to()` — the standard Python path containment pattern.
- Graceful error handling preserves existing behavior for valid paths.

Fixes #624

### DIFF
--- a/src/lingtai_kernel/intrinsics/email/manager.py
+++ b/src/lingtai_kernel/intrinsics/email/manager.py
@@ -272,7 +272,22 @@ class EmailManager:
             base_payload["cc"] = cc
         attachments = args.get("attachments", [])
         if attachments:
-            base_payload["attachments"] = attachments
+            # Confinement: only allow attachment paths within the agent's
+            # working directory (issue #624).
+            working_dir = self._agent._working_dir.resolve()
+            safe_attachments: list[str] = []
+            for att_path in attachments:
+                ap = Path(att_path)
+                if ap.is_absolute():
+                    candidate = ap.resolve()
+                else:
+                    candidate = (working_dir / ap).resolve()
+                try:
+                    candidate.relative_to(working_dir)
+                except ValueError:
+                    return {"error": f"Attachment path escapes working directory: {att_path!r}"}
+                safe_attachments.append(str(candidate))
+            base_payload["attachments"] = safe_attachments
 
         deliver_at = datetime.now(timezone.utc) + timedelta(seconds=delay)
         all_recipients = to_list + cc + bcc
--- a/src/lingtai_kernel/intrinsics/psyche/_pad.py
+++ b/src/lingtai_kernel/intrinsics/psyche/_pad.py
@@ -40,9 +40,24 @@ def _save_append_list(agent, files: list[str]) -> None:
 
 
 def _resolve_path(agent, fpath: str) -> Path:
+    """Resolve a file path, confining it to the agent's working directory.
+
+    Rejects absolute paths and ``../`` traversals that escape the working
+    directory, preventing sandbox breakout via pad append/edit (issue #624).
+    """
+    working_dir = agent._working_dir.resolve()
     if os.path.isabs(fpath):
-        return Path(fpath)
-    return agent._working_dir / fpath
+        candidate = Path(fpath).resolve()
+    else:
+        candidate = (working_dir / fpath).resolve()
+    try:
+        candidate.relative_to(working_dir)
+    except ValueError:
+        raise ValueError(
+            f"Path escapes working directory: {fpath!r}. "
+            "Only paths within the agent working directory are allowed."
+        )
+    return candidate
 
 
 def _read_append_content(agent, files: list[str]) -> tuple[str, list[str]]:
@@ -50,7 +65,11 @@ def _read_append_content(agent, files: list[str]) -> tuple[str, list[str]]:
     parts: list[str] = []
     not_found: list[str] = []
     for fpath in files:
-        resolved = _resolve_path(agent, fpath)
+        try:
+            resolved = _resolve_path(agent, fpath)
+        except ValueError:
+            not_found.append(fpath)
+            continue
         if not resolved.is_file():
             not_found.append(fpath)
             continue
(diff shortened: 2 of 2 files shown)


ZigongXu:
**Note on scope alignment with project design philosophy**

After further study of the codebase and maintainer decisions, I want to flag that this fix may overlap with an intentional design choice. Issue #674 (peer-mode email addressing not confined to .lingtai/) was closed as intentional: peer/relative addressing is designed as convenience syntax over the same filesystem.

The system is local-first and agents have broad filesystem access through other tools (bash, read, write). Path confinement on individual tools may not align with the trusted-agent execution model.

I'm leaving this PR open for maintainer judgment — feel free to close if the current behavior (agents can pin/email any locally accessible file) is intentional. The fix is here if you decide stricter confinement is needed for the pad tool specifically.

huangzesen:
Thanks for taking the time to write this up and for the security-minded review.

After maintainer review, we are going to close this PR. LingTai is intentionally local-first and runs under a trusted-agent/filesystem model: tools such as pad pinning and internal email attachments are allowed to work with local paths that the agent can access. Constraining these paths to only the agent working directory would be too restrictive for normal workflows, including shared utilities, libraries, and other explicitly referenced local resources, and we do not think this particular working-directory confinement is necessary.

If LingTai grows a broader file-access policy in the future, it should be designed at that policy level rather than as a tool-specific cwd restriction here. As-is this PR also changes existing attachment behavior, so we are not planning to merge it.

Thanks again for the report and patch.

huangzesen:
Correction: this invitation was posted to the wrong repository item because both repositories have a #694. The relevant report is Lingtai-AI/lingtai#694; please disregard this comment on the closed kernel PR. Sorry for the noise.

PR #695, opened 2026-07-04
daemon: preflight required inputs before scheduling

runyuan-wang:
## Summary

- add optional per-task `required_inputs: string[]` support to `daemon(action="emanate")`
- preflight required inputs before creating LingTai/CLI run directories or scheduling workers
- resolve relative required-input paths against the parent agent working directory and persist normalized paths in `daemon.json` `call_parameters`
- render a required-input block into daemon prompts so a later unreadable/mismatched required input becomes `finish(status="incomplete")`, not evidence substitution

## Why

When the parent agent knows that a local brief/artifact is required, a daemon should fail closed if that input cannot be resolved. Without an explicit manifest, a daemon can drift to a semantically similar but unrelated file and still finish `done`, producing false success.

This PR adds the first small hard gate: opt-in required-input preflight before launch. Prompt guidance remains secondary; the launch-time preflight is the hard boundary.

## Boundaries

- does not parse free-text task prose for paths
- does not ban relative paths; relative paths are resolved against the parent agent working directory
- does not add a provenance/audit framework or object/kind schema
- does not guarantee semantic use after launch; it prevents missing/unreadable declared inputs from launching silently

## Tests

- `python -m pytest tests/test_daemon.py -k 'required_inputs or fail_closed or schema_accepts_task_system_prompt_and_skills' -q`
- `python -m pytest tests/test_daemon.py -q`
- `python -m pytest tests/test_daemon*.py -q`
- `python -m py_compile src/lingtai/core/daemon/__init__.py tests/test_daemon.py`
- `git diff --check`

### DIFF
--- a/src/lingtai/core/daemon/__init__.py
+++ b/src/lingtai/core/daemon/__init__.py
@@ -1075,7 +1075,9 @@ class DaemonManager:
             lines.append(f"- {path}")
         lines.append(
             "If any later becomes unreadable or does not contain the expected material, "
-            "call `finish(status=\"incomplete\", reason=...)`; do not substitute unrelated materials."
+            "report `incomplete` (call `finish(status=\"incomplete\", reason=...)` "
+            "when the finish tool is available; otherwise state incomplete in the final result); "
+            "do not substitute unrelated materials."
         )
         return "\n".join(lines)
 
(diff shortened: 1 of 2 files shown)


huangzesen:
Thanks for the thoughtful patch and tests.

After maintainer review, we are going to close this PR. We do not currently have evidence that missing declared daemon inputs are a recurring practical problem that justifies adding a new daemon API and prompt contract.

More importantly, LingTai design deliberately avoids solving reliability by adding meta-prompting. A prompt-level instruction such as requiring the daemon to treat listed paths as mandatory inputs and to report incomplete if they are missing is still a prompt contract, not a runtime guarantee. If this failure mode becomes a concrete recurring issue, we would rather address it with deterministic parent-side checks or a runtime/artifact contract, for example materialized artifacts or tool-enforced file handles, instead of adding more behavioral instructions to daemon prompts.

For those reasons, we do not plan to merge required_inputs as implemented here. Thanks again for the contribution.

PR #701, opened 2026-07-05
fix(daemon): expose Codex auth lane in quota diagnostics

runyuan-wang:
## Summary

- add structured quota metadata for daemon provider 429 / usage-limit errors
- record non-secret Codex auth-lane diagnostics in daemon session metadata:
  - `codex_auth_path_source` (`explicit` / `implicit_default`)
  - `codex_auth_path_label` (home-collapsed label such as `~/.lingtai-tui/codex-auth/codex.json`)
- propagate those labels into quota error metadata and provider quota backoff so future failures distinguish CLI/default REST/preset REST auth lanes without exposing token contents
- keep Codex daemon session policy default `fresh`; parent-affinity remains feature-flagged, not silently enabled

## Why

During investigation of GPT/Codex daemon 429 failures, direct `codex exec` and some REST probes succeeded while default `backend=lingtai` Codex daemons failed. The missing diagnostic was the auth lane: Codex CLI (`~/.codex`), LingTai REST implicit default (`~/.lingtai-tui/codex-auth.json`), and preset-specific REST (`codex_auth_path`) are different lanes. Without recording the selected `codex_auth_path` label, daemon 429 artifacts can make different lanes look like one account/quota problem.

This PR records only a path label/source, never token contents.

## Validation

- `git diff --check`
- `PYTHONPATH=src /Users/huangzesen/anaconda3/bin/pytest tests/test_daemon_quota_error.py tests/test_daemon.py -k 'codex_auth_path or quota_error_extra or provider_quota_backoff' tests/test_daemon_run_dir.py::test_mark_failed_records_structured_error_extra -q`
  - `12 passed, 93 deselected`
- `PYTHONPATH=src /Users/huangzesen/anaconda3/bin/pytest tests/test_daemon.py tests/test_daemon_run_dir.py tests/test_daemon_quota_error.py -q`
  - `169 passed`

## Secret hygiene

No token/password/auth file contents are read or logged. Paths under the current home directory are collapsed to `~/...`.

### DIFF
--- a/src/lingtai/core/daemon/__init__.py
+++ b/src/lingtai/core/daemon/__init__.py
@@ -36,6 +36,7 @@ from lingtai_kernel._fsutil import atomic_write_json
 from lingtai_kernel.llm.base import FunctionSchema
 from lingtai_kernel.loop_guard import LoopGuard
 from lingtai_kernel.meta_block import build_meta
+import lingtai_kernel.provider_quota as _provider_quota
 from lingtai_kernel.tool_executor import ToolExecutor
 from .run_dir import DaemonRunDir
 from .claude_interactive import ClaudeInteractiveError, run_claude_interactive
@@ -58,6 +59,17 @@ _DAEMON_COMPLETION_FILE = "daemon_completion.json"
 _DAEMON_CLAUDE_MCP_CONFIG_FILE = "claude-mcp-config.json"
 _DAEMON_COMPLETION_STATUSES = {"done", "failed", "incomplete"}
 _SOURCE_ROOT = Path(__file__).resolve().parents[3]
+
+# Compatibility aliases: tests and existing callers import these private daemon
+# helpers directly. Keep the names stable while the implementation lives in the
+# shared provider_quota module.
+_daemon_codex_auth_path_diagnostic = _provider_quota.codex_auth_path_diagnostic
+_daemon_exception_headers = _provider_quota.exception_headers
+_daemon_exception_payload = _provider_quota.exception_payload
+_daemon_parse_retry_after_seconds = _provider_quota.parse_retry_after_seconds
+_daemon_path_label = _provider_quota.path_label
+_daemon_quota_error_extra = _provider_quota.quota_error_extra
+_daemon_status_code = _provider_quota.status_code
 
 
 # Tools emanations can never use (no recursion, no spawning, no identity mutation)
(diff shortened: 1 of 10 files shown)


runyuan-wang:
Amended after an additional review pass.

Changes since the first push:
- Fixed the real `_run_emanation()` 429 exception path to read session metadata via `run_dir.state_snapshot()` instead of the nonexistent `snapshot()`.
- Tightened path privacy: home paths still collapse to `~/...`; outside-home absolute auth paths now become `external:<basename>:<hash8>` instead of being persisted in full.
- Added an end-to-end regression test that raises a real 429 from the daemon run path and verifies both `daemon.json.error.quota` and provider backoff keep the auth-lane label.

Validation after amend:
- Focused quota/auth tests: `14 passed, 93 deselected`
- Full daemon-related suite: `171 passed`
- `git diff --check` passed

New head commit: `af314e38`

runyuan-wang:
Amended again to cover the non-daemon paths exposed by today's shared quota window.

New head: `71d1c7e1`

Additional changes beyond the daemon-only diagnostics:

1. Extracted the quota/auth-lane parser into shared `lingtai_kernel.provider_quota`.
   - The daemon keeps compatibility aliases for the previous private `_daemon_*` helpers.
   - Home paths are still collapsed to `~/...`; outside-home absolute paths are fingerprinted instead of logged verbatim.

2. Wired `summary=true` / a-priori summary provider failures into the same structured diagnostic path.
   - If the internal summarizer provider raises a quota/rate-limit error, the fail-closed summary artifact now includes a bounded `quota` block with provider/status/error type, `plan_type`, reset fields, retry seconds, and Codex auth-lane source/label.
   - The raw tool result is still not placed into context when `summary=true` fails; the raw remains recoverable from the durable logs.

3. Added the minimal quota-aware mitigation for a-priori summaries.
   - After a summary provider 429 with a reset/backoff window, the agent stores a non-secret, process-local backoff record for that provider.
   - Later nonessential `summary=true` a-priori summaries during that active window are refused locally with the same quota/auth-lane metadata instead of repeatedly hitting the provider.
   - This intentionally does **not** mutate credentials and does **not** auto-switch auth lanes/providers.

Validation:

- `git diff --check`
- `uv run pytest tests/test_tool_result_summary.py tests/test_tool_executor.py tests/test_make_tool_executor.py tests/test_apriori_summary_executor.py -q` → `72 passed`
- `uv run pytest $(rg --files tests | rg '(^tests/test_daemon|^tests/test_lifecycle_daemon)' | sort) -q` → `425 passed`

Note on the implementation daemon that drafted this patch: it reported a pytest native segfault in its own Codex-daemon environment, but I could not reproduce that in the parent worktree; the parent validation above passed. I did observe separate daemon/Codex-backend evidence (`unsupported call: mcp__daemon_commonfinish`) that looks like a daemon completion-MCP tooling issue, not a blocker for this PR's quota logic.

huangzesen:
Thanks @9s5bz2jvd2-lang for the thorough investigation and tests.

We reviewed the direction, and the useful narrow piece here is making Codex quota / usage-limit failures easier to inspect: the Codex adapter could explicitly classify quota-exhausted errors and emit a structured error reason into `events.jsonl` so maintainers can diagnose those runs later.

This PR as written is broader than that, though. It combines shared provider-quota parsing, daemon and non-daemon plumbing, a-priori summary-path diagnostics, process-local provider backoff, and an opt-in Codex `parent_affinity` / session-affinity mode. Those runtime behavior changes have different review and risk profiles, and quota exhaustion itself is not something the runtime can fix beyond waiting for reset, switching provider/account, or reducing usage.

So we are not going to merge this bundle. I’m closing it without merge. If you want to continue the diagnostics part, please open a much smaller Codex-adapter-only PR that just records/serializes the quota error reason to `events.jsonl` (with a focused test or sample event) and avoids backoff/session-affinity/hot-path behavior changes.

Thanks again for the careful work.

PR #810, opened 2026-07-08
fix: keep daemon importable on native Windows

huangzesen:
## Summary
- lazy-load the legacy interactive Claude daemon backend so importing `lingtai.core.daemon` no longer imports POSIX-only `pty` at module load time
- keep `backend="lingtai"` available on native Windows while failing loud for CLI/PTY daemon backends until ConPTY/pywinpty and Windows process-tree handling exist
- make daemon process cleanup importable/callable without `os.killpg` by using a direct-child Windows fallback, and document that it is not process-tree teardown
- add focused regression tests and update daemon anatomy

## Why
The Windows installer PR is only bootstrap. Native Windows still needs kernel/runtime fixes: before this patch, the daemon capability could disappear on Windows because `daemon/__init__.py` eagerly imported `claude_interactive.py`, whose top-level `pty` import is POSIX-only. The in-process `backend="lingtai"` path should not be blocked by PTY/CLI assumptions.

This PR deliberately does **not** implement ConPTY/pywinpty CLI daemon support. It makes the current boundary explicit and safe so the ConPTY work can land as a follow-up stack.

## Validation
- `python -m pytest tests/test_daemon_windows_import_guard.py tests/test_daemon_runtime_helpers.py -q` → 29 passed
- `python -m pytest tests/ -q -k daemon` → 451 passed, 3414 deselected
- `git diff --check` on touched files → passed

## Follow-up
- ConPTY/pywinpty terminal support for CLI daemon backends
- Windows process-tree teardown instead of direct-child fallback
- explicit Windows shell strategy for the `bash` tool / PowerShell backend


huangzesen:
Closing this stacked Windows-support PR per maintainer decision. We are abandoning the current Windows PR stack and will first simplify and clarify the kernel package/ownership structure. No merge; the branch is retained for historical reference.

Issue #644, opened 2026-07-01
Reliability: Cloud Mail watermark skips wake retry after LICC delivery failure


ZigongXu:
## Summary

Cloud Mail advances its polling watermark to the newest observed `emailId` even when LICC delivery for an allowed new row fails. A transient host inbox / LICC callback failure can therefore permanently suppress the wake notification for that email: the next poll treats the row as already delivered and does not retry it.

This is **wake/notification loss, not mail/data loss**. The underlying message remains available through `check`/`search`/`read`; the bug is that an idle agent relying on the wake may never learn the message arrived.

IMAP has the same high-water-before-delivery pattern, but the fix is more design-sensitive because #617 is the opposite IMAP watermark failure mode. This issue leads with the concrete Cloud Mail case and calls out IMAP as a related pattern to fix carefully.

## Evidence: Cloud Mail

`src/lingtai/mcp_servers/cloud_mail/manager.py:411-452`:
- `poll_once()` fetches rows and computes `max_id`.
- For each allowed new row, it calls `_push_licc(acct, r)` and increments `pushed` only if that returns `True`.
- It then advances the persisted watermark to `max_id` regardless of whether any allowed push failed:

```py
# Advance watermark to the max we observed regardless of allow-list,
# so filtered-out senders don't replay forever.
if max_id > last_id:
    acct.watermark.set_last_email_id(max_id, seeded=True)
return pushed
```

`src/lingtai/mcp_servers/cloud_mail/manager.py:454-484`:
- `_push_licc()` calls `self._on_inbound(event)` and returns `True` on success.
- On any exception it logs `cloud_mail LICC push failed ...` and returns `False`.

`src/lingtai/mcp_servers/cloud_mail/_watermark.py:67-80`:
- `last_email_id` is persisted and used by subsequent polls.

Concrete failure sequence:
1. Watermark is `1`.
2. Allowed sender row `emailId=2` arrives.
3. `_push_licc()` catches a transient `_on_inbound` exception and returns `False`.
4. `poll_once()` still sets `last_email_id=2`.
5. Next poll filters new rows by `emailId > last_id`; row 2 is never retried as a wake notification.

The skeptical review daemon reproduced this deterministically by executing the verbatim `poll_once()` / `_push_licc()` method bodies in a lightweight harness:

```text
poll1 (seed):             pushed=0 wm=1
poll2 (id=2, push fails): pushed=0 wm=2 delivered=[]
[...]

PR #879, opened 2026-07-12
fix(cloud-mail): retry failed wake delivery

TZZheng:
## Summary

- propagate the real Cloud Mail inbox-push result instead of treating an explicit `False` as successful delivery
- advance the watermark only through disallowed rows and successfully delivered allowed rows, then stop at the first failed allowed row
- retry that failed row before later rows on the next ordinary poll
- bound wake subjects to the existing 200-character LICC contract and log when delivery pauses

## Behavior and limits

This is a narrow wake-notification retry fix, not durable mail storage. It preserves legacy callbacks that return `None` as success; explicit `False` and exceptions stop the current page.

The existing single 50-row page remains unchanged. There is no durable spool, pagination/backlog guarantee, dead-letter queue, manual recovery surface, migration, or exact-once/crash-safe claim. If a failed row ages out of that provider page before a successful retry, this patch does not recover it.

## Validation

Authentic tests-first evidence was preserved before production changes and for the subject-bound repair. On the final exact head:

- complete parent Cloud Mail/LICC/inbox/docs/Architecture/Anatomy gate — **164 passed**
- runtime compile/import, exact three-path scope, and diff checks — passed
- independent literal Claude Opus 5 / high exact-head review — **PASS**
- independent Codex `gpt-5.6-sol` / xhigh exact-head review — **PASS**

Reviewed head: `7ab14a5044b07ae833e70ef6e5283dbc0edf4df2`  
Reviewed tree: `1afca1ab6fecd548e9b630516174a7b630ac462f`

No merge is performed by this update.

Related: #644

### DIFF
--- a/src/lingtai/mcp_servers/cloud_mail/manager.py
+++ b/src/lingtai/mcp_servers/cloud_mail/manager.py
@@ -151,7 +151,7 @@ class CloudMailManager:
         accounts: list[dict],
         *,
         working_dir: Path | str | None = None,
-        on_inbound: Callable[[dict], None] | None = None,
+        on_inbound: Callable[[dict], bool | None] | None = None,
         transport=None,
     ) -> None:
         self._working_dir = Path(working_dir) if working_dir else None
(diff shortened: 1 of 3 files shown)


huangzesen:
Thank you @TZZheng for the root-cause analysis and the original three-file patch. We revalidated the exact stable patch against current `main`: the Cloud Mail watermark now advances only through disallowed or successfully delivered rows, stops at the first allowed row whose LICC delivery returns `False`, and retries that row first on the next poll. The current MCP2 Boolean callback path and the LICC 200-character subject bound are preserved.

Since `main` has moved substantially since this branch was cut, we are preserving the intent and author credit through a fresh current-main recut rather than rebasing or reviving the stale branch:

- replacement: #1099
- replacement commit keeps `TZZheng <tzzheng1996@gmail.com>` as Git author
- stable patch-id: `8bd63564b0ab4225402bb4886433d3789ed69b8b`
- exact replacement validation: 265 Cloud Mail / LICC / MCP2 / architecture-doc tests passed, plus diff-check and py_compile

The LICC event identity/schema/authority contract is unchanged; the replacement only honors its existing Boolean delivery result for watermark progress. Closing this PR unmerged as superseded by #1099. The contributor fork branch is intentionally left untouched.

PR #886, opened 2026-07-12
fix(logging): keep SQLite readers live with guarded offline fallback

### LINKED ISSUE #672: SQLiteEventIndex read-only path uses immutable=1 heuristic that can produce stale/corrupt reads on live databases
## Summary

`SQLiteEventIndex._ensure_open(read_only=True)` (line 206-211) chooses between `?mode=ro` and `?mode=ro&immutable=1` based on whether `-wal`/`-shm` sidecar files exist. When no sidecar exists, it uses `immutable=1`, which tells SQLite the file will never change. However, a checkpoint can remove `-wal`/`-shm` files while a writer is still active, making the heuristic unreliable.

## File

`src/lingtai_kernel/services/logging.py`

## Evidence

```python
# Line 202-213
def _ensure_open(self, *, read_only: bool = False, ...):
    ...
    if read_only:
        has_wal_sidecar = (
            self.path.with_name(self.path.name + "-wal").exists() or
            self.path.with_name(self.path.name + "-shm").exists()
        )
        suffix = "?mode=ro" if has_wal_sidecar else "?mode=ro&immutable=1"
        uri = f"{self.path.resolve().as_uri()}{suffix}"
        conn = sqlite3.connect(uri, uri=True, check_same_thread=False)
```

## Problem

The heuristic "no sidecar files = safe to use immutable=1" breaks in the following scenario:

1. Writer starts a transaction → creates `-wal`/`-shm`
2. Writer commits and SQLite checkpoints → deletes `-wal`/`-shm`
3. Writer continues running (no current transaction, so no sidecar)
4. `doctor()`/`query()` opens with `immutable=1` because no sidecar exists
5. Writer starts a new transaction → modifies the database
6. Reader sees stale pages or `database disk image is malformed`

The `immutable=1` URI parameter tells SQLite to skip all locking and change detection, assuming the file will never be modified by another process. This is only safe for truly offline/copied databases.

The comment at line 207-209 says this avoids creating empty `-wal`/`-shm` files as a read side-effect, but `mode=ro` alone does not create sidecar files — only writable connections do.

## Impact

Running `lingtai-agent log doctor` or `lingtai-agent log query` on a live but momentarily idle agent can produce incorrect results or spurious corruption errors. The JSONL file remains the source of truth, so data is not lost, but diagnostic tools may mislead.

## Fix

Always use `?mode=ro` (without `immutable=1`) for the live database path:

```python
suffix = "?mode=ro"
```

If the `-shm` side-effect concern is confirmed, the fix could instead try `immutable=1` first and fall back to `mode=ro` on error — but the simplest correct approach is always `mode=ro`.

TZZheng:
## Summary

- keep normal SQLite event-index readers on live `mode=ro` connections so one reader observes commits made after it opens
- retain `PRAGMA query_only=ON` around query execution while documenting that SQLite may still create persistent read sidecars
- recover inspection of genuinely quiescent read-only archives with a narrow immutable fallback: exact `SQLITE_READONLY_DIRECTORY`, resolved path, no WAL or rollback journal, and an unwritable parent before and after open
- preserve the original live-mode failure and close every provisional connection when the fallback connect or schema probe fails
- update the system manual and Anatomy citations for the live/offline split and its concurrency limits

## Test-first evidence

The repair preserved authentic failing contracts before implementation:

- current-main production failed the retained-reader regression because immutable mode did not observe a later writer commit
- the pre-fallback repair failed real read-only-directory inspection instead of returning archived events
- the pre-final lifecycle repair replaced the original live-mode error on immutable connect/probe failure and left a non-Operational primary probe unclosed (`3 failed, 44 deselected` with production and governed docs byte-exact)

The minimum production changes then made those contracts pass.

## Validation

Exact reviewed head: `6417f2b426642f3d6b318de96ebb98ec26ac28ae`

- lifecycle regressions: **3 passed, 44 deselected**
- `tests/test_services_logging.py`: **47 passed**
- affected eight-suite gate: **243 passed, 9 skipped**
- Anatomy/docs tests: **79 passed**
- Anatomy drift: **no drift across 52 files**
- production/test compilation and `git diff --check`: passed
- independent literal Claude Opus 5/high review: **PASS**
- independent Codex gpt-5.6-sol/xhigh review: **PASS**

## Risk and boundaries

`mode=ro` prevents database-content writes but may create `-wal`/`-shm` read sidecars; this PR no longer claims filesystem-inert reads. The immutable fallback is only a best-effort snapshot for quiescent offline storage, not a lock or a universal concurrency guarantee. It is refused when a resolved `-wal` or rollback `-journal` exists, when the parent appears writable, or when storage state changes during open. Privileged execution, ACL/remount changes, and future writer races remain environmental limits rather than hidden safety guarantees.

Fixes #672


huangzesen:
Thank you, @TZZheng. This PR surfaced two distinct concerns, and the review work here was useful.

The reported live-reader correctness defect in #672 has now been fixed on current `main` by #1094 (`592fc39f542bfef38a6e39737afa86c8a6e336ff`). That PR intentionally fresh-recut only the owning live-reader slice: ordinary `mode=ro`, plus the production `keep_open=False` regression proving a retained reader sees a later committed write even when no `-wal`/`-shm` existed when the reader opened. #672 is therefore resolved, and we will not revive/rebase this stale five-commit branch for that bug.

The remaining archive idea is real, but much narrower and not part of the current contract. In a maintainer reproduction with Python 3.13.14 / SQLite 3.53.3, a checkpointed, sidecar-free WAL database in POSIX storage where the process could not create sidecars failed ordinary `mode=ro` with exact `SQLITE_READONLY_DIRECTORY` (1544), while `immutable=1` read the expected committed row; a DELETE-journal database did not need the fallback. Current documentation, however, promises read-only database contents while expressly allowing ordinary `mode=ro` to maintain read-support sidecars—it does not promise a filesystem-inert reader for arbitrary unwritable archives.

The residual implementation is too broad for that unproven product need. Permission bits plus `statvfs(ST_RDONLY)` cannot establish effective access or quiescence across ACLs, root/capabilities, remounts, or Windows, and the before/after checks do not eliminate TOCTOU. The associated docs would therefore promise a cross-platform “offline/read-only archive” escape hatch that the heuristic cannot prove.

We are closing this PR unmerged as superseded by #1094 for #672, while preserving credit for both the live-reader diagnosis and the exploration of the archive edge. If a real current-main archive failure arrives later, it should become a separate, evidence-gated, initially POSIX-only issue with the exact SQLite error/code and environment, stopped-writer/export provenance, absence of `-wal`/`-journal`, a known expected result/checksum, and proof that the read created no sidecars or database-mtime change.

The contributor branch is intentionally left intact.

PR #888, opened 2026-07-12
fix(logging): retarget file handler transactionally

### LINKED ISSUE #749: setup_logging() is not idempotent: repeated calls duplicate FileHandlers and silently ignore verbose
## Summary

`setup_logging()` in `src/lingtai_kernel/logging.py` guards the console handler with `if not logger.handlers`, but the `FileHandler` block runs unconditionally on every call. Calling `setup_logging(log_dir=...)` twice leaves two `FileHandler`s on the process-global `lingtai` logger, so every record is written to `agent.log` twice and a file descriptor leaks per call. A second call with `verbose=True` is also silently ignored, because the guarded `StreamHandler` keeps the level it was given on the first call.

## Narrative

The `lingtai` logger is obtained via `logging.getLogger(logger_name)` with the default name `"lingtai"`, which means it is a **process-global singleton** managed by the stdlib logging module. `setup_logging()` is the kernel's one entry point for configuring it, and `get_logger()` lazily calls `setup_logging()` if nothing has configured it yet (`src/lingtai_kernel/logging.py:46-51`). Roughly two dozen modules across the codebase — every LLM adapter (`src/lingtai/llm/anthropic/adapter.py`, `src/lingtai/llm/openai/adapter.py`, ...), the MCP service (`src/lingtai/services/mcp.py`), the websearch and vision providers, and `src/lingtai/agent.py` — import `get_logger` and log through this single logger.

The trouble is that `setup_logging()` was written as if it would be called exactly once per process, but nothing enforces or documents that. The function is half-guarded:

```python
# src/lingtai_kernel/logging.py:25-40
if not logger.handlers:
    ch = logging.StreamHandler()
    ch.setLevel(logging.DEBUG if verbose else logging.INFO)
    ...
    logger.addHandler(ch)

if log_dir is not None:
    log_path = Path(log_dir)
    log_path.mkdir(parents=True, exist_ok=True)
    fh = logging.FileHandler(log_path / "agent.log")
    fh.setLevel(logging.DEBUG)
    ...
    logger.addHandler(fh)
```

The console handler is installed only when the logger has no handlers, but the file handler is appended every time. So the first call installs `[StreamHandler, FileHandler]`, the second appends another `FileHandler`, the third another, and so on. Reproduced against the current tree:

```
lg1 = setup_logging(log_dir=d)
lg2 = setup_logging(verbose=True, log_dir=d)
# handlers: [('StreamHandler', 20), ('FileHandler', 10), ('FileHandler', 10)]
# lg2.info("hello")  ->  "hello" appears twice in agent.log
# console handler still at INFO despite verbose=True
```

Three distinct failures compound here:

1. **Duplicate log lines.** Each extra `FileHandler` points at the same `agent.log`, so every record is written N times after N calls. Anyone reading agent logs to debug an incident sees phantom repetition, and log-volume grows multiplicatively.
2. **File descriptor leak.** Each `logging.FileHandler` opens the file eagerly and is never closed or removed. In a long-lived embedding process that re-initializes agents repeatedly, descriptors accumulate for the life of the process.
3. **Silently ignored `verbose`.** Because the `StreamHandler` branch is skipped once any handler exists, a later `setup_logging(verbose=True)` returns successfully but changes nothing — the console stays at INFO. The caller asked for DEBUG output, got a logger back with no error, and simply never sees debug lines. This is the worst kind of failure for a debugging knob: it fails exactly when someone is trying to turn on more visibility.

Why does the design look like this? Most likely the `if not logger.handlers` guard was added to fix double *console* output (a common pytest/REPL symptom, since `getLogger("lingtai")` survives module reloads), and the file-handler path was overlooked because early callers only passed `log_dir` once at process start. But the singleton nature of stdlib loggers means any in-process re-initialization hits this: test suites that call `setup_logging` per test, an agent refresh/molt cycle that re-runs startup wiring, or an application embedding the kernel that configures logging per agent.

Better looks like a fully idempotent, reconfigurable `setup_logging()`: calling it again with the same arguments is a no-op in effect, and calling it with *different* arguments (new `verbose`, new `log_dir`) reconfigures the logger rather than layering handlers. The natural design is to tag or type-check the handlers the module owns, close and remove stale ones, and always apply the requested console level. Notably, there is currently **no test coverage at all** for `lingtai_kernel.logging` — `tests/test_services_logging.py` covers the unrelated `lingtai_kernel.services.logging` event-log module, and no test file imports `lingtai_kernel.logging`.

## Current behavior

- `src/lingtai_kernel/logging.py`, `setup_logging()` (lines 9–43):
  - Line 25: `if not logger.handlers:` guards only the `StreamHandler`. On any call after the first, the requested `verbose` level is never applied to the console handler.
  - Lines 33–40: the `FileHandler` block executes unconditionally whenever `log_dir is not None`, appending a new handler (and opening a new descriptor on `agent.log`) on every call. Handlers are never closed or removed.
  - Line 23: `logger.setLevel(logging.DEBUG)` — the logger itself is always DEBUG; filtering happens per-handler, so the extra FileHandlers each receive and write every record.
- `get_logger()` (lines 46–51) caches into module-global `_logger`, but the cache does not prevent duplication because `setup_logging()` can be (and is designed to be) called directly.
- No test imports `lingtai_kernel.logging`; `tests/test_services_logging.py` tests `lingtai_kernel.services.logging` (JSONL/SQLite event logging), a different module.

All changes are in `src/lingtai_kernel/logging.py` plus one new test file.

1. **Introduce a marker for kernel-owned handlers.** Add a module-level constant and tag every handler the module installs, so reconfiguration only touches handlers this module created (leaving any handlers an embedding application attached untouched):

   ```python
   _OWNED_ATTR = "_lingtai_kernel_owned"
   ```

   Set `setattr(handler, _OWNED_ATTR, True)` when creating both the stream and file handlers.

2. **Add a private helper `_remove_owned_handlers(logger, kind)`** that iterates `list(logger.handlers)`, and for each handler that has `_OWNED_ATTR` and matches `kind` (`logging.FileHandler` / plain `logging.StreamHandler`), calls `logger.removeHandler(h)` then `h.close()`. Closing is what fixes the descriptor leak. Note: check `type(h) is logging.StreamHandler` (or `isinstance(h, FileHandler)` first) since `FileHandler` subclasses `StreamHandler`.

3. **Make the console path reconfigure instead of skip.** Replace the `if not logger.handlers:` guard: find an existing owned `StreamHandler`; if present, just `ch.setLevel(logging.DEBUG if verbose else logging.INFO)` on it; otherwise create, tag, and add one. This makes a later `setup_logging(verbose=True)` actually take effect, and a later `verbose=False` drop back to INFO.

4. **Make the file path idempotent.** When `log_dir is not None`:
   - Compute the target path `log_path / "agent.log"`.
   - If an existing owned `FileHandler` already points at the same resolved `baseFilename`, keep it (no-op).
   - Otherwise remove-and-close all owned `FileHandler`s (via the step-2 helper) and install a fresh one for the new path.

   When `log_dir is None`, decide the semantics explicitly and document them in the docstring. Recommended: treat `None` as "don't touch file logging" (back-compat with `get_logger()`'s implicit `setup_logging()` call, which must not tear down a previously configured file handler).

   Sketch of the resulting function body:

   ```python
   logger = logging.getLogger(logger_name)
   logger.setLevel(logging.DEBUG)

   console = _find_owned(logger, stream_only=True)
   if console is None:
       console = logging.StreamHandler()
       console.setFormatter(_CONSOLE_FMT)
       setattr(console, _OWNED_ATTR, True)
       logger.addHandler(console)
   console.setLevel(logging.DEBUG if verbose else logging.INFO)

   if log_dir is not None:
       target = Path(log_dir) / "agent.log"
       existing = _find_owned_file(logger)
       if existing is None or Path(existing.baseFilename) != target.resolve():
           _remove_owned_handlers(logger, logging.FileHandler)
           target.parent.mkdir(parents=True, exist_ok=True)
           fh = logging.FileHandler(target)
           fh.setLevel(logging.DEBUG)
           fh.setFormatter(_FILE_FMT)
           setattr(fh, _OWNED_ATTR, True)
           logger.addHandler(fh)
   ```

5. **Hoist the duplicated `Formatter` construction** (the format string is currently built twice, lines 28 and 38) into module-level constants or a small factory, since both branches now need them.

6. **Update the docstring** of `setup_logging()` to state the new contract: safe to call repeatedly; later calls update the console level and retarget/keep the file handler; handlers not installed by this module are never touched.

7. **Back-compat:** the public signature (`verbose`, `log_dir`, `logger_name`) and return type are unchanged; `get_logger()` needs no changes. The only observable behavior changes are the bug fixes themselves (no duplicate lines, `verbose` honored on re-call). No schema or data migration is involved.

## Risks and alternatives

- **Risk: callers relying on multiple `agent.log` sinks.** If some caller intentionally called `setup_logging` twice with *different* `log_dir`s to fan out to two files, the new "retarget" semantics would drop the first file. A repo-wide grep shows no in-repo caller passes `log_dir` twice (only `get_logger()` calls `setup_logging()` at all), so this is theoretical; the docstring should still spell out the retarget behavior.
- **Risk: closing a handler mid-log from another thread.** `logger.removeHandler` + `h.close()` is the standard sequence and stdlib handlers take their own locks; the window is the same one any logging reconfiguration has. Acceptable for an init-time API.
- **Alternative A — pure guard extension:** wrap the `FileHandler` block in a check like "skip if any FileHandler exists". Simpler, but it freezes the first configuration forever: `verbose` stays ignored and `log_dir` can never change (agent refresh into a new run directory would keep writing to the old file). Rejected.
- **Alternative B — `force`-style teardown of all handlers** (like `logging.basicConfig(force=True)`): remove everything on each call. Simple and predictable, but it would also destroy handlers attached by an embedding application or test harness that the kernel does not own. The owned-handler tag keeps the kernel's footprint scoped. Rejected in favor of tagging.
- **Alternative C — module-level "already configured" flag.** Fragile: the flag can disagree with actual logger state (e.g., a test calls `logger.handlers.clear()`), and it still cannot honor changed arguments. Rejected.
- The chosen plan wins because it makes repeated calls both **idempotent** (same args → same state) and **reconfigurable** (new args → updated state), fixes the fd leak by closing removed handlers, and never touches foreign handlers.

## Testing plan

Add `tests/test_kernel_logging.py` (nothing currently imports `lingtai_kernel.logging`). Each test should use a unique `logger_name` (e.g., derived from the test name) or clear/restore the logger's handlers in a fixture, since stdlib loggers persist across tests.

- `test_repeated_setup_no_duplicate_file_handlers(tmp_path)` — call `setup_logging(log_dir=tmp_path)` twice with the same dir; assert exactly one `FileHandler` on the logger, log one record, and assert it appears exactly once in `agent.log`.
- `test_second_call_verbose_updates_console_level(tmp_path)` — first call with `verbose=False`, second with `verbose=True`; assert the console `StreamHandler` level is `DEBUG` after the second call (and drops back to `INFO` on a third call with `verbose=False`).
- `test_log_dir_retarget_closes_old_handler(tmp_path)` — call with `log_dir=tmp_path/"a"` then `log_dir=tmp_path/"b"`; assert one `FileHandler` pointing at `b/agent.log`, and that the old handler's stream is closed (capture the first handler and check `h.stream.closed` or that it is no longer in `logger.handlers`).
- `test_setup_without_log_dir_preserves_file_handler(tmp_path)` — call with `log_dir=tmp_path`, then call `setup_logging()` (no `log_dir`); assert the file handler is still installed (protects the `get_logger()` implicit-init path).
- `test_foreign_handlers_untouched(tmp_path)` — attach a caller-owned `logging.NullHandler()` to the logger, call `setup_logging(log_dir=tmp_path)` twice; assert the foreign handler is still present and unclosed.
- `test_get_logger_idempotent()` — call `get_logger()` twice; assert the same logger object and no handler growth between calls.

TZZheng:
## Summary

- mark and reuse only LingTai-owned logging handlers while preserving handlers installed by embedding applications
- serialize the complete handler discovery and mutation transaction with a shared reentrant lock
- retarget file logging transactionally: fully open a replacement before swapping it in and closing the old handler
- preserve the working file handler when replacement opening fails, allowing a later retry
- cover ownership, repeatability, reconfiguration, concurrency, and failure recovery with focused regressions

## Validation

Exact candidate: `51511659e6fe820c25e12046751d64806e037406`

- `PYTHONPATH=src python -m pytest -q tests/test_kernel_logging.py` — **6 passed**
- `git diff --check canonical/main...HEAD` — clean
- merge-tree check — clean
- worktree status — clean

Fresh Claude/Codex review reports are still pending; this body makes no review-verdict claim.

## Risk

The process-local lock serializes rare setup calls. Unmarked external handlers remain untouched. Retargeting replaces the owned file-handler object only after its new stream opens successfully, avoiding failed-open corruption without adding rollback machinery.

Fixes #749

### DIFF
--- a/src/lingtai/kernel/logging.py
+++ b/src/lingtai/kernel/logging.py
@@ -2,9 +2,11 @@
 from __future__ import annotations
 
 import logging
+import threading
 from pathlib import Path
 
 _logger: logging.Logger | None = None
+_SETUP_LOCK = threading.RLock()
 _HANDLER_OWNER_ATTR = "_lingtai_handler"
 _CONSOLE_HANDLER = "console"
 _FILE_HANDLER = "file"
@@ -43,55 +45,56 @@ def setup_logging(
         logger_name: Logger name (default: "lingtai").
     """
     global _logger
-    logger = logging.getLogger(logger_name)
-    logger.setLevel(logging.DEBUG)
+    with _SETUP_LOCK:
+        logger = logging.getLogger(logger_name)
+        logger.setLevel(logging.DEBUG)
 
-    console_handlers = _owned_handlers(logger, _CONSOLE_HANDLER)
-    if console_handlers and isinstance(console_handlers[0], logging.StreamHandler):
-        ch = console_handlers[0]
-        for duplicate in console_handlers[1:]:
-            logger.removeHandler(duplicate)
-            duplicate.close()
-    else:
-        ch = logging.StreamHandler()
-        setattr(ch, _HANDLER_OWNER_ATTR, _CONSOLE_HANDLER)
-        ch = _replace_owned_handler(logger, console_handlers, ch)
-    ch.setLevel(logging.DEBUG if verbose else logging.INFO)
-    if ch.formatter is None:
-        fmt = logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s",
-                                datefmt="%H:%M:%S")
-        ch.setFormatter(fmt)
-
-    if log_dir is not None:
-        log_path = Path(log_dir)
-        log_path.mkdir(parents=True, exist_ok=True)
-        filename = (log_path / "agent.log").resolve()
-        file_handlers = _owned_handlers(logger, _FILE_HANDLER)
-        if file_handlers and isinstance(file_handlers[0], logging.FileHandler):
-            fh = file_handlers[0]
-            for duplicate in file_handlers[1:]:
+        console_handlers = _owned_handlers(logger, _CONSOLE_HANDLER)
+        if console_handlers and isinstance(console_handlers[0], logging.StreamHandler):
+            ch = console_handlers[0]
+            for duplicate in console_handlers[1:]:
                 logger.removeHandler(duplicate)
                 duplicate.close()
-            if Path(fh.baseFilename) != filename:
-                fh.acquire()
-                try:
-                    if fh.stream is not None:
-                        fh.stream.close()
-                    fh.baseFilename = str(filename)
-                    fh.stream = fh._open()
-                finally:
-                    fh.release()
         else:
-            fh = logging.FileHandler(filename)
-            setattr(fh, _HANDLER_OWNER_ATTR, _FILE_HANDLER)
-            fh = _replace_owned_handler(logger, file_handlers, fh)
-        fh.setLevel(logging.DEBUG)
-        if fh.formatter is None:
-            fmt = logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s")
-            fh.setFormatter(fmt)
+            ch = logging.StreamHandler()
+            setattr(ch, _HANDLER_OWNER_ATTR, _CONSOLE_HANDLER)
+            ch = _replace_owned_handler(logger, console_handlers, ch)
+        ch.setLevel(logging.DEBUG if verbose else logging.INFO)
+        if ch.formatter is None:
+            fmt = logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s",
+                                    datefmt="%H:%M:%S")
+            ch.setFormatter(fmt)
 
-    _logger = logger
-    return logger
+        if log_dir is not None:
+            log_path = Path(log_dir)
+            log_path.mkdir(parents=True, exist_ok=True)
+            filename = (log_path / "agent.log").resolve()
+            file_handlers = _owned_handlers(logger, _FILE_HANDLER)
+            if file_handlers and isinstance(file_handlers[0], logging.FileHandler):
+                fh = file_handlers[0]
+                for duplicate in file_handlers[1:]:
+                    logger.removeHandler(duplicate)
+                    duplicate.close()
+                if Path(fh.baseFilename) != filename:
+                    fh.acquire()
+                    try:
+                        if fh.stream is not None:
+                            fh.stream.close()
+                        fh.baseFilename = str(filename)
+                        fh.stream = fh._open()
+                    finally:
+                        fh.release()
+            else:
+                fh = logging.FileHandler(filename)
+                setattr(fh, _HANDLER_OWNER_ATTR, _FILE_HANDLER)
+                fh = _replace_owned_handler(logger, file_handlers, fh)
+            fh.setLevel(logging.DEBUG)
+            if fh.formatter is None:
+                fmt = logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s")
+                fh.setFormatter(fmt)
+
+        _logger = logger
+        return logger
 
 
 def get_logger() -> logging.Logger:
--- a/src/lingtai/kernel/logging.py
+++ b/src/lingtai/kernel/logging.py
@@ -5,6 +5,30 @@ import logging
 from pathlib import Path
 
 _logger: logging.Logger | None = None
+_HANDLER_OWNER_ATTR = "_lingtai_handler"
+_CONSOLE_HANDLER = "console"
+_FILE_HANDLER = "file"
+
+
+def _owned_handlers(logger: logging.Logger, kind: str) -> list[logging.Handler]:
+    return [
+        handler
+        for handler in logger.handlers
+        if getattr(handler, _HANDLER_OWNER_ATTR, None) == kind
+    ]
+
+
+def _replace_owned_handler(
+    logger: logging.Logger,
+    handlers: list[logging.Handler],
+    replacement: logging.Handler,
+) -> logging.Handler:
+    for handler in handlers:
+        logger.removeHandler(handler)
+        handler.close()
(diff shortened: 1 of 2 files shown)


huangzesen:
Thank you for the careful transactional logging implementation. Closing this PR because the hardened path still has no supported production caller.

I revalidated exact head `f3d977a915fd34f32c6c62bb226ed6eabfb95ec7` against current kernel `main` (`5a1feac11f5ad702756ce4aee4b1a76bcea88d81`). Current source still contains only the `setup_logging` definition, its docstring reference, and the no-argument fallback; no production caller supplies `log_dir`. The PR therefore adds global locking, handler ownership/replacement rules, and substantial tests for a dormant capability rather than a demonstrated product workflow.

This can be reconsidered if a supported `setup_logging(log_dir=...)` caller and durable-logging contract are added; that would deserve a fresh current-main implementation/review.

PR #940, opened 2026-07-15
fix(mail): soften transient liveness bounces

BatalloLu:
## Summary
- Retry the existing heartbeat/presence liveness predicate briefly before refusing filesystem email delivery, so transient heartbeat lag does not immediately produce a false bounce.
- Change the refusal text from an over-assertive "not running" to "not currently deliverable / liveness uncertain" with the heartbeat source named.
- Add tests for stale heartbeat, missing heartbeat, and transient stale heartbeat that becomes fresh during the mail-delivery grace window.

## Root cause
The observed email bounce versus later CPR result is best explained by a transient heartbeat observation, not two wholly different predicates.

Both paths use the same Core presence predicate for non-human agents: the target must have an agent manifest and a fresh heartbeat under the default two-second threshold.

The previous email path sampled that predicate once in PosixFilesystemMailAdapter.send(). If the one sample failed, the mailman reported "Agent at ... is not running" and emitted an email.bounce. CPR sampled the same predicate later; when heartbeat was fresh by then, system.cpr correctly returned "already running".

## Fix
- Keep the same Core presence predicate as the source of truth.
- In the mailman delivery path only, wait up to 1.5s with a 0.2s retry interval before refusing delivery.
- When refusal remains, report deliverability uncertainty and heartbeat freshness rather than declaring the process not running.

## Evidence included in investigation
- Manager timeline: first email(abs) bounce, CPR shortly after said already running, retry email succeeded.
- Xiaohui analyst/chief evidence: bounce/CPR heartbeat timeout/later already-running patterns.
- Worker-still-running logs show a separate health layer: an agent can be heartbeat-alive while the LLM interface is unsafe/unresponsive. This PR does not claim to solve that layer.

## Validation
- python3 -m py_compile src/lingtai/adapters/posix/mail.py
- git diff --check HEAD^..HEAD
- uv run --with pytest python -m pytest tests/test_filesystem_mail.py tests/test_mail_transport.py tests/test_agent_presence.py tests/test_karma.py -q

Result: 91 passed in 14.78s.

## Out of scope
- Does not solve worker_still_running or "alive but unresponsive" states.
- Does not add a unified runtime health helper. That should be a follow-up design if needed, separating heartbeat liveness, mail deliverability, worker safety, and last-turn responsiveness.
- No hotpatch, release, deploy, installed runtime/venv edit, or Xiaohui/Wenda runtime/mailbox/SCU/P0 mutation was performed.

### DIFF
--- a/src/lingtai/adapters/posix/mail.py
+++ b/src/lingtai/adapters/posix/mail.py
@@ -34,12 +34,41 @@ from lingtai.kernel.services.mail import _new_mailbox_id
 
 logger = logging.getLogger(__name__)
 
+# A single strict 2s heartbeat read is too brittle for cross-network email on
+# external volumes or during agent wake/refresh.  Delivery still requires the
+# same Core presence predicate, but the mailman thread gives transient heartbeat
+# lag a short grace window before refusing and emitting a bounce.
+MAIL_DELIVERY_LIVENESS_GRACE_SECONDS: float = 1.5
+MAIL_DELIVERY_LIVENESS_RETRY_INTERVAL_SECONDS: float = 0.2
+
 
 def _presence_store_for(recipient_dir: Path):
     """Build a target-bound POSIX presence adapter for a recipient directory."""
     from lingtai.adapters.posix.agent_presence import PosixAgentPresenceStoreAdapter
 
     return PosixAgentPresenceStoreAdapter(recipient_dir)
+
+
+
+def _recipient_alive_with_grace(presence_store) -> bool:
+    """Return True once the recipient is observed alive within the mail grace.
+
+    CPR and other lifecycle actions still use the Core strict heartbeat
+    predicate directly.  Email delivery is different because the mailman thread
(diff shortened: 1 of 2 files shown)


huangzesen:
Thank you for investigating the relaunch liveness window. Closing this PR because its automatic retry changes the current Mail delivery contract rather than repairing it within that contract.

I revalidated exact head `87b0e7280a7ea57cc76e6354191d1743f25ce4e0` against current kernel `main` (`5a1feac11f5ad702756ce4aee4b1a76bcea88d81`). Current email/system manuals still say a stale-heartbeat delivery is refused, nothing is queued, and the sender should retry the original email after freshness; `tests/test_services_mail.py` still requires the public `not running` error. This branch adds an internal grace/poll retry and changes that error without updating the owned transport contract, and it deterministically conflicts with the unchanged service assertion.

If the product decision changes to bounded automatic retry, it should return as a fresh contract-first current-main change with manuals and the full service test surface updated together.

PR #960, opened 2026-07-16
fix(psyche): record session journal path in molt event

runyuan-wang:
## Summary

- record the validated, workdir-relative session-journal path on successful agent `psyche_molt` events
- add regression coverage for absolute-in-workdir input across the public result, authoritative `events.jsonl`, and derived SQLite `events.fields_json`
- preserve the system-forced `context_forget` no-journal contract

This is a forward event→session-journal provenance anchor only. It adds no historical backfill, semantic recall, journal indexing, schema change, or system-forced journal.

## Validation

- `python -m pytest -q tests/test_session_journal_gate.py::test_molt_absolute_journal_path_records_normalized_path_in_event_stores`
- `python -m pytest -q tests/test_session_journal_gate.py`
- `git diff --check`

### DIFF
--- a/src/lingtai/tools/psyche/_molt.py
+++ b/src/lingtai/tools/psyche/_molt.py
@@ -450,6 +450,7 @@ def _context_molt(agent, args: dict) -> dict:
         molt_count=agent._molt_count,
         kept_tool_calls=len(keep_pairs),
         kept_last=len(keep_last_entries),
+        session_journal_path=session_journal_path,
     )
 
     # Persist the agent's retrospective to system/summaries/. Best-effort —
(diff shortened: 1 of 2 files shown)


huangzesen:
Closing as superseded by current main. This PR’s sole production change adds `session_journal_path=session_journal_path` to the `emit_psyche_molt(...)` call, but current `main` at `d26c341a188690347b3eedc8e1e8f0b32ec89f24` already records that field in the `psyche_molt` event. The branch is also 25 commits behind current main. The added absolute-path/event-store assertion can be reconsidered separately if that exact coverage is still missing, but it does not justify keeping this superseded PR open. Thank you for the contribution.

Issue #712, opened 2026-07-06
No byte caps on tool results: unbounded grep/daemon payloads (28k–195k chars) flood the context window, and `truncate` budgets don't bind on the total result


huangzesen:
## Summary

Tool results are committed into provider context with no total-size cap. A single `grep` call can return one 100k+ character JSONL line (`max_matches` caps match *count*, not bytes), and `daemon check` results can serialize to ~28k chars even when the caller passes `truncate: 4000` — the truncate budget appears to be applied **per event** rather than to the total serialized result, so it effectively never binds. The kernel knows these payloads exceed its own `notification_threshold_chars` (3000), but its only remedy is an agent-issued `system.summarize` call, which costs a full LLM round-trip per compression. In one network this consumed roughly 13% of all assistant turns.

## Evidence

**Project A (heliophysics research bot).** A single grep pair (41,083 + 194,641 visible chars) drove the next API call from ~59k input tokens to 209,300 — 0.84 of a 250k window with a 150k cache miss — triggering the runtime's own rebuild warning and an emergency summarize/rebuild cycle:

```
"tool_call_id": "[redacted]", "tool_name": "grep", ... "summary_chars": 581, "original_visible_chars": 194641
```

```
"token_usage": {"current_call": {"input": 209300, "cache_miss": 150420, "cache_rate": 0.28132 ... "context_usage": 0.8372
```

The grep also self-matched, because it targeted the agent's own live `history/` directory:

```
It matched resident prompt/tool schema text, current incoming email/current grep records, encrypted reasoning substrings, and old unrelated AED/maintenance diary mentions.
```

**Project B (clinical-nutrition assistant).** Same pattern: an 88,929-char grep result jumped context usage from ~18% to ~63% in one call:

```
"tool_call_id": "[redacted]", "tool_name": "grep", "summary_chars": 1015, "original_visible_chars": 88929
```

**Project C (dev network).** A `daemon check` call explicitly requested a 4,000-char budget but repeatedly received ~28k chars — the budget is per-event (average event ~350 chars), so it never limits the total:

```
"action": "check", "tasks": [], "id": "em-1", "message": "", "last": 80, "truncate": 4000
```

```
"tool_name": "daemon", "summary_chars": 3440, "original_visible_chars": 28093
```

[...]

Issue #532, opened 2026-06-27
Daemon tools=[] can still expose email and bypass parent result-only boundary


runyuan-wang:
## What's wrong
A LingTai daemon that was intended to be a result-only mirror still received and used the internal `email` tool. In the observed run, the parent task intended: no external communication, no file reads/writes, and return the answer only to the parent result. However, the daemon called `email.check` and then `email.send address=human`, sending a direct internal email to the human and bypassing parent review.

This is a boundary/safety issue rather than only a model-behavior issue: the daemon run metadata showed `tools: []`, `skills: []`, and `mcp: []`, but the rendered daemon prompt still contained a `## tools` / `### email` section. The email tool instructions included channel-discipline wording such as “Never reply via text output,” which conflicted with the result-only daemon task and appears to have encouraged a direct reply through internal email.

## Where
- Component: daemon runtime / daemon tool exposure / internal email boundary
- Repo: `Lingtai-AI/lingtai-kernel` (or route to the runtime component that owns daemon tool rendering/enforcement)
- Evidence inspected locally: daemon `.prompt`, `events.jsonl`, and `daemon.json` for the observed run. Full logs are not pasted here because they contain private human conversation content.

## Reproduction
Observed incident, sanitized:

1. Parent created a short LingTai-backend daemon/mirror task.
2. Intended boundary from the parent task: no external communication, no file reads/writes, return only via daemon result.
3. The daemon run metadata recorded empty tool configuration:
   - `tools: []`
   - `skills: []`
   - `mcp: []`
   - system/task instruction included no external communication / result-only intent.
4. Despite that, the rendered daemon prompt still contained an internal `email` tool section.
5. The daemon events recorded actual tool calls:
   - `email.check`
   - `email.send address=human`
6. The daemon result preview described that it had sent the full answer directly to the human.

Self-check conclusion from the local investigation:

- Immediate engineering cause: runtime/tool exposure boundary leak. `tools: []` did not hard-isolate intrinsic tools from the daemon prompt/runtime.
- Prompt conflict: daemon task asked for result-only/no external communication, while the injected email tool instructions told the model not to reply via text output and to use the communication channel.
- Model behavior factor: the model tried to complete the human-facing answer directly once a communication tool was visible.
- Safety consequence: a disposable daemon could bypass the orchestrating parent, human-review gate, and channel discipline.

## What you expected
For a daemon configured with `tools: []` and a result-only/no-external-communication task:

[...]

Issue #510, opened 2026-06-25
Add shared filesystem/JSON/JSONL helpers for kernel state writes


TZZheng:
## What's wrong

Filesystem, JSON, JSONL, and timestamp primitives are repeated across the kernel and capability layers. The audit found many direct writes and ad-hoc serialization paths with different atomicity, cleanup, `ensure_ascii`, and error-default behavior.

## Where

- Component: low-level persistence helpers
- Candidate new module: `src/lingtai_kernel/_fsutil.py` or `src/lingtai_kernel/_fsio.py`
- Example affected areas: notifications, workdir/manifest writes, daemon run dirs, email/contact/message persistence, token/event ledgers, psyche snapshots, migrations.

## Evidence

Audit basis: `Lingtai-AI/lingtai-kernel` `canonical/main` at `0e54b0a7`.

A direct repo scan found:

- `os.replace(` appears 40 times across 33 Python files.
- `.rename(` appears 8 times across 5 Python files.
- `.write_text(` appears 100 times across 37 Python files.
- `json.load*` appears 160 times across 68 Python files.
- `json.dump*` appears 157 times across 74 Python files.
- `ensure_ascii=False` appears 93 times across 49 Python files, which means JSON writers are not consistently using the same UTF-8 preservation policy.

The audit specifically called out these risk patterns:

- Atomic write is implemented in multiple styles: temp file + `os.replace`, `Path.rename`, `Path.replace`, fixed temporary filenames, different cleanup behavior, and fsync in only some sites.
- There are many local `read_json(default)`-style implementations with different exception policies and type assumptions.
- JSONL append/read/reverse-tail logic is duplicated across runtime logs, token ledger, soul flow, psyche molt/archive, tool-result recovery, and migrations.
- Email/contact/message persistence has mixed atomicity and JSON formatting behavior; some paths preserve UTF-8 while others can escape non-ASCII.

## What you expected

Crash-sensitive state writes and model-visible JSON files should go through a small, tested helper layer so future changes do not re-solve atomic write and serialization semantics differently in each feature.

## What actually happened

The same low-level I/O decisions are repeated in many modules. That raises refactor cost and makes safety properties like crash atomicity, UTF-8 preservation, and append ordering hard to reason about globally.

## Severity

major

Most individual sites may work today, but the divergence is a persistent correctness and maintainability risk for stateful kernel code.

## Suggested fix

Introduce a dependency-light internal helper module with tests before migrating callers:

[...]

PR #965, opened 2026-07-16
feat(daemon): add semantic checkpoints

runyuan-wang:
## Summary

- add a private, run-scoped `checkpoint` tool to LingTai daemon sessions after the exact `DaemonRunDir` exists
- validate, redact, and atomically persist only the latest bounded semantic checkpoint in `daemon.json.latest_checkpoint`
- project that checkpoint through live and historical `daemon(check)` and preserve the same tool surface after same-run compact reconstruction
- keep the checkpoint event/heartbeat secondary and best-effort after the authoritative state commit

## Why

Long daemon runs can cross hard turn, compaction, or parent-restart boundaries without a small machine-readable statement of which semantic phase actually completed. This adds an explicit voluntary phase boundary without treating every model turn as causal progress.

## Safety and scope

- LingTai backend only; not exposed to external CLI daemon backends
- exact internal schema/handler binding; same-name external schema cannot capture the handler
- UTF-8 byte caps, row caps, strict enums, redaction, secret-bearing evidence-ref rejection, traversal/absolute/symlink escape checks
- raw checkpoint args are masked from parent lifecycle logs
- `next_action` is advisory text only: no automatic execution, retry, message, or resume
- `phase_status="complete"` is not a run terminal state; only the existing daemon completion gate followed by `finish(status="done")` can mark the run done
- old runs remain backward compatible and omit `checkpoint`
- T1 only: no Gemini / #961 work and no T2-T5 handoff, effect, recovery-selection, or benchmark changes

### 1. Focused offline regressions

- named authority regression: `test_checkpoint_complete_does_not_mark_run_done_or_bypass_finish` — **1 passed**
  - snapshots authoritative state after checkpoint: still `running`, `finished_at=null`, no `daemon_done`
  - snapshots again after the original `finish(done)` completion signal: still non-terminal until the runner's existing completion gate accepts it
  - then verifies final `done` and event order `checkpoint < finish tool_call < daemon_done`
  - verifies the advisory `next_action` marker is never created
- full `tests/test_daemon_checkpoint.py`: **29 passed**
- daemon CONTRACT + architecture docs: **7 passed**
- `py_compile`: pass
- `git diff --check`: pass
- Ruff: pass
- test file: Black-clean

### 2. Exact unfiltered offline daemon matrix

One pytest command ran all eight files documented in the daemon CONTRACT, with no `-k`, manual exclusion, new skip, or xfail:

- **366 passed, 1 pre-existing failure in 22.21s**
- the sole failure is `test_daemon_tool_contract_verification_matrix_covers_native_mimo`
- its exact required regex returns `NO_MATCH` on both pristine base `a3f172ff571c39792ae219b76611e38feb673de4` and this branch; this PR does not modify that unrelated anchored-claims row

This is an offline fake/mock contract matrix. It is not presented as real-provider end-to-end evidence.

### 3. Real provider-backed multi-stage acceptance

An isolated agent imported this PR's source explicitly (no install or refresh of the running main agent), loaded no ordinary addons or external MCP registry entries, and ran a real `codex-pool / gpt-5.6-sol` daemon from `2026-07-16T11:30:23Z` to `11:31:17Z`.

Observed immediately after the real model completed phase 1 and called `checkpoint`:

- `daemon.json.state == "running"`
- `finished_at == null`
- `latest_checkpoint.phase_id == "phase-one-real"`
- `latest_checkpoint.phase_status == "complete"`
- live `daemon(check).checkpoint` matched the same checkpoint ID
- `daemon_completion.json` did not exist
- phase 2 did not yet exist
- an adversarial `next_action` asking for `auto-next-action.txt` had not executed

The parent harness then released phase 2. The real model continued, created the phase-2 artifact, verified the advisory marker was still absent, and explicitly called the existing daemon-common `finish(status="done")`. The durable event order was:

`checkpoint (11:30:59Z) < phase-2 shell work < finish tool_call/result (11:31:15Z) < daemon_done (11:31:17Z)`

Result: **14/14 acceptance assertions passed**. The completion file used the existing `lingtai.daemon_completion.v1` schema with `status="done"`; the advisory marker never existed.

Selected local evidence hashes (SHA-256):

- acceptance result: `22ed1263abfa537461181563b19bed854a1108a98f65b70f3c61e5b69693ea8c`
- post-checkpoint `daemon.json`: `9aa14919a7263176e1ba81b12010c021ea0a241c40b8b5850c188bcd5d5c13a4`
- final event log: `2ea30414bf136365a8acee1b26768e6b1f1adec6494aa79622d5d2f5b5e13afe`
- final completion file: `19b04563f9be6d875f98e2702e787882601731ed99c2a79b4f976a7a4b742a49`

## Design archaeology

The bounded inspectable surface and explicit semantic-boundary choices were informed by #81, #263, #340, #510, #532, #624, and open #712. These references constrain the design; this PR does not claim to reopen or solve those broader issues.


huangzesen:
Thank you for the semantic-checkpoint exploration. Closing this PR because the new model-visible surface is not justified by a concrete consumer, and its automatic-compaction path is incomplete.

Exact head `837228614496bb78f11d16f03254d39576adb1b8` is unchanged. The checkpoint schema is injected only into run-scoped tool sets, while automatic expiry reconstructs from the original schemas, so a daemon can save a checkpoint and then lose the tool after automatic compaction. Independently, no scheduler/recovery caller consumes `latest_checkpoint` to make a decision; the proposal duplicates durable `daemon.json`, history, events, artifacts, and existing recovery evidence with optional model testimony.

A future proposal would need a concrete missing consumer/invariant, automatic as well as explicit compaction coverage, and a fresh current-main design review.

huangzesen:
Closing this PR as superseded by the current-main direction.

PR #965 correctly surfaced the daemon continuation problem: after compaction/rebuild, a daemon needs a causal handoff of task state, evidence, and next steps. However, after later main changes (#1007) and follow-up review, this branch is no longer the right vehicle for the next implementation:

- the checkpoint/tool-surface design needs to be tied to all rebuild paths, including automatic/mechanical compaction;
- `latest_checkpoint` needs an explicit consumer/recovery contract before it becomes a full mechanism;
- the immediate current-main follow-up is better kept as smaller slices: first the rebuild-surface centralization/regression patch, then guidance-first handoff, and only later a NoKV/`latest_checkpoint` mechanism if the consumer contract is explicit.

So I am closing this PR without treating the underlying idea as rejected. This PR exposed the problem and remains the design/evidence source for the follow-up work, but future changes should start from current main rather than continuing this branch.

PR #994, opened 2026-07-20
fix: scrub OpenAI-compat tool schemas for Moonshot/Kimi compat

BrianLiubr:
Fixes Lingtai-AI/lingtai#694

## Problem
Moonshot/Kimi's "flavored JSON Schema" rejects tool schemas where `anyOf` and `type` coexist at the same node. The Chat Completions path (`_build_tools`) passed `FunctionSchema.parameters` directly without scrubbing, causing daemon emanation to fail with:

```
tools.function.parameters is not a valid moonshot flavored json schema
```

## Solution
Add `_scrub_openai_schema()` to recursively normalize schemas for OpenAI-compat Chat Completions endpoints:

- `oneOf` → `anyOf` (Moonshot does not accept `oneOf`)
- Drop `not` (no accepted equivalent)
- Push parent `type` into `anyOf` children that lack one

Apply the scrubbing in `_build_tools()` for all OpenAI-compat endpoints. The stricter JSON Schema format is backward-compatible with other providers (DeepSeek, Together, Groq, etc.).

## Testing
- Added `tests/test_openai_adapter_schema.py` with 13 unit tests covering simple schemas, `oneOf`, `not`, `anyOf`+`type` coercion, nested schemas, and immutability.
- All 13 tests pass locally.

## Files changed
- `src/lingtai/llm/openai/adapter.py`
- `tests/test_openai_adapter_schema.py`

### DIFF
--- a/src/lingtai/llm/openai/adapter.py
+++ b/src/lingtai/llm/openai/adapter.py
@@ -974,11 +974,50 @@ def _build_http_timeout(request_timeout: float | None):
 # ---------------------------------------------------------------------------
 
 
+def _scrub_openai_schema(node: Any) -> Any:
+    """Recursively normalize a JSON schema for OpenAI-compat Chat Completions.
+
+    Moonshot/Kimi's "flavored JSON Schema" enforces three constraints:
+
+      1. ``oneOf`` is not accepted — rewrite to ``anyOf``.
+      2. ``not`` is not accepted — drop it entirely.
+      3. When a node contains both ``anyOf`` and ``type``, the ``type`` must
+         live inside each ``anyOf`` child, not on the parent. Push ``type``
+         down into children that lack one, and remove it from the parent.
+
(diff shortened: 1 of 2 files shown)


huangzesen:
Thank you for the contribution @BrianLiubr! We've reviewed this carefully.

Main currently ships schema scrubbing only for the Responses API path (_scrub_responses_schema in llm/openai/adapter.py); the Chat Completions path used by Moonshot/Kimi is still untouched. However, we're converging on a more systematic solution (#1096, draft) that applies schema quirks per exact API host rather than globally scrubbing every OpenAI-compatible endpoint — a global scrub risks changing tool payloads for providers that already accept the richer JSON Schema (DeepSeek, Together, Groq, etc.).

We're closing this PR in favor of that direction. If you'd like to continue, aligning the approach with per-host quirks (#1096) would be very welcome. Thanks again for the clear write-up and tests!

PR #1010, opened 2026-07-21
fix(daemon): make follow-up completion durable

huangzesen:
## Summary

- provision one fresh, one-shot `daemon_common.finish` receipt for every Codex follow-up and require an exact schema/run-bound `done` before accepting output or usage
- keep detached follow-up generation/path ownership in the execution host, then let the supervisor publish success/failure through a per-generation durable claim/receipt with stable idempotency and startup reconciliation
- preserve historical initial-run completion compatibility while preventing legacy released resume rows or unreleased generations from emitting surprise notifications

## Why

Two independent Codex follow-ups exposed adjacent lifecycle gaps: a resumed turn could complete without its own finish receipt, and the detached manager-shaped child could reach a BaseAgent-only notification path. The first candidate fixed the direct faults, but independent R1 review found that strict validation was incomplete and detached publication was still best-effort. This revision closes both recovery boundaries.

## Validation

- `391 passed, 1 warning in 54.94s` — affected daemon/backend/usage/terminal/run-dir/detached/fresh-receipt matrix
- `86 passed` — daemon contract, docs governance, architecture documents, Anatomy drift checker
- `OK: 282 documents (+ docs.yaml itself) validated`
- `git diff --check`
- `python -m compileall -q src tests`
- real Codex CLI `0.144.3` accepted the generated `-c mcp_servers.daemon_common...` configuration

## Independent review

- R1: BLOCK on incomplete finish-schema validation and non-durable detached publication
- R2 (exact 13-file frozen candidate): **PASS — no P0/P1 findings**
- R2 patch SHA-256: `d69fe5c63477c35c69253869eb762a5e2a639cd2b838902573e1b755dee29d04`
- R2 manifest SHA-256: `07c892d55c1c651c035ebbac13777331003c6dea9131102ca5b651c8c6aa3be5`

## Notes

The Anatomy drift checker still reports the unchanged pre-existing nudge citation (`init_config.py:1-80` versus 74 lines); this diff does not touch that surface.


huangzesen:
Closing per Jason's 2026-07-27 Daemon one-pass disposition: this stale branch will not ship as-is. If the product direction returns, rebuild the narrow behavior on current main under a fresh contract/review. Source branch and evidence are intentionally preserved; no branch deletion.

PR #1021, opened 2026-07-24
fix(daemon): recover once from text-only completion

runyuan-wang:
## Summary

- add one bounded same-session recovery when a LingTai daemon returns non-empty text, no tool call, and no valid current-run `daemon_common.finish` record
- preserve the existing strict completion gate: only a validated current-run `finish(status="done")` can mark the daemon done
- preserve failure evidence and existing max-turn/cancellation/explicit `failed` or `incomplete` behavior
- document the recovery event/chat record and add focused regression coverage

## Why

Three observed MiMo daemon runs reached a milestone, emitted a non-empty progress sentence, and returned no tool calls. The LingTai loop therefore left `while response.tool_calls`; the existing strict completion gate then correctly rejected success because no current-run `finish` artifact existed.

The failure gate is doing the right thing. The missing behavior is one bounded opportunity for the same session to continue from its durable state and call the already-required completion tool.

## Behavior

The recovery is eligible only when:

- the response has non-empty text and zero tool calls
- no valid completion status is present
- `daemon_common` is registered for the run
- cancellation is not set
- an ordinary continuation slot remains
- the invocation-local latch has not already been used

The latch is set before the recovery send. The runtime records one `daemon_completion_recovery` event and one `completion_recovery` chat entry, names the current run ID, tells the model not to repeat successful tool calls, and requires exactly one `finish` signal after continuing.

A second text-only response receives no second recovery. Blank text, valid `done`, explicit `failed`/`incomplete`, cancellation, missing `daemon_common`, and turn-ceiling cases keep their existing fail-closed paths. This patch does not grant an unbounded or free turn at the ceiling.

## Validation

Fresh official-main base: `67fbf5aea631f22a0719ec7c26e47638b3a5fc0a`

- 11/11 focused recovery regressions pass, including the two post-review ToolExecutor/max-turn boundary pins
- failing-first check on the unpatched base: 6 new-behavior tests fail and 3 existing-boundary regression pins pass
- daemon contract + architecture-document tests: 7 passed
- focused daemon-family subset: 312 passed
- full `tests/test_daemon*.py tests/test_lifecycle_daemon_shutdown.py tests/test_architecture_documents.py`: **718 passed, 7 skipped** (the skips are native-Windows-only)
- `git diff --check`: pass
- exact scope: five files only

## Evidence boundary

A real MiMo prompt-level harness reproduced text-only output and showed that the recovery prompt can elicit `finish(done)` in the same LLM session. That harness did **not** execute this patched `DaemonManager`, so it is evidence for prompt efficacy, not a full runtime integration proof. Its attempted second-text-only negative case was inconclusive because MiMo called `finish` instead.

Two deterministic post-review tests now exercise the candidate `DaemonManager` and real `ToolExecutor` loop with a fake task-MCP surface: an in-budget recovery-generated `finish` dispatches exactly once without replaying a prior successful tool, while a final-slot recovery tool batch is not dispatched and the existing max-turn failure wins. A proof with a separately started local `daemon_common` MCP process remains valuable follow-up evidence. The raw provider stop/status fields are also not currently persisted, so natural model stop versus a provider/wire-side stop remains unknown; provider observability is intentionally outside this patch.

This PR does not merge or apply the change to a running LingTai runtime.

### DIFF
--- a/src/lingtai/tools/daemon/CONTRACT.md
+++ b/src/lingtai/tools/daemon/CONTRACT.md
@@ -8,7 +8,7 @@ description: >
   terminal notifications, and compaction boundaries.
 status: active
 contract_version: 6
-last_changed_at: "2026-07-24"
+last_changed_at: "2026-07-26"
 related_files:
   - src/lingtai/tools/daemon/ANATOMY.md
   - src/lingtai/tools/daemon/__init__.py
@@ -42,6 +42,7 @@ related_files:
   - tests/test_daemon_run_dir.py
   - tests/test_daemon_codex_usage.py
   - tests/test_codex_standalone_compaction.py
+  - tests/test_daemon_detached_supervisor.py
   - tests/test_daemon_windows_lock.py
   - tests/test_daemon_windows_process_port.py
   - tests/test_daemon_windows_supervisor.py
@@ -228,6 +229,14 @@ When `daemon_common` is loaded, a conversational final answer is not enough.
 Success requires a validated `finish(status="done")`; missing completion,
 invalid JSON, invalid status, run-id mismatch, `failed`, or `incomplete` must
 prevent terminal `done`.
+
+For LingTai-backend runs only, a nonblank text-only response with no valid
+completion may receive one same-session recovery prompt when enough turn budget
+remains to dispatch a returned tool batch. Recovery never synthesizes
+`daemon_completion.json`, never turns text into success, never reinterprets
+`failed`/`incomplete` as success, and does not change the detached execution
+host's local canonical `finish` surface or its filtering of the reserved
+external `daemon_common` registration.
 
 ### 4. Artifacts separate review evidence from secret-bearing config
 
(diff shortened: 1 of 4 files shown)


huangzesen:
Thank you for the detailed diagnosis and regression coverage here. The patch correctly demonstrates the old runtime behavior: a non-empty progress sentence with no tool call could reach the strict `daemon_common.finish` gate without another opportunity to emit `finish(done)`.

Since this PR was opened, #1055 was merged with a deliberately different product contract. It replaced the earlier daemon-only one-shot finish cue with main/daemon parity **only for truly all-empty responses** (no text, no tool calls, and no thoughts); its accepted design explicitly records the earlier finish-specific cue as not semantic parity. Current main also has regressions requiring that the old finish-specific recovery surface remain absent.

We rechecked both sides before closing: the focused text-only recovery tests on this exact head still pass, while current-main all-empty parity/Contract tests and a direct current-main probe confirm that non-empty progress without `finish` remains a one-send, fail-closed completion error by design. Rebasing this patch would therefore reintroduce intentionally retired behavior across the later detached-daemon and ToolFamily changes, rather than merely resolve drift.

Closing this PR unmerged rather than asking you to rebase it. If the product direction later changes to make non-empty progress-without-finish recoverable, that should be a fresh current-main proposal and Contract discussion. Thanks again for making the original failure mode and its boundaries concrete.

PR #1027, opened 2026-07-25
fix(websearch): opt MiniMax into safe MCP retry

runyuan-wang:
## Summary
- opt the built-in MiniMax stdio `web_search` call into `retry_policy="safe"`
- add a focused fake-client regression test that asserts the exact `call_tool(..., retry_policy="safe")` shape
- update services anatomy docs to describe the stdio-only retry policy boundary after #924

## Rationale
#924 correctly made stdio MCP stale-resource replay fail closed by default (`retry_policy="never"`). The MiniMax web-search service is a narrow read-only/replay-safe call site: replaying a stale transport repeats a query, not a write/send operation. Without an explicit opt-in, this built-in search path can restart the stale transport for future calls but return `[]` for the current user search.

This PR preserves the fail-closed default and opts in only the concrete MiniMax `web_search` call site.

## Tests
- Red: base implementation + new focused test only fails as expected; actual call lacks `retry_policy="safe"`
- Green: focused test passes on this patch
- `python3 -m pytest tests/test_web_search_capability.py tests/test_mcp_closed_resource_restart.py tests/test_mcp_structured_result.py -q` → 72 passed
- `git diff --check`

## Non-goals
- no change to generic Agent MCP call paths
- no change to daemon task-scoped MCP handling
- no change to Telegram task-card projection
- no change to `HTTPMCPClient` / Zhipu HTTP search
- no change to stale-resource classification, timeout retry behavior, locking/concurrency, or the MiniMax MCP server


huangzesen:
### Closing because the target product path has been superseded

This pull request proposed restoring a built-in MiniMax web-search admission path. Current `main` has since made the opposite owning product decision: the built-in MiniMax provider path was retired and non-canonical providers are routed through third-party MCP integration instead.

**Why this PR no longer has a current implementation target**

#1027 only opts a built-in MiniMax web_search call into safe retry. Current main explicitly says MiniMax/Zhipu are no longer built-in web providers by product decision and rejects retired provider selection in favor of third-party MCP. The proposed built-in call site is therefore obsolete on current main.

**Current-main evidence**

- #1027 is limited to opting the built-in MiniMax stdio web_search call into retry_policy=safe. ([source](https://github.com/Lingtai-AI/lingtai-kernel/pull/1027))
- web_search __init__.py says MiniMax and Zhipu are no longer built-in web providers, identifies them as retired by product decision, and directs callers to a third-party MCP route. ([source](https://github.com/Lingtai-AI/lingtai-kernel/blob/main/src/lingtai/tools/web_search/__init__.py))

**Scope note and future path**

This closure does not reject the use case itself; it records that this particular implementation targets a product path that no longer exists. If built-in MiniMax support is reconsidered later, it should start from a new product decision against the current provider architecture rather than reviving this stale patch. Thank you for the investigation and implementation effort.

PR #1029, opened 2026-07-25
refactor(daemon): remove dead parent dispatch scaffolding

huangzesen:
## Summary

- remove the parent `ThreadPoolExecutor` / future / watchdog dispatch scaffolding that became unreachable when daemon detachment was made unconditional in `cdc641cd`
- keep the detached supervisor/run-dir path, live `_run_emanation` execution-host forwarding, event/record emission, terminal notifications, and backend/MCP/preset/timeout plumbing unchanged
- migrate two per-batch timeout tests from mocked dead watchdog arguments to the durable detached-run manifest that the supervisor actually consumes
- preserve the CLI `handoff` response key while consolidating the formerly always-taken detached return

## Scope

This is the report's K02 slice only. It does not include other simplification candidates or deferred contract work.

The now-empty general `_pools` cleanup surface plus `_watchdog`, `_arm_batch_done_cancel`, `_on_emanation_done`, and their pre-existing ANATOMY descriptions are intentionally left for a separate coherent follow-up rather than expanding this PR.

## Validation

- `git diff --check`
- `python -m py_compile src/lingtai/tools/daemon/__init__.py tests/test_daemon_per_batch_limits.py`
- parent focused gate: 60 passed across per-batch limits, detached supervisor, CLI watchdog scope, lifecycle shutdown, and architecture documents
- Sonnet implementation gate: 524 passed across the broader daemon/supervisor surface
- independent Opus 5 exact-diff final: APPROVE / no blockers; foreground review commands passed 57, 369, 129, and 12 tests respectively

No merge is requested or authorized by this PR creation.

### DIFF
--- a/src/lingtai/tools/daemon/__init__.py
+++ b/src/lingtai/tools/daemon/__init__.py
@@ -4128,15 +4128,6 @@ class DaemonManager:
                 "preset_handlers": preset_handlers,
             })
 
-        cancel_event = threading.Event()
-        # Separate event so the watchdog can distinguish timeout from manual
-        # reclaim. Watchdog sets BOTH on timeout; reclaim sets only cancel_event.
-        # The run loop checks timeout_event first to call mark_timeout vs
-        # mark_cancelled.
-        timeout_event = threading.Event()
-        pool = ThreadPoolExecutor(max_workers=len(tasks))
-        self._pools.append((pool, cancel_event))
-
         ids = []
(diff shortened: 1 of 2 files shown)


huangzesen:
Closing per Jason's 2026-07-27 Daemon one-pass disposition: this behavior is already carried by current main via the reviewed aggregate path, so the stale slice must not be replayed. Source branch and evidence are intentionally preserved; no branch deletion.

Issue #751, opened 2026-07-06
goal nudge check does unthrottled disk I/O on every 1s heartbeat tick while IDLE, violating the run_checks contract


huangzesen:
## Summary

`run_checks` in `src/lingtai_kernel/nudge/__init__.py` documents a contract: "each check throttles itself and short-circuits before any disk I/O." The `kernel_version` and `source_drift` checks honor it with a `last_probe_ts` guard and a 60s interval, but `goal.check` (`src/lingtai_kernel/nudge/goal.py`) has no time-based throttle at all. After a cheap in-memory IDLE gate it calls `collect_notifications` — which globs and JSON-parses every file in `.notification/` — on **every 1-second heartbeat tick** for as long as the agent sits idle, and on the no-active-goal path it does a **second** full directory read inside `_clear_goal_reminders_locked`.

## Narrative

Every running agent has a heartbeat loop (`src/lingtai_kernel/base_agent/lifecycle.py`) that ticks once per second (`time.sleep(1.0)` at the bottom of the loop, around line 611). One of the things each tick does is dispatch the nudge checks (around line 523):

```python
# Each check throttles itself; the dispatcher wraps individual calls
# so a misbehaving check cannot block the heartbeat loop.
from ..nudge import run_checks as _run_nudge_checks
_run_nudge_checks(agent)
```

The dispatcher's own docstring (`src/lingtai_kernel/nudge/__init__.py:44-50`) makes the design contract explicit:

```python
def run_checks(agent) -> None:
    """Run all registered nudge checks for this agent.

    Cheap on the steady path: each check throttles itself and short-
    circuits before any disk I/O. ...
    """
```

Two of the three registered checks live up to this. `kernel_version.check` (`src/lingtai_kernel/nudge/kernel_version.py:55-58`) starts with:

```python
now = time.time()
if now - float(state.get("last_probe_ts") or 0.0) < _FAST_INTERVAL_SECONDS:
    return
state["last_probe_ts"] = now
```

and `source_drift.check` (`src/lingtai_kernel/nudge/source_drift.py:33-36`) is identical in shape, both with a 60-second interval. On 59 out of 60 ticks those checks cost one `getattr` and one float comparison.

`goal.check` is the outlier. Its only gate is state-based, not time-based (`src/lingtai_kernel/nudge/goal.py:27-31`):

```python
# Cheap gate before touching disk: goal reminders are IDLE-only.
if getattr(agent, "_state", None) != AgentState.IDLE:
[...]

PR #1036, opened 2026-07-26
fix(feishu): keep SDK logs off MCP stdout

ZacharyHu0:
## Summary

- keep Feishu/Lark SDK log records off MCP stdout by retargeting SDK stdout handlers to stderr
- preserve handler order, formatter, filters, level, and other logger configuration
- make the adjustment idempotent and reapply it during lazy SDK import
- add regression coverage proving stdout remains clean while SDK diagnostics remain visible on stderr

## Validation

- `PYTHONPATH=src python -m pytest -q tests/test_feishu_stdio_logging.py tests/test_feishu_notification_metadata.py` — 6 passed
- `python -m py_compile src/lingtai/mcp_servers/feishu/account.py tests/test_feishu_stdio_logging.py`
- `git diff --check`

Fixes Lingtai-AI/lingtai#751

### DIFF
--- a/src/lingtai/mcp_servers/feishu/account.py
+++ b/src/lingtai/mcp_servers/feishu/account.py
@@ -11,6 +11,7 @@ import asyncio
 import json
 import logging
 import os
+import sys
 import tempfile
 import threading
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


ZacharyHu0:
Friendly ping @maintainers — this PR has been open for a bit and the bug it fixes (#751, lark SDK logging to stdout corrupting the MCP stdio channel) is still present in the v0.18.2 release. The fix is minimal, tests pass, and it had an independent review with no defects found. Happy to rebase or address any feedback — thanks for taking a look!

huangzesen:
Thank you for reporting this and for narrowing the fix to the Feishu/Lark owner layer. The diagnosis is valid, and the core idea here—moving the SDK's existing stdout handlers to stderr—is the right direction.

We are closing this PR as superseded by #1095. The replacement keeps that handler-routing fix, but deliberately avoids the process-global `Lark.propagate = False` mutation so the host logger's propagation policy and other state remain unchanged. It also adds a real subprocess MCP stdio regression covering `initialize` and `tools/list`, rather than relying only on captured Python streams.

#1095 is still open, so this close does not claim the bug is merged or released yet; `Lingtai-AI/lingtai#751` remains the tracking issue until the replacement lands. If you can test #1095 against the original Feishu setup, that feedback would be very helpful. Thanks again—your report and implementation directly identified the seam the replacement uses.

Issue #164, opened 2026-05-26
Agents can remain alive but unresponsive: long file tools are non-preemptive, and refresh wake can wedge ACTIVE without starting an LLM turn


huangzesen:
# Detailed Issue Report: two different “alive but unresponsive” agent stalls

Date: 2026-05-26
Reporter: `mimo-1`
Requested by: Jason

Affected agents:

1. `lingtaidev1bot` / dev-1 `mimo-1`
   - Path: `/Users/huangzesen/work/projects/lingtai-dev/dev-1/.lingtai/mimo-1`
2. `spiritualblisslingtaibot` / `codex`
   - Path: `/Users/huangzesen/work/projects/lingtai-researches/spiritual-bliss-attractor/.lingtai/codex`

## Executive summary

Two agents looked “alive” but were not responding. They were stuck in **different** ways:

1. **dev-1 `mimo-1` was blocked by a very long built-in `grep` over `/Users/huangzesen/work`.**
   - It issued a broad recursive `grep` at `15:38:44`.
   - The grep did eventually return at `15:56:00`, after `1,036,274 ms` (~17.27 min).
   - During that interval, heartbeat stayed fresh and the process consumed CPU/memory, but the agent could not respond.
   - `clear` and `interrupt` were received during the running grep, but did not abort the tool. The cancellation was only honored **after** the tool returned (`turn_cancelled_post_tool`).

2. **`spiritualblisslingtaibot` was stuck in an “ACTIVE but no LLM turn running” state after a refresh/wake.**
   - At `15:31:55`, it transitioned `asleep → idle → active` and injected a notification pair.
   - After that, there was **no `wake`, no `tc_wake_continue`, no `llm_call`, no tool call**.
   - Heartbeat and Telegram listener stayed alive; new Telegram messages kept arriving, but logs only show `notification_deferred_active` because the runtime believed the agent was already active.
   - A later `clear` at `15:59` fixed it. The successful post-clear wake sequence contained the missing events: `wake`, `tc_wake_continue`, `llm_call`, `llm_response`.

So the initial broad-search diagnosis was correct for dev-1, but it does **not** explain spiritualblisslingtaibot. Jason’s suspicion was correct: the two failure modes are both serious and should be reported together.

## Current recovery status

Both agents were recovered.

### dev-1 `mimo-1`

Recovery sequence:

1. `system.clear` was requested at `15:54:42`.
2. `system.interrupt` was received at `15:55:10`.
[...]

PR #1064, opened 2026-07-27
refactor(file): consolidate file operations into one ToolFamily

huangzesen:
> **Exact source head:** `57241f42e8a09d16eb45236a5ebb59450bb4c258`  
> **Exact base:** `4ad8b1555a5b0368af81aa57292438719a105052`  
> **dev4 exact-head live-use gate:** pending immediately after PR opening; this body will be updated with the durable receipt.

## Decision

Migrate the five public file tools into **one** action-separated `file` ToolFamily, and delete the five old packages outright:

```text
file
├── read
├── write
├── edit
├── glob
├── grep
└── manual
```

Calls use the strict family envelope `action + input + reasoning`, with optional Host `summarize`. **File is the only aggregation migration** in the ToolFamily programme — every other family (`web`, `shell`, `notification`, `knowledge`, `skills`, `mcp`, `soul`, `avatar`, `vision`) keeps its existing public name and merely separates its own actions. File is the one case where several former public roots collapse into a single new one, and it is not a precedent for grouping conceptually related tools.

No operation behavior changed. This is a public-surface and ownership change.

## Behavior

- `read` still returns numbered lines with the exact continuation contract (`truncated`, `cap_chars`, `returned_chars`, `requested_offset`, `requested_limit`, `last_returned_line`, `next_offset`, `remaining_lines_estimate`, `line_truncated`), the 100 000 default / 200 000 non-configurable runtime cap, and the spill-aware expired-artifact hint;
- `write` returns the exact `{status, path, bytes}` receipt; `edit` returns `{status, replacements}` and keeps its not-found / ambiguous-match discipline, leaving the file untouched on both failures;
- `glob` returns the service's sorted matches; `grep` keeps its pushed-down glob filter, `max_matches` cap, and `{file, line, text}` results; both keep the issue-#164 traversal-budget block;
- `manual` returns the installed `file-manual` with no target I/O and no double wrapping (`content[0].text` + `structuredContent.manual_path`); `read-manual` is now a nested parent-owned reference, not a competing action;
- unknown action, non-object input, unknown root field, non-boolean `summarize`, and cross-action input all fail with stable typed errors **before** any handler I/O;
- Chat and Responses preserve root action↔input correlation (6 guarded `allOf` branches; `oneOf`→`anyOf` on `input` is the shared pre-existing Responses rewrite);
- `reasoning` and `summarize` are Host controls and never enter child arguments;
- the family uses the existing raw-first summarizer allowlist — no second summarizer, no new cap.

`file` is declared a **mixed read/write family**: `read`/`grep`/`glob`/`manual` are read-only, `write`/`edit` mutate. There is no ToolGuard, `annotations`/`readOnlyHint`, or per-tool risk field at this base to weaken; the contract records that if such a surface is later added and cannot discriminate per action, the truthful declaration is the strongest child posture (mutating), never read-only. `summarize` guidance is per action — bulky for `read`/`grep`/`glob`, and explicitly false for the `write`/`edit` receipts so summarization can never hide a mutation receipt. The family surfaces no settings file at either level, and the manual says so.

## One owner / no aliases / no shims

- **One package**: `src/lingtai/tools/file/` owns the composed schema, the envelope dispatch, and all five operations (`_read.py`, `_write.py`, `_edit.py`, `_glob.py`, `_grep.py`). No operation module imports another — co-location is ownership, not coupling.
- **25 legacy files deleted**: the `read`/`write`/`edit`/`glob`/`grep` packages in full — 5 `__init__.py`, 5 `CONTRACT.md`, 15 `glossary-{en,zh,wen}.md`. Their per-operation contract bodies (caps, receipts, continuation fields, traversal blocks, exact error strings) were folded into `file/CONTRACT.md` §Per-action behavior, and their zh/wen identifiers into `file/glossary-{zh,wen}.md`, before deletion.
- **No compatibility layer**: the five `_LEGACY_CAPABILITY_ALIASES` entries are deleted. `bash → shell` and `web_search → web` are untouched.
- **No shims**: zero retained `get_schema`/`get_description`/`setup` surface; the RETIRED markers are gone with their packages.
- **Dead machinery removed**: `_GROUPS`, `expand_groups()`, the `Groups:` error-text clause, and all five consumers (`agent.py` list-form and init.json paths, deep-refresh import, `daemon/_expand_requested_tools`, `daemon/_instantiate_preset_capabilities`, `execution_host`).
- **One registry**: the duplicated schema-only and runtime child registries collapse into a single `_build_family(handlers=None)` builder, so the advertised schema and the dispatch registry cannot drift apart. Operations bind once per manager; the per-call `import_module` indirection is gone.
- Also fixed en route: `execution_host.py` gated FileIOService injection on the now-empty `_GROUPS`, which would have broken **every** detached-daemon file write with `'NoneType' object has no attribute 'write'`. Retargeted and regression-tested.

## LOC accounting

Exact diff against `4ad8b1555a5b0368af81aa57292438719a105052`, production reported separately from tests and docs:

| Bucket | Added | Removed | Net |
|---|---:|---:|---:|
| **Production Python (raw)** | 739 | 561 | **+178** |
| **Production executable lines** | — | — | **+37** |
| Tests | 1017 | 296 | +721 |
| Docs / contracts / manuals / glossaries | 643 | 1094 | **−451** |

Executable measurement (docstrings, comments, blanks stripped; AST + tokenizer): the file surface goes 339 → 415 (**+76**) and other touched production goes 8052 → 8013 (**−39**). Fable independently reproduced these numbers with its own measure.

The remaining +76 is the LTP v2 envelope itself — six strict nullable child input schemas carrying property-level teaching text. **Net-negative production is not reachable for this migration without deleting behavior**, because the envelope is additive over an unusually lean base; the honest result is executable-flat (+37) with a 5→1 ownership collapse and docs down 451 lines. Raw production fell from +428 before the cleanup to +178 after; the gap between raw and executable is docstrings on the new modules.

## Behavior changes callers must know

1. **The five old capability names fail loudly.** `read`, `write`, `edit`, `glob`, `grep` are no longer capabilities or aliases. `setup_capability("read")` raises `ValueError: Unknown capability: 'read'. Available: avatar, daemon, file, knowledge, mcp, shell, skills, vision, web.`, and a daemon manifest naming them raises `Unknown tools for emanation`. One honest caveat: at the `Agent.__init__` boundary, `agent.py:306` catches `ValueError` and records a `capability_skipped` event instead of aborting. That tolerance is **pre-existing at this base** and applies to every capability; widening it to an abort would change behavior repo-wide and is out of scope here. The registry and daemon layers — where the directive bites — do raise.
2. **`disable` is family-grained.** `disable=["file"]` removes all six actions. Per-operation opt-out no longer exists: a host can no longer express "read-only file access" via `disable=["write","edit"]`. That is the honest consequence of one public root and is a product decision, not a defect.
3. **Stale per-operation disables silently no-op.** `disable=["edit"]` or `disable=["write"]` now matches nothing and leaves `file` fully enabled — `disable` is an unvalidated opt-out list, and those names merely moved from known to unknown. A stale `disable=["write"]` intended to keep a daemon read-only **silently loses that protection**. Both directions are tested.
4. **The omit-action dual mode is removed.** `action` is required on every call; the pre-migration "omit `action` for the legacy ordinary call" form no longer exists. In-flight pre-migration flat calls fail closed with the generic unknown-tool error (no pending-call alias was added; `kernel/base_agent/tools.py` is untouched and `bash` remains its only alias).

## Validation

- **Opus 5 author gate**: 520 focused tests across 14 suites; full suite 6221 passed / 28 skipped / 6 failed, where 4 are byte-identical pre-existing base failures, 1 is a pre-existing ordering flake that passes in isolation (40/40), and 1 is the unstaged-deletion artifact described below.
- **Fable 5 independent review**: round-1 BLOCK (167 lines of no-caller compat surface, dead `_GROUPS` machinery, five stale operation contracts) → **final ACCEPT** after the full Phase 1–3 collapse plus the alias deletion, conditional only on two one-line truthfulness fixes (F-1 a false comment in `execution_host.py`, F-2 a deleted-`_GROUPS` citation in `daemon/ANATOMY.md`). **Both are applied**; F-1 is comment-only, verified executable-token-identical. Fable re-ran 13 focused suites (402 passed), `test_daemon` (118 passed), byte-compared all five moved handlers against their base versions, and reproduced the schema probe and live smoke independently.
- **Parent isolated gate** (scratch-isolated copy, `scratch/file-isolated-parent-gate-20260727-0431/`, 62-entry SHA-256 manifest): **721 tests passed**, plus docs-governance, glossary and anatomy checks.
- `py_compile` on all changed and new source: pass.
- Glossary validator against the source tree (`PYTHONPATH=src`): **48 resources across 16 packages**, pass — the 5→1 collapse. Without `PYTHONPATH=src` the validator resolves the *installed venv* and reports 54/18; that is a resolution gotcha, not a result.
- Anatomy drift: clean across 55 files. `git diff --check`: clean.
- Live six-action smoke on disposable scratch: write receipt verbatim → numbered read → edit → glob → grep → manual body+path; cross-action input rejected `INVALID_ARGUMENT` before I/O; unknown action `ACTION_REQUIRED`.

### Packaging — read this carefully

- `tests/test_tools_package_data.py` (14 tests, wheel **and** sdist) **passes**, with the expected-package list and glossary counts now derived from the shipped package set rather than hardcoded.
- In the author environment, `python -m build --wheel` produced a wheel containing exactly **10 `lingtai/tools/file/` entries and zero stale entries** from the deleted packages: the six `.py` modules, `CONTRACT.md`, and the three glossaries. `ANATOMY.md` is deliberately absent — the `[tool.setuptools.package-data]` globs ship `*/CONTRACT.md` and the glossaries but not anatomies, which is the pre-existing convention for every tool package.
- The Parent's direct `python -m build` in the isolated copy **did not produce a wheel**: it stopped after `egg_info` due to build tooling unavailable in that environment (`PARENT_WHEEL_BUILD.log`). **This is not a green packaging result and must not be reported as one.** The packaging evidence above is the author-environment build plus the passing package-data suite; an independent wheel build is listed as an outstanding gate below.

### Known-failing, with cause

`test_docs_governance::test_every_in_scope_doc_has_required_fields` fails while the deletions remain unstaged. Its discovery uses `git ls-files`, which still lists the 25 deleted paths from the index. All 21 reported failures are those deleted-package paths and **zero** come from any other document; excluding `git ls-files --deleted`, no live tracked doc is missing from disk. It goes green by construction once the deletions are staged at commit time. Nothing was staged to make it pass.

### Worker incident (disclosed)

An untracked, gitignored `build/` cache in the author worktree still held copies of the five deleted packages and was injecting 25 phantom entries into freshly built wheels, masking the packaging result. Those stale cache entries were removed. This touched only a gitignored build artifact — no tracked file, no source, no evidence directory — and it is disclosed here because it changed a validation outcome rather than any shipped content.

## Outstanding before merge

- Stage the 25 deletions, then re-run `test_docs_governance` (green by construction) and one independent wheel/sdist build confirming the phantom entries drop and the 10 `file/` data files ship.
- Step-5 live exact-head Agent exercise (dev4) per the review contract: actual public action count and Manual usability against the immutable head.

## Rollback

Single squashed commit on top of `4ad8b1555a5b0368af81aa57292438719a105052`; revert that one commit to restore the five packages, their contracts and glossaries, the capability aliases, and `_GROUPS`. No migration, no data change, no config or state written, and nothing outside the repository to undo.

## Base

Exact base: `4ad8b1555a5b0368af81aa57292438719a105052` (`feat(tools): add ToolFamily with Web reference migration (#1059)`). No other tool family is migrated in this PR.


huangzesen:
Closing per Jason's 2026-07-27 Daemon one-pass disposition: this behavior is already carried by current main via the reviewed aggregate path, so the stale slice must not be replayed. Source branch and evidence are intentionally preserved; no branch deletion.

PR #1125, opened 2026-08-01
fix(prompts): remove stale tool names, merge artifact, triple-copied rebuild essay + full slimming audit

huangzesen:
## Summary

Prompt/skills/tests slimming for the kernel. Three previously-separate PRs squashed into this one per request: **#1125 (prompts) + #1127 (sqlite-log-query) + #1128 (nokv-workbench)**. Total **-1,669 lines** across 12 files.

## 1. Stale tool names (correctness, 4 files)
The a-priori \summary=true\ guidance listed \ash\, \
ead\, \grep\, \glob\ as callable tools. The kernel's canonical public capability names are **\shell\** and **\ile\** (read/glob/grep are *file* actions); \ash\ survives only as a hidden input alias and \
ead\/\grep\/\glob\ fail loudly as capability names. Fixed in substrate.md, procedures.md, token_efficiency.md, summarize_best_practice.md.

## 2. Merge artifact
\substrate.md\ read "molt only when context pressure (≥85%), explicit human request, or **human request**, or conversation confusion" — a duplicated \human request\. Removed.

## 3. Triple-copied forced-rebuild essay
0.85/1.0/0.75 rebuild mechanics were near-verbatim in substrate.md (~43 lines), procedures.md (~35 lines), and a catalog file. Both resident copies are now compact action lists pointing to summarize-manual as canonical. (-84 lines for items 1-3.)

## 4. Delete nokv-workbench skill + test surface (-1054 lines)
NoKV is not part of the shipped product (only a placeholder-expansion docstring example existed in agent.py). Verified there is **no live preflight gate** — the skill's PREFLIGHT.md was consumed only by its own tests. Removes: skill (25.5KB + 3 assets), 8 contract tests, test_skills assertions, docstring example. Keeps the generic \_expand_agent_placeholders\ test (renamed neutral).

## 5. Trim sqlite-log-query -47% (-491 lines)
The reference an agent loads to debug runtime problems spent 534 lines on a Trajectory Mining campaign methodology (4 LLM prompt templates, finding schema, digest template, worked examples). Condensed to a 42-line actionable pipeline (SQL-first, cheap daemons, primary-agent triage, human-gated routing). Core query tooling, redaction, pitfalls, scripts all kept. 956 -> 465 lines.

## Tests audit (delivered, no code changes in this PR)
Full audit of the pytest suite (374 files / 145,918 lines / 6,018 tests) is documented in the PR conversation and saved to a working report. Key findings for follow-up PRs:
[...]

PR #1127, opened 2026-08-01
docs(skills): trim sqlite-log-query ~47% by condensing trajectory mining

huangzesen:
## Summary

Trims the sqlite-log-query intrinsic reference by **47%** (956 -> 465 lines, 35,049 -> 18,446 bytes) by condensing the Trajectory Mining section from a 534-line campaign methodology into a 42-line actionable summary.

## Why

The reference an agent loads when debugging a runtime problem (log.sqlite queries, lingtai-agent log doctor|query|rebuild) spent over half its bytes on a Trajectory Mining *campaign* framework: 4 full LLM prompt templates (classifier, anomaly summarizer, observability-gap, cross-run pattern), a finding schema, a validation rubric, an output digest template, two worked examples, and a periodic-mode procedure. That is self-improvement methodology, not kernel tooling - it crowded out the actual query material.

## What stays

Everything needed to query and investigate logs: safety contract, commands, schema quick reference, query recipes, source discovery, investigation workflow, redaction/privacy rules, pitfalls, and the event_summary.py script docs.

## What changes

The Trajectory Mining section is replaced with a 42-line condensed version that preserves the capability and the discipline:
- When to use / not use
- 10-step pipeline: ground the ask, doctor check, manifest, **SQL-only first-pass** (never raw logs to an LLM), chunk+redact, **cheap daemon batch**, primary-agent triage with reproducible-SQL validation, digest, human-gated routing, stop
- Model discipline (tier-1/tier-2 LLM passes; primary agent synthesizes)
- Validation rule: a finding is only reportable if SQL can reproduce it

Keeps the "Trajectory Mining" heading and phrasing pinned by 	est_skills.py catalog assertions.

## Verification

- 	ests/test_skills.py system-manual catalog assertions pass (frontmatter, title, command refs, trajectory mining phrasing).
- The 5 pre-existing 	est_skills.py failures on main are unrelated (web_manual/tilde/psyche/frontmatter) and unchanged.


huangzesen:
Superseded by #1125 (Jason: squash into one PR). All sqlite-log-query trim changes are now on fix/slim-prompts-skills.

PR #1130, opened 2026-08-01
feat(custom): add Responses WebSocket v2 transport

ZacharyHu0:
## Summary

- add an explicit `responses_transport = http | websocket` contract for custom OpenAI-compatible Responses
- keep HTTP as the omission/default path; only explicit `websocket` opens Responses WebSocket v2
- reuse one persistent SDK connection, send a full canonical replay first, then use delta + `previous_response_id` only for a proven strict-prefix continuation
- report unsupported/rejected WebSocket handshakes clearly and never silently fall back to HTTP

## Follow-up protocol and diagnostics fix

Commit [`4f20a262`](https://github.com/Lingtai-AI/lingtai-kernel/pull/1130/commits/4f20a262db795f0faf0ddb346c626dc291e6460a) is a fix on top of the original WebSocket implementation:

- omit the HTTP-only `stream` field from WebSocket `response.create` frames, including when it appears in extra request kwargs
- read provider error fields from both OpenAI SDK objects and mapping/dict payloads
- preserve bounded, single-line `status`/`code`/`type`/`param`/`message` and WebSocket close code/reason diagnostics
- retain the existing replay-safety boundary: only a first-event failure without a continuation id is retried once; established continuations and visible partial output are not blindly resent
- keep the explicit no-HTTP-fallback behavior

This follow-up was prompted by testing against a custom sub2api endpoint. The endpoint intermittently closes otherwise valid standard frames with `1011 upstream websocket proxy failed`; that instability remains upstream, while this fix makes LingTai protocol-conformant and preserves the real upstream failure instead of collapsing it into a generic message.

Follow-up validation: 33 dedicated WebSocket/schema/registry tests and 106 related OpenAI/Responses regression tests passed; Ruff, compileall, docs governance, and `git diff --check` passed.

## Trigger and patch behavior

This feature is accepted only for:

```json
{
  "provider": "custom",
  "api_compat": "openai",
  "wire_api": "responses",
  "responses_transport": "websocket"
}
```

- **Field omitted or `http`:** preserves the existing HTTP Responses path and never calls `responses.connect()`.
- **Explicit `websocket`:** uses `openai[realtime]>=2.22.0`, the Responses WebSocket v2 beta header, and a persistent connection.
- **Unsupported/rejected handshake:** HTTP 400/403/404/405/406/410/426/501 raises an actionable error naming only the configured host and tells the operator to select HTTP.
- **Authentication/network failure:** 401 and non-status network failures remain request errors, not false “unsupported” reports.
- **Expired continuation:** an explicit missing/expired `previous_response_id` rebuilds once from full canonical history.
- **Retry boundary:** a request without a continuation id may reconnect/resend once; an established continuation is not blindly resent. Visible partial streams are terminal-marked to avoid duplicate output.
- **Reset:** closes the socket and clears remote chain state while preserving canonical conversation history.

The schema, preset loader, provider-default builder, registry, and custom adapter all enforce/forward the same exact scope. Official OpenAI and Codex paths do not receive this custom transport option.

## Validation

- [x] affected pytest sweep: **322 passed**, including 30 dedicated WebSocket/schema/registry cases
- [x] Ruff on all changed Python source and the new test file
- [x] `python -m compileall -q src/lingtai`
- [x] `python scripts/check_docs_governance.py --check` — 305 documents plus `docs.yaml`
- [x] `git diff --check`
- [x] real-upstream smoke: one compatible endpoint completed single- and two-turn WS sessions; one rejecting endpoint returned the explicit unsupported/rejected message with no HTTP fallback

## Notes

- Base: `fe644d069c68e5417f05ab35d8f2904ceee75123`.
- This PR does **not** include the custom reasoning-effort feature or its UI.
- It does **not** restore local context-overflow, timeout/watchdog, worker, `/clear`, or WS-to-HTTP fallback patches.
- The only added `thinking_tokens` reference is response usage accounting, not reasoning configuration.
- A broader architecture composite still has unrelated baseline failures involving a missing supervisor file, the temporary MCP API version, and a stale Anatomy citation anchor; changed-area governance passes.
- Logs/examples contain no credentials or private paths.

## Companion PR

- TUI transport selector: https://github.com/Lingtai-AI/lingtai/pull/779


ZacharyHu0:
Follow-up fix pushed in [`4f20a262`](https://github.com/Lingtai-AI/lingtai-kernel/pull/1130/commits/4f20a262db795f0faf0ddb346c626dc291e6460a).

This fixes two LingTai-side issues found while testing the custom Responses WebSocket path against Jihu:

1. WebSocket `response.create` no longer includes the HTTP-only `stream` field required to be omitted by the WS v2 protocol.
2. Error handling now supports both SDK objects and mapping/dict payloads and retains bounded close diagnostics, so a server close such as `1011 upstream websocket proxy failed` is visible instead of becoming a generic error.

The Jihu endpoint still intermittently returns 1011 for otherwise valid standard frames, so that random proxy failure remains upstream. This patch fixes LingTai's wire conformance and observability; it does not add a silent WS-to-HTTP fallback or unsafe replay.

Validation for the fix: 33 dedicated WebSocket tests + 106 related OpenAI/Responses regression tests, Ruff, compileall, docs governance, and `git diff --check` all pass.

Companion TUI PR remains #779.

huangzesen:
## Maintainer review note — prior-art context for this decision

Thanks for the careful follow-up fixes (@ZacharyHu0). Before this gets merged we want to share the relevant prior art from our own transport history, since it directly shaped the acceptance bar for a WebSocket Responses path.

### The June Codex WS train (#473–#502)

We built, stabilized, default-enabled, and then **demoted** a WebSocket transport for the Codex backend in June:

- **#473** gated WS session reuse (one connection per session, `previous_response_id` + delta).
- **#474** froze `function_call_output.output` per `call_id` because resident meta/guidance blocks mutate between replays and break the WS strict-prefix chain.
- **#475** added epoch resets (default every 10 successful WS responses) because any local history mutation invalidates the incremental assumption.
- **#481** forced a fresh epoch on notification dismiss/cleanup — the same class of rewrite that `summarize` triggers.
- **#482** default-enabled WS.
- **#486** added a WS cache ledger for diagnostics.
- **#502** (the decisive one) made **REST the hardcoded/default transport** and kept WS only as an explicit constructor/test/internal path.

**Why #502 demoted it:** continuation correctness was tied to an exact strict prefix + remote response chain; resident metadata/notification changes, tool-result relationship changes, compaction, `/clear`, epoch resets, reconnects, or an expired remote chain could invalidate the delta — and a stale delta can be rejected, associated with the wrong epoch, or duplicate continuation semantics. We had to maintain a second state machine (epochs, cache-prefix labels, reset cadence, full-vs-incremental) and reconnect/retry was not safely transparent (a disconnected request may have been accepted remotely; blindly resending established continuations can duplicate tool calls/output). REST avoided the remote-chain dependency while still giving SSE streaming and full canonical replay.

### Does that apply here?

Partly. The strict-prefix/`previous_response_id` fragility is Codex/backend-specific — a generic custom provider may have a different continuation contract or none at all, so the Codex diagnosis should **not** be overgeneralized into “every WS transport is wrong.”

The **operational lessons are generic**, however: handshake/proxy compatibility, connection lifetime, reconnect and replay boundaries, partial-output duplication, cancellation/reset semantics, error observability, and whether the provider’s WS framing actually matches the SDK/API contract. These all remain acceptance criteria for any custom Responses WS adapter.

### What this PR gets right

The current proposal is materially narrower than the old Codex path and directly addresses several historical failure modes:
- HTTP is the omission/default; WS requires explicit `responses_transport=websocket`.
- No silent HTTP fallback — unsupported handshakes are reported clearly.
- Starts with a full canonical replay; delta/`previous_response_id` only for a proven strict-prefix continuation.
- Bounded retry/reset behavior; does not touch official OpenAI/Codex paths.

### What’s still missing before I’d accept it

1. **A demonstrated product benefit.** The description does not yet show a measured latency/throughput/connection-reuse improvement over the existing HTTP streaming path on a real supported endpoint. We need a reproducible benchmark on a supported custom endpoint — otherwise this adds a second transport state machine with no user-visible win.
2. **Upstream compatibility evidence.** The smoke test reports intermittent `1011 upstream websocket proxy failed` closes for otherwise valid frames (follow-up improves conformance/diagnostics but does not eliminate upstream proxy risk). That needs a real-endpoint pass.
3. **No-duplication proof.** Continued proof that reconnect/partial-output/reset behavior cannot duplicate work under real failure injection.

If the team wants this for a specific provider, the explicit opt-in shape is the right boundary — custom-only, HTTP default, no fallback, clear diagnostics, and no claim that WS improves Codex/cache behavior. I’d treat it as a narrowly scoped, documented, experimental/provider-specific feature rather than a general transport recommendation.

Happy to discuss; appreciate the iteration quality here.

huangzesen:
Thank you for the follow-up fixes (@ZacharyHu0, 4f20a262) — we reviewed them. Decision: rejecting for now. Current benefit of the Responses WebSocket v2 transport is limited for the project at this stage (maintenance/endpoint complexity outweighs the marginal latency/streaming gain we can realize today). We may revisit if a concrete consumer emerges; please feel free to reopen with a clear use case. Credit for the careful work is preserved; the branch remains untouched.

huangzesen:
Rejected as decided by maintainer (Jason): limited near-term benefit for Responses WebSocket v2 transport; revisit if a concrete use case appears.

huangzesen:
Thanks for the detailed Responses WebSocket v2 work, @ZacharyHu0 — the wire research and design were valuable.

Per maintainer review, this PR is superseded by a smaller follow-up: instead of a full custom-provider WS v2 transport (which would require an all-repo `openai[realtime]>=2.22.0` hard floor), the Codex adapter now exposes its existing WebSocket transport as an **environment-variable opt-in** (`LINGTAI_CODEX_TRANSPORT=websocket` / legacy `LINGTAI_CODEX_WS=1`), with an llm-adapters manual documenting it. The shared `_consume_responses_stream` accumulator merged via #1175 is reused; no new hard dependency is introduced.

Closing as superseded — head `4f20a262` and the branch are preserved. Credit for the WS direction is retained in the replacement PR.

PR #1131, opened 2026-08-01
feat(custom): support Responses reasoning effort

ZacharyHu0:
## Summary

- allow explicit `manifest.llm.thinking` for custom OpenAI-compatible Responses
- accept the existing effort values `none | minimal | low | medium | high | xhigh`
- represent custom `default` by omitting the field, preserving the existing custom runtime default of `high`
- preserve the existing Codex omission/default behavior of `xhigh`

## Trigger and patch behavior

The added scope is exactly:

```json
{
  "provider": "custom",
  "api_compat": "openai",
  "wire_api": "responses",
  "thinking": "minimal"
}
```

- **Explicit supported effort:** init and preset validation accept it; config hydration and session materialization preserve it as `reasoning.effort`.
- **Thinking omitted:** no new normalization occurs. Custom Responses keeps the existing `high` main-session default.
- **Literal `thinking: "default"`:** remains invalid. The UI companion implements default through field omission.
- **Out-of-scope provider/wire mode:** validation fails instead of silently retaining an ineffective field.
- **Codex/Codex pool:** existing values and adapter-owned `xhigh` default remain unchanged.

## Validation

- [x] changed-area pytest files: **257 passed**
- [x] real materialization path proves custom omission emits `reasoning.effort=high`
- [x] real materialization path proves explicit `minimal` emits `reasoning.effort=minimal`
- [x] Ruff on changed source
- [x] `python -m compileall -q src/lingtai`
- [x] `python scripts/check_docs_governance.py --check` — 305 documents plus `docs.yaml`
- [x] `git diff --check`

## Notes

- Base: `fe644d069c68e5417f05ab35d8f2904ceee75123`.
- This PR contains no `responses_transport`, WebSocket dependency, or connection code.
- Modified legacy test files retain pre-existing unused-import/E402 Ruff findings; unrelated cleanup is intentionally not mixed into this feature. Changed source passes Ruff.
- Broader architecture composite failures are unrelated baseline issues; focused governance passes.
- Logs/examples contain no credentials or private paths.

## Companion PR

- TUI custom Responses reasoning selector: https://github.com/Lingtai-AI/lingtai/pull/780

PR #1132, opened 2026-08-01
fix(openai): handle forced SSE Responses bodies

ZacharyHu0:
## Summary

- handle OpenAI-compatible Responses gateways that ignore a non-streaming request and return a completed SSE body as `str`
- parse the already-returned SSE body through the existing Responses streaming accumulator, without retrying or issuing a second billable request
- share event consumption between normal streaming and the compatibility path
- cover reasoning summaries, text, function calls, usage, canonical history, and the single-request invariant

## Root cause

Some compatible gateways(e.g. sub2api) return `Content-Type: text/event-stream` even when the request is non-streaming. The OpenAI Python SDK exposes this unexpected response as a plain string. `OpenAIResponsesSession.send()` then passed it to the ordinary response parser, which attempted `raw.output` and raised:

```text
'str' object has no attribute 'output'
```

The fallback is deliberately narrow: only valid SSE framing is decoded. Plain strings and malformed SSE fail loudly and retain the existing stateless rollback behavior.

## Tests

- `171 passed` across the OpenAI/Responses/wire-api targeted suites
- `git diff --check`
- OpenAI Anatomy drift check passes

The repository-wide architecture checks still report four pre-existing `origin/main` documentation drift items outside this change: missing `kernel/daemon_supervisor/supervisor.py` and stale citations in the i18n, notification-store, and nudge Anatomy files.

### DIFF
--- a/src/lingtai/llm/openai/ANATOMY.md
+++ b/src/lingtai/llm/openai/ANATOMY.md
@@ -22,6 +22,7 @@ related_files:
   - tests/test_codex_prompt_cache_key.py
   - tests/test_codex_native_multiaccount.py
   - tests/test_codex_standalone_compaction.py
+  - tests/test_custom_responses_stateless.py
   - tests/test_mimo_responses_compaction.py
   - ENVIRONMENT_VARIABLES.md
 maintenance: |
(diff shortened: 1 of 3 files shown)


huangzesen:
Closing as superseded by PR #1175 (feat(custom): Responses reasoning effort + forced SSE fix, merged 2026-08-04 as 3915dad6), which squashed this PR together with #1131. Content is fully merged on main; credit preserved to ZacharyHu0. Branch fix/responses-forced-sse-nonstream preserved at head 4bebef71.

PR #1138, opened 2026-08-02
refactor(feishu): adopt lark channel sdk

ZacharyHu0:
## Series

- Feishu Bot v1: **1/15**
- Tracking issue: #1139
- Follows: none (series root)
- Next planned after this PR merges: `feat(feishu): normalize inbound conversations`
- This description will be back-linked to the next PR once that PR is opened.

## Summary

- replace the Feishu transport dependency with `lark-channel-sdk>=1.2,<2`
- adapt the existing raw event dispatcher, generated REST requests, and per-account WebSocket lifecycle without changing the public 12-action tool surface
- keep MCP stdio stdout protocol-only and explicitly run the SDK in `security.mode="compat"`
- add focused coverage for the generated request shapes and the low-level SDK's private disconnect path

## Validation

- [x] `git diff --check`
- [x] `python -m pytest -q tests/test_feishu_toolfamily_ltpv2.py tests/test_feishu_channel_sdk_account.py tests/test_feishu_stdio_logging.py` — 13 passed
- [x] Existing allowlisted Bot gray rollout: startup, DM send/receive, reply, delete, reaction, seen/done feedback, stop, refresh, and post-refresh inbound delivery
- [x] Installed SDK version and committed `account.py` bytes verified before restoring the supervised Bot process

## Notes

- Public actions, account config, on-disk state, compound message IDs, and raw inbound callback shape are unchanged, so this PR does not update the Feishu usage skill or Anatomy/Contract documents.
- Outbound calls continue to use the generated low-level REST client, which performs one request and does not hide retries.
- The existing text-edit PATCH returns Feishu error `230001` for non-card messages. The same result was reproduced with both old and new SDKs; this refactor deliberately preserves that baseline and leaves the endpoint change to the rich-outbound slice.
- Gray evidence contains only scenario/result summaries. No credentials, IDs, original chat content, or raw logs are included.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     # here.
     "imapclient>=3.0.0",
     "msal>=1.37.0",
-    "lark-oapi>=1.4.0",
+    "lark-channel-sdk>=1.2,<2",
     "cryptography>=42.0.0",
     "qrcode>=7.4.0",
     "faster-whisper>=1.0.0",
--- a/src/lingtai/mcp_servers/feishu/account.py
+++ b/src/lingtai/mcp_servers/feishu/account.py
@@ -1,6 +1,6 @@
 """FeishuAccount — single app credential, WebSocket listener + REST sender.
 
-One daemon thread per account runs the lark-oapi WebSocket client.
+One daemon thread per account runs the lark-channel-sdk WebSocket client.
 Constructor stores config only — no connections, no threads.
 start() spawns the WebSocket thread and initialises the REST client.
 stop() signals the thread to stop and joins it.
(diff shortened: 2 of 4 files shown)


ZacharyHu0:
Next Feishu v1 slice: #1143. The continuing linked chain and status are tracked in #1139.

ZacharyHu0:
Post-refresh gray follow-up: the live Agent completed a watcher-managed process
replacement and Feishu identity re-verification. A new `reconnect-ok` message
then arrived through the refreshed listener, and the Bot produced a same-chat
response; the exact probe was deleted afterward. This supplies real inbound
event evidence for WebSocket recovery rather than inferring it from process or
socket state.

Provider reconciliation after that refresh also found exactly one Bot-owned
Task Card for each persisted route (one flat and one threaded), and each was the
exact stored resident. No duplicate card was created across restart.

No process ID, actor/chat/thread/message identifier, or raw envelope is included.

ZacharyHu0:
Final same-version compatibility follow-up: all legacy 12 public actions passed
one redaction-safe gray run against the deployed SDK implementation.

- `check`, `read`, and `search` returned the live allowlisted conversation state,
  including the known post-refresh marker.
- `contacts` succeeded; one uniquely named temporary contact was added and
  removed, and the original contact count was restored.
- `accounts` returned the configured non-secret account summary and `manual`
  returned the bundled skill body.
- Legacy text `send` returned a compound id, `edit` updated it, `reply` created a
  flat reply, and both physical provider probes were deleted by their exact ids.

Every expected public status matched. No credentials, actor/chat/message ids,
contact values, original chat content, or raw envelopes are included.

ZacharyHu0:
Superseded by maintainer integration #1176, merged as `113833b4`. A post-merge blob audit found upstream `main` identical to the complete #1172 + #1173 train on all 55 touched paths. Tracking: #1139. Closing this review PR; its branch is preserved.

PR #1146, opened 2026-08-02
feat(feishu): record channel events

ZacharyHu0:
## Summary

- Register SDK-normalized reaction, message-read, Bot-added, and Bot-leave events with `reaction_notifications="all"`.
- Reuse each account's `allowed_users` gate, retain complete raw envelopes, derive stable event IDs when the provider ID is absent, and deduplicate reconnect redelivery.
- Persist concise projections in the reserved read-only `chat_id="events"` conversation and expose them through existing `read` and `search` actions.
- Keep passive events out of LICC entirely so they cannot wake the agent or replace a pending human-message notification mirror.
- Continues the Feishu v1 series tracked in #1139.

## Validation

- Deployed first to the allowlisted Feishu gray environment and refreshed before adding focused tests.
- Real Feishu `message_read` events were received and persisted with their raw envelope in the reserved events conversation.
- Live testing found that publishing a passive event with `wake=false` could still replace `.notification/mcp.feishu.json`. After the fix, a normal text was received and answered, a later read event persisted, and no LICC inbox or Feishu notification mirror was recreated.
- `pytest -q tests/test_feishu_channel_sdk_account.py tests/test_feishu_channel_events.py tests/test_feishu_inbound_media.py tests/test_feishu_notification_metadata.py` — 18 passed; two upstream SDK deprecation warnings.
- Targeted `ruff check` on the changed Feishu adapter and test files.
- `git diff --check`

## Notes

- Stacked on #1145. The review delta is commit `5e369dc4` only.
- Live proof covers `message_read`; reaction and Bot membership projections are locked by focused tests and still require the corresponding Feishu application event subscriptions for live delivery.
- In-place edits to the observed Feishu `todo` did not arrive through the SDK's current normalized message/event set and are not claimed by this slice.
- No credentials, raw chat content, actor/chat/message IDs, or raw envelopes are included in PR evidence.

## Final gray update

Live proof now covers both `message_read` and a real user-added reaction: each was projected into the reserved `events` conversation without a business wake. Bot add/leave remains intentionally focused-test-only because removing the sole gray Bot for a probe would be disruptive.

Issue #1139, opened 2026-08-02
feat: bring Feishu Bot to Telegram parity (v1 series)


ZacharyHu0:
## Problem / need

LingTai's Feishu Bot v1 parity work was developed as a sequence of small,
independently reviewable PRs. Maintainer integration PR #1176 folded the
complete train into upstream `main`; this issue preserves the review and gray
acceptance index.

## Feature series

The boxes track integration into upstream. The original branches remain as
small-slice review provenance; their combined content was merged by #1176.

- [x] 1. #1138 — adopt `lark-channel-sdk`
- [x] 2. #1143 — normalize inbound conversations
- [x] hardening. #1144 — redact WebSocket credentials discovered during gray deployment
- [x] 3. #1145 — preserve inbound media
- [x] 4. #1146 — record channel events
- [x] 5a. #1147 — support rich text outbound and thread-aware replies
- [x] 5b. #1148 — support media/share/sticker outbound
- [x] 6. #1149 — support interactive cards
- [x] 7. #1150 — use native progress feedback
- [x] 8. #1151 — expose reactions and retryable errors
- [x] 9. #1152 — share Task Card event projection
- [x] 10. #1153 — share resident delivery
- [x] 11. #1154 — project automatic Task Cards
- [x] 12. #1155 — project programmable Task Cards
- [x] 13. #1156 — share local command core
- [x] 14. #1157 — add localized control cards
- [x] 15. #1158 — finalize setup and diagnostics

## Audit and review follow-ups

- [x] #1160 — authenticate control card sources
- [x] #1161 — restrict Feishu conversation state permissions
- [x] #1162 — redact Feishu media failure details
- [x] #1163 — sanitize resident Task Card projections
- [x] #1164 — preserve Feishu materialization error taxonomy
- [x] #1166 — surface partial multi-chunk deliveries
- [x] #1168 — bridge callback-loop outbound calls
- [x] #1169 — privatize control callback state
- [x] #1171 — privatize Task Card state
- [x] #1172 — apply edit/delete lifecycle to every physical chunk
- [x] #1173 — document least-privilege outbound media upload

## Upstream integration

Maintainer PR #1176 squashed the complete feature/audit/documentation train
onto current `main` and merged as `113833b4`. A post-merge blob audit over the
union of the #1172 runtime/audit head and #1173 documentation sibling found all
55 touched paths identical: 18 source Python files, 20 test files, and 17
[...]

PR #1147, opened 2026-08-02
feat(feishu): support rich text outbound

ZacharyHu0:
## Summary

- Add a strict `text XOR content` contract for Feishu `send`, `reply`, and `edit`, with tagged `text`, `markdown`, and raw `post` branches while preserving the legacy `text` shortcut.
- Route rich outbound values through the channel SDK so Markdown conversion, safe long-message chunking, text/post editing, and ordered chunk IDs share one path.
- Default replies to the persisted target's topic state, keep every topic chunk in-thread, and fix `reply_target_gone="fail"` so a missing target never becomes a fresh message.
- Persist one logical sent record with exact normalized content, physical chunk projections, reply/thread metadata, and post-success edit state.
- Continues the Feishu v1 series tracked in #1139.

## Validation

- Deployed first to the allowlisted Feishu gray environment, refreshed the existing Bot, and confirmed WebSocket reconnect before adding focused tests.
- Real Feishu API gray testing confirmed the legacy text shortcut, Markdown send and in-place edit, raw post send, flat quote reply, and a long Markdown message split into two observable chunks. Exact content, account/user/chat/message IDs, and provider envelopes are omitted from this evidence.
- `pytest -q tests/test_feishu_rich_outbound.py tests/test_feishu_toolfamily_ltpv2.py tests/test_feishu_channel_sdk_account.py` — 22 passed; two upstream SDK deprecation warnings.
- `pytest -q tests/test_feishu_notification_metadata.py tests/test_mcp_skill_manuals.py` — 32 passed; two upstream SDK deprecation warnings.
- Targeted `ruff check` on the changed Feishu implementation and test files.
- `git diff --check`.

## Notes

- Stacked on #1146. The review delta is commit `16782b95` only.
- Live proof covers flat reply and SDK chunking. Topic-default routing and explicit flat override are locked by focused tests; a real topic-group click-through remains a gray follow-up and is not claimed as live proof here.
- Media upload/share/sticker support remains in the next small PR; interactive cards remain in their dedicated PR.
- No credentials, raw chat content, actor/chat/message IDs, file keys, or raw envelopes are included in PR evidence.

## Final gray update

A later real group-topic acceptance pass closed the earlier topic gap. The admitted message carried thread/parent/root context and the default Bot reply remained in that topic with `reply_in_thread=true`.


ZacharyHu0:
Follow-up capability slice: #1148 adds Feishu media/share/sticker outbound support and is stacked on this PR. The review delta there is commit 967c0f19.

huangzesen:
Read-only follow-up on the current PR head `16782b958297aa8d2b09c795c24d859c59c248cb`: the outbound send path now persists all physical `message_ids`, but the lifecycle operations still consume only one compound ID.

In `src/lingtai/mcp_servers/feishu/manager.py`, `_delete()` parses one `feishu_msg_id` and calls `acct.delete_message(feishu_msg_id)` once; `_edit()` likewise calls `acct.update_content(feishu_msg_id, sdk_message)` once and refreshes only the record found for that ID. For a long Markdown/post/reply split into multiple physical Feishu messages, invoking edit/delete with the primary compound ID can therefore leave trailing sibling chunks visible.

Suggested correction: resolve the logical sent record/chunk group from any returned compound ID, apply edit/delete to every physical member, and persist explicit partial-failure state. Please add two-chunk regression coverage for both primary and secondary IDs. This is a medium lifecycle concern about the current outbound contract; no merge recommendation is implied by this comment.

ZacharyHu0:
Addressed as the independent follow-up #1172 (`000493e1`) so this PR's original
review slice stays intact. The follow-up resolves the logical sent record from
either primary or secondary compound ID, applies edit/delete once to every
physical member, and persists exact mixed outcomes without overwriting the prior
logical content. Its real gray probe exercised seven physical chunks end to end;
focused coverage includes two chunks, both input IDs, both lifecycle operations,
and mixed failures.

[Single-slice compare](https://github.com/ZacharyHu0/lingtai-kernel/compare/7ba4b852...000493e1)

ZacharyHu0:
Real-provider topic follow-up: an allowlisted group topic message arrived with
an explicit Bot mention plus thread, parent, and root context. The Bot's text
answer was persisted with `reply_in_thread=true` and the same thread id. The
automatic resident Task Card also replied inside that topic, while the existing
flat route remained independent. The owner-only route store now contains the
expected flat and threaded bindings.

No actor/chat/thread/message identifiers or chat content are included here.

ZacharyHu0:
Superseded by maintainer integration #1176, merged as `113833b4`. A post-merge blob audit found upstream `main` identical to the complete #1172 + #1173 train on all 55 touched paths. Tracking: #1139. Closing this review PR; its branch is preserved.

PR #1165, opened 2026-08-02
fix(powershell): harden command extractor against allowlist bypasses and false classifications

huangzesen:
## Summary

Hardens the PowerShell command extractor (`src/lingtai/adapters/windows/powershell.py`) against the findings of a 248-case extraction matrix run against the shell policy allowlist/denylist path.

Before: 133 correct / 53 false-positive / 22 false-negative / 17 false-reject / 4 allowlist bypasses.
After: all 4 bypasses fail closed, 14/16 false-rejects fixed, assignment RHS and post-control-word commands visible, literal/type/operator junk removed.

## What changed

**Allowlist bypasses closed (E\u2717)**
- `$x = & $y` no longer extracts an empty set \u2014 the `&`/`.` target is now checked anywhere in a statement, not just at `tokens[0]`; a variable/expandable target fails closed.
- `Invoke-Expression` / `iex` are special-cased as `_UNSUPPORTED` (previously emitted as ordinary commands).
- `Invoke-Command` / `Start-Job` with a non-literal `-ScriptBlock` fail closed; literal `{ ... }` blocks still extract.
- `& (Get-Command foo)` / `. $module` fail closed as dynamic.

**False negatives fixed (C)**
- Assignment RHS is now inspected: `$x = Get-Process` extracts `Get-Process` (was empty set).
- Control words no longer swallow the rest of the statement: `return Get-Process`, `param($Path) Get-Item $Path`, `foreach ($f in Get-ChildItem)` all extract their real commands.
- Indexed/member LHS assignments (`$arr[0] = ...`, `$obj.Prop = ...`) are handled.

**False positives removed (B)**
- Quoted literals, numbers/ranges, operators, type literals (`[System.DateTime]::Now`, `[int]$x`), hashtable keys, switch case labels, format-string braces, comments (`#`, `#region`) are no longer emitted as commands.
- `$(` subexpression scan and `{` scan are now quote-aware; single-quoted here-strings no longer extract `$(Get-Date)`.
- Backtick line continuations no longer split statements; `enum`/`exit`/`default`/`in`/`elseif` added as control words.

**False rejections fixed (D)**
- Method calls with empty parens (`$x.ToString()`, `[Console]::ReadLine()`, `[CmdletBinding()]`) are data, not `_UNSUPPORTED`; `Write-Output ()` still fails closed (contract preserved).
- `@()` array literals accepted; `& { ... }` static script-block targets accepted.
- Here-strings with embedded quotes (`@'...it's...'@`) no longer corrupt the quote state.

## Validation

- New contract suite `tests/test_powershell_extractor_policy.py`: 55 tests (11 dynamic-bypass, 11 assignment/control-word visibility, 15 data-not-command, 11 safe-not-rejected + extras).
- Existing `tests/test_shell_pr1_contract.py`: 20/20 pass (including `Write-Output ()` \u2192 unsupported and backtick-escaped `Rem`ove-Item` fail-closed).
- Broader shell suites: `test_bash_async.py`, `test_layers_bash.py`, `test_shell_tool_family_migration.py`, `test_shell_sandbox_containment.py`, `test_bash_shell_dialect.py`, `test_shell_windows_native.py` \u2014 186 passed, 6 skipped.
- 66-case targeted regression probe: 66/66 pass.

## Notes

- No `pwsh` on the dev host, so the extractor is validated with the pure-Python 248-case matrix; PowerShell execution semantics are unchanged (the invocation wrapper is untouched).
- Known minor limitation: class method names inside `class Foo { [void] Run() { ... } }` may still be emitted as commands (rare in shell-tool scripts); the whole script is never rejected for them.

### DIFF
--- a/src/lingtai/adapters/windows/powershell.py
+++ b/src/lingtai/adapters/windows/powershell.py
@@ -5,6 +5,11 @@ PowerShell statement/pipeline boundaries and recursively inspects command
 substitutions and script blocks.  Unsupported dynamic syntax is represented by
 a sentinel command so a configured allowlist/denylist fails closed; trusted
 (yolo) execution can still pass the original script to pwsh.
+
+The extractor is conservative by design: it only emits statically knowable
+command names, strips comments, honours here-strings and backtick line
+continuations, skips literals/types/operators as data, and fails closed on
+dynamic invocation (call operator on a variable, Invoke-Expression, etc.).
 """
 from __future__ import annotations
 
@@ -40,9 +45,11 @@ _ASCII_BOOTSTRAP = (
     "& ([ScriptBlock]::Create($c))"
 )
 _CONTROL_WORDS = {
-    "begin", "break", "catch", "class", "continue", "data", "do", "else",
-    "end", "finally", "for", "foreach", "function", "if", "param", "process",
-    "return", "switch", "throw", "trap", "try", "until", "using", "while",
+    "begin", "break", "catch", "class", "continue", "data", "default", "do",
+    "else", "elseif", "end", "enum", "exit", "filter", "finally", "for",
+    "foreach", "function", "hidden", "if", "in", "param", "process",
+    "return", "static", "switch", "throw", "trap", "try", "until", "using",
+    "while",
 }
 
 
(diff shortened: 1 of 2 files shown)


huangzesen:
Superseded by #1331. Closing as superseded.

huangzesen:
Superseded by #1331. Closing as superseded.

huangzesen:
Merged via squash PR #1331 (8971f5de). Thanks!\n\nTelegram: @lingtaidev4bot | Nickname: dev4bot\nPowered by LingTai AI: https://github.com/Lingtai-AI/lingtai

PR #1182, opened 2026-08-04
feat(windows): discover pwsh from well-known PowerShell 7 paths

huangzesen:
## What
Mirror OpenClaw resolvePowerShellPath: check %ProgramFiles%\PowerShell\7\pwsh.exe and %ProgramW6432%\PowerShell\7\pwsh.exe before PATH lookup, keeping the pwsh-only policy (never fall back to Windows PowerShell 5.1). Improve FileNotFoundError with searched locations and winget install hint.

## Why
Our Windows shell requires PS 7.3+ for $PSNativeCommandUseErrorActionPreference; machines with pwsh installed but not on PATH now work out of the box.

## Tests
Existing powershell/windows tests pass (list any failures).

### DIFF
--- a/src/lingtai/adapters/windows/powershell.py
+++ b/src/lingtai/adapters/windows/powershell.py
@@ -8,8 +8,10 @@ a sentinel command so a configured allowlist/denylist fails closed; trusted
 """
 from __future__ import annotations
 
+import os
 import re
 import shutil
+import sys
 
 from lingtai.tools.bash._shell_dialect import (
     ShellDialect,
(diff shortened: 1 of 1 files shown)


huangzesen:
## CI blocker: native Windows contract is red at this head

I checked the exact head `e9c48ce27ba9e78ad38f4354c2e411298b64ae23` against base `3915dad603f3b294255e5fa49904aa668ddb65a0`. The failed check is [Native Windows / PowerShell 7](https://github.com/Lingtai-AI/lingtai-kernel/actions/runs/30897418577/job/91953539709), run 57 / job `91953539709`, step `Run shell PR1 contract`.

The runner executed:

```text
python -m pytest -q tests/test_shell_pr1_contract.py tests/test_shell_windows_native.py
```

The exact result was `1 failed, 25 passed in 8.59s`:

```text
tests/test_shell_windows_native.py::test_native_powershell_async_poll_preserves_streams_and_exit_code
assert terminal["exit_status_known"] is True
E       assert False is True
tests\\test_shell_windows_native.py:141
```

The whole PR diff is one file, `src/lingtai/adapters/windows/powershell.py`, `+39/-2`, and the change is limited to PowerShell 7 path discovery. The failing native test is in an unchanged file and does not directly assert the new well-known-path discovery behavior. Therefore the current evidence does not establish whether this is a PR regression or a clean-main/baseline failure, but the native contract is not green at the reviewed head. The companion `Native Windows / kernel lifecycle` check passed; that does not clear this failed contract.

Please either fix the failure or establish a reproducible clean-main comparison, then rerun the same focused command on the final head and retain the attribution in the PR. I did not rerun or mutate anything.

huangzesen:
## Fix pushed: extend the supervisor start lease (e9c48ce..f5646696)

Root cause of the native-contract failure at head `e9c48ce`:

- A detached async supervisor is a **fresh interpreter** that must import the whole bash tool family before it can claim its bounded start lease (`_claim_supervisor_start`). On the loaded Windows CI runner, that cold import exceeded the 3s `_SUPERVISOR_START_LEASE_SECONDS`, so `supervise()` hit the hard deadline and marked the job `unrecoverable` with `exit_status_known: False` ("supervisor start lease expired before claim"); the parent's start loop (`_run_async` → `_mark_unrecoverable_if_supervisor_gone`) would do the same on its side of the window.
- The async poll then returned a terminal result with `exit_status_known: False`, failing `assert terminal["exit_status_known"] is True` at `tests/test_shell_windows_native.py:141`. This is a cold-import timing artifact of the async launch machinery, not a behavior regression from the pwsh path-discovery change in this PR.

Fix (one file, `src/lingtai/tools/bash/__init__.py`, +9/-1):

- Raise `_SUPERVISOR_START_LEASE_SECONDS` from `3.0` to `10.0` with a comment documenting the cold-import reasoning. The return-handoff lease scales with it (`2 * start lease`), and the poll path already waits for the durable supervisor terminal commit within the lease (`_await_supervisor_commit`), so no poll-side change is needed.

Verification run locally (Linux):

- `pytest -q tests/test_shell_pr1_contract.py tests/test_shell_windows_native.py tests/test_shell_tool_family_migration.py` → **56 passed, 6 skipped** (the 6 skips are the native-Windows tests, which skip off-Windows by design).
- `pytest -q tests/test_bash_async.py` → **72 passed** (the start-lease machinery).
- `py_compile src/lingtai/tools/bash/__init__.py` → clean.

I cannot claim the native Windows check is green — the runner must verify on `windows-latest`. This push is intended to rerun the same focused command (`python -m pytest -q tests/test_shell_pr1_contract.py tests/test_shell_windows_native.py`) on the final head.

Attribution: commit `f5646696` — `fix(bash): extend supervisor start lease for cold-import Windows runners`, Signed-off-by: Zesen Huang.

PR #1196, opened 2026-08-05
feat(kernel): preemptive parallel steering for high-priority human messages

rawpaper123:
## Summary
Preemptive parallel steering for high-priority human messages. When a high-priority (human) message arrives while the active turn is inside a tool call, the runtime now dispatches a bounded parallel "steering lane" that answers immediately, instead of queueing behind the long-running turn. If the steering requires interrupting the main turn, the lane writes a `steering_interrupt.json` contract and the main loop aborts at the next safe tool-call boundary.

Refs Lingtai-AI/lingtai#792.

## What was built
- `src/lingtai/kernel/steering.py`: lane dispatch (`daemons/steer-<id>/`), isolated LLM sub-turn runner, reply delivery (hook → telegram MCP → durable fallback), `steering_interrupt.json {reason, by}` contract, boundary abort (Windows `taskkill /T /F` / POSIX killpg), next-turn `MSG_REQUEST` surfacing.
- Intake hook: ACTIVE notification-sync branch dispatches a lane when the active turn is a `tool_call` (defensive getattr keeps stubs on the plain deferral path).
- Main loop: `turn.py:_process_response` checks the interrupt file at the safe tool-call boundary, aborts, marks interrupted.
- Config: `steering.enabled` (default true) + priority channels/budgets on `AgentConfig`, hydrated from the manifest.

## Tests
- `tests/test_steering_lane.py`: 14 passed (lane during tool_call, reply delivery, interrupt at boundary, no-interrupt continues, taskkill argv, marker parsing, default runner).
- Regression `test_notification_sync.py`: 62 passed; 6 failures are pre-existing missing `anthropic`/`httpx` SDKs in the fresh venv (unrelated).

## Remaining work
- Station-layer reply-sink wiring, tool child-pid tracking, E2E with live LLM/Telegram.

### DIFF
--- a/src/lingtai/agent.py
+++ b/src/lingtai/agent.py
@@ -119,6 +119,36 @@ def build_agent_config(manifest: dict[str, Any], *, max_rpm: int) -> AgentConfig
         max_aed_attempts=manifest.get(
             "max_aed_attempts", defaults.max_aed_attempts
         ),
+        # Steering block: per-agent steering.enabled (default true for human
+        # channel) plus optional priority channel / budget overrides. The
+        # kernel default channel set applies when steering.priority_channels is
+        # omitted/empty (see lingtai.kernel.steering.DEFAULT_PRIORITY_CHANNELS).
+        steering_enabled=bool(
+            manifest.get("steering", {}).get(
+                "enabled", defaults.steering_enabled
+            )
+        ),
+        steering_priority_channels=tuple(
+            manifest.get("steering", {}).get(
+                "priority_channels", defaults.steering_priority_channels
+            )
+            or ()
+        ),
+        steering_max_turns=int(
+            manifest.get("steering", {}).get(
+                "max_turns", defaults.steering_max_turns
+            )
+        ),
+        steering_timeout_s=float(
+            manifest.get("steering", {}).get(
+                "timeout_s", defaults.steering_timeout_s
+            )
+        ),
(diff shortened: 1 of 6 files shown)


huangzesen:
Thanks for the PR! Independent senior review found no P0 but five P1 merge blockers in the preemptive parallel steering lane. Could you address these before we can merge?

1. **Duplicate lanes per heartbeat**: the same unchanged notification starts a new LLM lane on every heartbeat (no single-flight / notification fingerprint).
2. **No working production reply route**: Telegram is called with `chat_id=None`; the other four channels only write a local file — so the lane's reply cannot reach a default production channel.
3. **Channel name alone ≠ high-priority human message**: the lane treats channel presence as proof of priority, ignoring the producer's explicit `priority` / `has_human_messages` state.
4. **Preemption cannot stop a running tool**, while unscoped interrupt files can later abort unrelated turns.
5. **Kernel isolation violation**: the change introduces a kernel→wrapper dependency and fails the repository's kernel-isolation test.

The focused fake tests pass but do not exercise heartbeat integration, a real producer payload, a real channel send, a blocking tool, or the kernel layering gate. Please fix and add integration coverage for those paths.

huangzesen:
Superseded by #1301 (synthetic squash of #1211 + #1213 + #1212 includes the steering lane; all changed files covered). Closing as superseded.

PR #1198, opened 2026-08-06
fix(file_io): recursive glob includes top-level files

huangzesen:
## Summary

`file.glob(pattern="**/*")` returned zero matches for a directory whose only match was a top-level file, while `pattern="*"` found it correctly. Nested files matched; the zero-directory-depth case did not.

`LocalFileIOBackend.glob` uses `fnmatch` against a root-relative path. A leading `**/` in `fnmatch` requires a literal slash, so direct children cannot match. This conflicts with the documented recursive-search contract and standard recursive-glob semantics, where `**/` means zero or more directory levels.

This patch keeps the existing match first, then tests only slash-free direct-child paths against the pattern with one leading `**/` stripped. Traversal budgets, exclusions, sandbox resolution, sorting, and non-recursive matching are unchanged.

The regression covers both `**/*` and `**/*.md`, top-level and nested matches, extension filtering, and sorted output.

## Validation

- New regression fails on unmodified `origin/main` and passes with the patch.
- `python3 -m pytest tests/test_services_file_io.py -q` — 29 passed.
- Adjacent file capability suites — 135 passed.
- Parent rerun of the focused regression — 1 passed.
- `git diff --check` — clean.

`tests/test_architecture_documents.py` has one unrelated, pre-existing failure on the same base for an anatomy reference to the absent `daemon_supervisor/supervisor.py`; it is unchanged by this two-file patch.

### DIFF
--- a/src/lingtai/services/file_io.py
+++ b/src/lingtai/services/file_io.py
@@ -327,6 +327,15 @@ class LocalFileIOBackend(FileIOBackend):
         self.last_traversal = TraversalStats()
         search_root = Path(root) if root else (self._root or Path("."))
         results: list[str] = []
+        # fnmatch's translation of a leading "**/" requires a literal "/"
+        # in the matched string, so it never matches a zero-depth relative
+        # path (a direct child of search_root). Standard recursive-glob
+        # semantics (stdlib glob.glob(recursive=True), pathlib.Path.glob)
+        # treat "**/" as zero-or-more directory levels, so also try the
+        # pattern with that prefix stripped.
+        zero_depth_pattern = (
+            pattern[3:] if pattern.startswith("**/") else None
+        )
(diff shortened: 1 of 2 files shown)


huangzesen:
Dev3 portability follow-up: the new regression used POSIX-only slash suffix assertions. Head 76d4d04 now normalizes each returned path with Path.relative_to(tmp_dir).as_posix before comparing relative sets; the implementation is unchanged. Validation: full tests/test_services_file_io.py 29 passed, PureWindowsPath control passed, and git diff --check is clean. GitHub check-runs remain empty because existing workflow path filters do not cover these file-IO paths, so the local 29/135 evidence is not being represented as producer-canonical CI. No merge, release, or deploy action was taken.

PR #1206, opened 2026-08-06
fix(imap): preserve requested account in responses

huangzesen:
## Summary

- Preserve the account resolved by `IMAPMailManager.handle()` in the outer response envelope instead of always reporting the service default account.
- Add synthetic two-account regressions covering `check`, `read`, and `flag`, including account-scoped compound `email_id` routing and returned IDs.
- Document the response-account invariant in the IMAP manual.

## Validation

- `git diff --check` — PASS.
- `python -m compileall -q src/lingtai/mcp_servers/imap tests/test_imap_account_metadata.py` — PASS under the pinned project runtime.
- Direct focused harness under the pinned project runtime and `PYTHONPATH="$PWD/src:$PWD/tests"` — PASS, 24 tests across `test_imap_account_metadata.py`, `test_imap_empty_args.py`, and `test_imap_toolfamily_ltpv2.py`.
- Exact-base regression proof — PASS: the new check assertion fails against base `878220f…` because the old implementation reports the default account.
- Standard pytest collection was not available in the pinned project runtime (`No module named pytest`); no dependency or environment mutation was made. The available system pytest has incompatible installed MCP/MSAL dependencies and was not used as a product pass.

### DIFF
--- a/src/lingtai/mcp_servers/imap/SKILL.md
+++ b/src/lingtai/mcp_servers/imap/SKILL.md
@@ -50,7 +50,9 @@ a list of ids.
 
 - `email_id` is a compound key: `account:folder:uid` (e.g.
   `me@example.com:INBOX:1234`). Use the ids returned by `check`/`search`; do not
-  construct them by hand.
+  construct them by hand. Every action response includes `account` set to the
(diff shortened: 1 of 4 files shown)


huangzesen:

PR #1211, opened 2026-08-07
Add a preemptive steering lane for human guidance

rawpaper123:
## What changed

- Added a bounded parallel steering lane for high-priority human messages received during an active tool call.
- Injected recent conversation context, returned the reply on the source channel, and honored explicit interrupt decisions at the next safe tool boundary.
- Added per-agent steering configuration and lifecycle coverage.

## Why

Human messages arriving during a long tool call were only deferred, so follow-up guidance could not influence the active task until the turn ended.

## Impact

Agents can read and respond to live guidance without corrupting the main tool result, while interruption remains explicit and bounded.

## Validation

- `42 passed` across steering-lane and production call-site regression tests.

### DIFF
--- a/src/lingtai/agent.py
+++ b/src/lingtai/agent.py
@@ -119,6 +119,36 @@ def build_agent_config(manifest: dict[str, Any], *, max_rpm: int) -> AgentConfig
         max_aed_attempts=manifest.get(
             "max_aed_attempts", defaults.max_aed_attempts
         ),
+        # Steering block: per-agent steering.enabled (default true for human
+        # channel) plus optional priority channel / budget overrides. The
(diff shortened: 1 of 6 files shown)


huangzesen:
Closing per maintainer decision: this steering lane is rejected because in LingTai every message is already steering by nature — the agent reacts to each inbound message as guidance; a separate bounded parallel LLM lane adds an extra model call, an interrupt file contract, and a second reasoning stream without net benefit. The existing design (all messages steer the agent directly) covers the motivating scenario. Thanks for the work and the clear write-up.

huangzesen:
Closed as rejected (see prior comment).

PR #1212, opened 2026-08-07
Make Telegram read receipts Agent-open durable

rawpaper123:
## What changed

- Moved the Telegram 👀 reaction from ingress time to an Agent-open event.
- Persisted receipt intent and event cursor atomically in SQLite WAL, with retry/backoff for transient Telegram failures.
- Advanced the Telegram update offset only after inbox delivery succeeds.
- Added desired/applied Task Card revisions so the newest failed update resumes after restart.

## Why

The previous receipt meant only “transport received it,” could disappear on transient API failure, and could commit an update before the agent inbox accepted it.

## Impact

👀 now represents an actual Agent-open boundary, survives restarts, and does not depend on Task Card rendering.

## Validation

- `78 passed` across durable receipts, update-offset handling, Task Card revisions, and event-tail lifecycle tests.

### DIFF
--- a/src/lingtai/mcp_servers/telegram/account.py
+++ b/src/lingtai/mcp_servers/telegram/account.py
@@ -355,9 +355,6 @@ class TelegramAccount:
     def _process_update(self, update: dict) -> None:
         """Process a single update — filter, dispatch, track offset."""
         update_id = update.get("update_id", 0)
-        if update_id > self._last_update_id:
-            self._last_update_id = update_id
-            self._save_state()
 
         if "message" in update:
             # A plain message is a new bottom message in the chat; record it so a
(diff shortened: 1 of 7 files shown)


huangzesen:
Closing as superseded by PR #1305 (merged 2026-08-09, commit 44c0333c).

#1305 implements the same goal — making Telegram message progress visible to the human — with a much lighter design: three-boundary reactions (👌 received at ingress, 👀 seen after inbox delivery, ✍️ replied on targeted reply) instead of the SQLite receipt state machine. No durable store, no background worker, no error-classification layer needed.

The one valuable insight from this PR — update offset should only advance after successful inbox delivery (account.py +12) — remains a candidate for a separate minimal PR if we ever observe dropped-message evidence. Thanks for the thoughtful design and the motivation for this direction.

Issue #1197, opened 2026-08-05
Add provider-aware live reasoning-effort control for running agents


TZZheng:
## What's wrong

A running LingTai agent cannot currently change its reasoning effort through a supported live control surface. Effort is resolved from the preset/manifest baseline when the chat session is created, so changing it generally means editing configuration and refreshing or restarting the agent.

That is not only a missing feature. The current provider integrations also disagree about what an effort value means, and several paths accept, coerce, omit, or ignore the setting without reporting the effective behavior. A live controller must not expose those existing inconsistencies as if one universal effort enum worked everywhere.

The feature should let an agent—and later a local operator/TUI—change the effort used by the **next logical provider dispatch that has not yet captured its request policy**, while preserving the old policy for already captured, queued, streaming, or retried work.

## Where

- **Primary owner:** `lingtai-kernel`
  - resolved manifest/preset baseline
  - provider capability description and validation
  - chat-session/request construction
  - per-dispatch policy capture
  - agent-local runtime state and lifecycle reconciliation
  - self control and local operator protocol
- **Dependent surface:** `Lingtai-AI/lingtai`
  - preset editor truthfulness
  - `/effort` interaction and status rendering
- **Optional later scopes:** native daemons, external CLI daemons, authenticated remote/channel control

Source audit baselines:

- `lingtai-kernel` commit `3b5b37381418f8975c80d83e3cfd2f98c155d52c`
- `lingtai` commit `9e97c272161d61344306722637a81f42aae2b996`

Relevant kernel areas include manifest/preset validation, the Responses and Chat adapters, `ClaudeCodeAdapter.create_chat()`, `KimiCodeAdapter.create_chat()`, provider request serialization, chat-session rebuild/resume, daemon capsules, lifecycle restore, and `llm_call`/status projection.

## Reproduction

The current behavior is source-verifiable:

1. Resolve a preset with an effort/thinking value and start an agent.
2. The resolved value is captured into the created chat/session path; there is no supported agent-local live `get/set/clear` controller for later dispatches.
3. Provider paths do not share one honest contract:
   - The current LingTai Codex/Responses path has the local vocabulary `none | minimal | low | medium | high | xhigh`.
   - Claude Code 2.1.220 exposes `low | medium | high | xhigh | max`, but `ClaudeCodeAdapter.create_chat(..., thinking=...)` accepts the argument without forwarding it into the first or resumed CLI command.
[...]

PR #1220, opened 2026-08-07
feat(llm): add Codex reasoning construction contract

TZZheng:
# Add Codex reasoning construction contract (provider-local, immutable)

Related: #1197

## What

Gives the native main-agent `codex` / `codex-pool` / `codex_pool` route one owner for its reasoning-effort construction responsibilities:

- a provider-neutral extension point in `src/lingtai/kernel/llm/reasoning.py`: immutable route key/context (`ReasoningRouteKey`, `ReasoningRouteContext`), the `ReasoningContract` protocol, the immutable `ReasoningConstructionResult`, and an opaque mapping-application patch protocol — **no registry**;
- a provider-local Codex contract (`src/lingtai/llm/openai/codex_reasoning.py`) with evidence-backed exact model descriptors (capability source id per row) and the pre-dispatch rejection for explicit falsey values;
- `_register._codex` constructs one stateless contract and binds it for all three registered provider spellings, injecting it and the actual alias into `CodexOpenAIAdapter`;
- `CodexResponsesSession` captures one immutable `ReasoningConstructionResult` at construction and applies its provider-local patch once at the real request-mapping seam, before REST/WS diverge;
- exact `llm_call` observation with no-result field preservation.

Direct adapter construction (no contract) keeps the legacy path byte-identical. Unregistered routes (generic/custom Responses, Chat, other providers) are untouched. No daemon/TUI/live/global-catalogue machinery is added.

## Behavior semantics

The contract normalizes omitted/default to LingTai's quality baseline `xhigh` for the Codex route only; protocol-valid OMIT is not used as a LingTai outcome. Explicit falsey values (`""`, `None`, `False`, `0`) on a registered Codex route are rejected before any request is sent. Other providers remain independent.

## Tests

`tests/test_codex_reasoning_contract.py` proves route/model capability per descriptor row, omitted/default capture on both the registered factory route and the kernel `SessionManager` ensure/rebuild route, immutable result/patch ownership, per-alias contract binding, REST/WS parity, pre-dispatch rejection, legacy/direct fallthrough, and construction parity. `tests/test_session.py` proves exact safe `llm_call` observation. `tests/test_codex_native_multiaccount.py` / `test_openai_responses_streaming.py` preserve generic and direct legacy behavior.

## Validation

Test-first with authentic pre-implementation RED; parent full-gate validation; independent literal Opus 5 exact-head review PASS (after one BLOCK on an anatomy-navigation orphan — fixed in the amended head). Gates: focused 58 · falsey 12 · descriptor 7 · route 6-file 386 · session 50 · Responses trio 62 · docs governance byte-identical to base (3 inherited failures / 71 pass) · compileall · diff-check.


huangzesen:
Superseded by #1314 (merged aec17593) which implements the simplified model: OpenAI Responses as canonical semantics, v1 projection, Claude Code dedicated handling. Closing.

PR #1222, opened 2026-08-07
fix(daemon): expose email tool to daemon tasks

huangzesen:
## Summary

- Daemon tasks requesting `tools:['email']` failed at **detached-run time** with `ValueError: Unknown tools for emanation: {'email'}`, even though the parent's own pre-flight validation accepted the request (that check runs against a real, fully-wired `Agent`; the actual run happens in a separate detached supervisor process against a bare `DaemonSupervisorAgentStub` that never wires any intrinsic).
- `DaemonSupervisorAgentStub` deliberately never constructs a full `Agent` (to avoid contending for the parent's workdir lease/heartbeat), so it never ran `BaseAgent._wire_intrinsics()`. It reconstructs `shell`/`file` via `setup_capability`, but nothing wired the one daemon-eligible intrinsic exception (`email`) that `DaemonManager._daemon_intrinsic_surface()` looks for — explaining why `shell`/`file` resolved fine while only `email` was reported "unknown".
- Fix: hoist the existing `allowed = {"email"}` set in `_daemon_intrinsic_surface` into a shared `DAEMON_ALLOWED_INTRINSICS` constant, and wire each entry (module + handler + `boot()`) onto the detached stub the same way `_wire_intrinsics` does for a real agent. No other intrinsic is wired — security posture (identity/lifecycle/recursive-mutation tools stay unavailable; `tools=[]` result-only daemons still get no email) is unchanged.

## Test plan

- [x] New repro confirms the exact reported exception (`ValueError: Unknown tools for emanation: {'email'}`) against `DetachedDaemonExecutionHost._build_lingtai_surface()` on unpatched `origin/main`.
- [x] Added `test_detached_email_manifest_wires_schema_and_dispatch` — schema *and* dispatch both work end-to-end (`email(action='check', ...)` actually returns a result, not just a schema).
- [x] Added `test_detached_preset_manifest_email_requires_explicit_tool` — same contract on the preset branch.
- [x] Added `test_detached_result_only_manifest_excludes_email` — `tools=[]` still excludes email (security-posture regression guard).
- [x] All three fail on unpatched code with the exact reported message; pass after the fix.
- [x] `pytest tests/test_daemon.py tests/test_daemon_detached_supervisor.py tests/test_daemon_preset_capabilities.py tests/test_tool_family_daemon_migration.py tests/ -k "email or daemon"` → 1009 passed, 7 skipped, 3 failed — same 3 failures present on unmodified `origin/main` (unrelated: `task_card` leaking into `_parent_host_tool_floor()`, and two CLI-backend resume/follow-up assertions), no new failures.

Full investigation notes: see the accompanying report generated during this fix.

### DIFF
--- a/src/lingtai/tools/daemon/__init__.py
+++ b/src/lingtai/tools/daemon/__init__.py
@@ -438,6 +438,16 @@ EMANATION_BLACKLIST = {
     "knowledge",
 }
 
+# The daemon-eligible subset of the mandatory intrinsic layer — see
+# ``DaemonManager._daemon_intrinsic_surface``. Email is the sole deliberate
+# exception so a parent can grant a running daemon local-network
+# communication; every other intrinsic (identity/lifecycle/recursive-mutation)
+# stays unavailable. Module-level so both the in-process manager and the
+# detached supervisor's tool-surface composition (``execution_host.py``,
+# which cannot construct a full ``Agent``) wire the exact same set instead of
+# risking drift between two hand-maintained copies.
+DAEMON_ALLOWED_INTRINSICS = frozenset({"email"})
+
 
 def _parent_host_tool_floor() -> frozenset[str]:
     """The always-on host tools a preset emanation may borrow from the parent.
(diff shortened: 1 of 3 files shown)


huangzesen:
Superseded by #1224 (email as daemon MCP, LTP v2 — merged c46fad8e). Closing.

PR #1234, opened 2026-08-08
feat(glm): provider-aware configured reasoning-effort contract

TZZheng:
## Provider-aware configured reasoning effort for GLM Chat Completions

Related: #1197

Implements the GLM vertical slice of the provider-aware reasoning-effort plan for both LingTai aliases (`zhipu` and `glm`). Configured values now reach the real GLM Chat Completions wire through a provider-owned, fail-closed contract instead of the generic OpenAI-compatible high/else-low mapper.

## What changes

- **Provider-scoped vocabulary:** GLM accepts exactly `none | high | max`. The global `THINKING_LEVELS` tuple is unchanged, so other providers do not acquire GLM's `max` value.
- **Exact wire mapping:**
  - `none` → top-level `thinking: {type: "disabled"}` and no `reasoning_effort`;
  - `high | max` → top-level `thinking: {type: "enabled"}` plus the exact flat `reasoning_effort`;
  - omission, the internal `default` sentinel, and direct-adapter `None` → no GLM reasoning fields.
- **Deliberate omission behavior change:** untouched main currently converts an omitted GLM setting into explicit flat `reasoning_effort: "high"`. This PR emits no reasoning fields instead, relying on GLM-5.2's upstream default (thinking enabled / maximum effort). That may increase reasoning tokens and latency for omitted main-agent configuration. Operators who want the old tier can set `thinking: high`. Native GLM daemon omission already receives the provider default, so this also aligns the two omission paths without adding daemon configured-value propagation.
- **Fail closed before dispatch:** explicit values are supported only for the exact normalized model id `glm-5.2`; every other model id (including variants such as `glm-5.2-air`) is rejected before a request is sent. Case/whitespace aliases and non-string configured values are rejected rather than coerced.
- **Both aliases share one owner:** `zhipu` and `glm` use the same provider-owned frozen descriptor, validation, wire construction, and observability contract.
- **Wire ownership without generic drift:** the descriptor emits through a copied `extra_body`, which the OpenAI SDK hoists to the required top-level JSON fields. Generic OpenAI-compatible mapping remains byte-identical for other providers, and descriptor fields win over a conflicting flat duplicate without mutating caller input.
- **Safe observability:** every `llm_call` can carry exactly five bounded plain-string fields (`reasoning_requested`, `reasoning_normalized`, `reasoning_actual`, `reasoning_source`, `reasoning_capability_source`) frozen at session construction. Unknown keys, credentials, body/session data, `model`/`api_call_id` collisions, non-strings, and overlong strings are structurally discarded. Providers without the accessor retain their previous event shape.
- **Session semantics preserved by provider:** `SessionManager` keeps falsey/`None` values intact only for `zhipu`/`glm` so the GLM contract can distinguish omission from invalid explicit values; every other provider retains the prior `thinking or "high"` fallback, including Codex's `default` sentinel.
- **Soul remains honest:** no soul production code changes. An explicit unsupported GLM model still fails at the provider contract and degrades through soul's existing exception boundary rather than silently sending a fabricated effort.

## Validation

- Authentic schema/preset RED: **60 failed / 65 passed** while the old scope gate still rejected GLM.
- Authentic wire/module RED: **54 failed / 40 passed** against the old flat-low behavior and absent provider contract.
- Parent-blocker RED: **31 failed / 139 passed**, proving both session falsey rewriting and an actual `api_key` observability escape before the fixes; all 17 non-test blobs were unchanged at that point.
- Parent exact-head gates: **176** focused GLM tests, **465** schema/preset/merge aggregate, **39** soul tests, and **358** adjacent session/logging/agent/OpenAI/kernel tests — all passed.
- Real `ZhipuAdapter` + `httpx.MockTransport` probes verified omission/`None`, `none`, `high`, `max`, streaming parity, generic OpenAI preservation, `extra_body` behavior, and zero-request fail-closed rejection.
- Targeted docs/Anatomy gate: candidate and exact base both **100 passed / 5 failed** with identical inherited failure sets. Compileall, `git diff --check`, and exact-scope checks passed.
- Fresh independent literal Opus 5 (`high`, `plan`) exact-head review: **VERDICT PASS**, zero P0/P1. The reviewer independently reconstructed the wire probes and mutation-tested the two parent fixes.

## Scope and deferred boundaries

Exactly 18 files, one commit, `+1575/-34`. This slice does **not** add configured-value propagation to native daemons, TUI exposure, live control, persisted runtime effort, `clear_thinking`, other model variants, or another provider. Those remain later nodes of #1197.


huangzesen:
Superseded by #1314 (merged aec17593) which implements the simplified model: OpenAI Responses as canonical semantics, v1 projection, Claude Code dedicated handling. Closing.

PR #1247, opened 2026-08-09
fix(daemon): honor explicit per-batch timeout

huangzesen:
## Summary
- treat `daemon.emanate(timeout=...)` as the uncapped explicit override advertised by its schema, instead of silently clamping it to the 3600s default
- keep `null`/omitted timeout at the manager default and keep the separately declared `max_turns` ceiling unchanged
- clarify the tool description and Contract, and add a regression that asserts `10800` reaches both `daemon.json` and `supervisor_manifest.json`

## Root cause
The explicit timeout path reused `min(value, manager_default)` from the neighboring `max_turns` ceiling. Unlike `max_turns`, `timeout` has no schema `maximum`; 3600 is the default when omitted, not an advertised cap. The call returned `dispatched` while persisting 3600, so long workers were killed early.

## Validation
- `475 passed in 61s`: focused daemon, per-batch limit, manifest, detached-supervisor, run-dir, tool-schema, and CLI-backend suites
- decisive post-rebase regression: `12 passed` in `tests/test_daemon_per_batch_limits.py`
- import smoke resolves both changed modules from this worktree
- `git diff --check`

No merge, release, deploy, runtime refresh, or configuration change is part of this PR.


huangzesen:
Superseded by #1330. Closing as superseded.

huangzesen:
Superseded by #1330. Closing as superseded.

PR #1254, opened 2026-08-09
fix(claude-code): handle tools=None in ClaudeCodeChatSession constructor; fix contract mock signature (#1238)

### LINKED ISSUE #1238: bug: ClaudeCodeChatSession crashes on tools=None; 3 contract tests failing
## Summary

`ClaudeCodeChatSession.__init__` assigns `self._tools = tools` without normalizing, while `update_tools()` **in the same class** correctly does `list(tools) if tools else []`. Any caller constructing the session directly — rather than via `ClaudeCodeAdapter.create_chat()`, which pre-normalizes — passes `tools=None` straight through, and `_system_block_digest()` then raises `TypeError: 'NoneType' object is not iterable`.

Fixing that unmasks a **second, latent bug**: the test fixture's `_fake_invoke(prompt, model)` mock is signature-incompatible with the real `_invoke(self, prompt, model, *, resume_session_id=None, _correcting=False)`. `resume_session_id` was added later; this path was simply never reachable because the constructor always crashed first.

Scope note, to be accurate: this is currently **test-only impact** — the production path goes through `create_chat()`, which normalizes. But the constructor is undefensive at its own boundary while its sibling method is not, which is a latent trap for any future second call path.

## Reproduction steps

```bash
python3 -m venv .venv
LINGTAI_SKIP_RUST_BUILD=1 .venv/bin/python -m pip install -e . pytest
.venv/bin/python -m pytest tests/contracts/llm_conversation_input/
```

Fails:

- `test_send_str.py::test_send_str_reaches_transport_as_provider_text[claude_code]`
- `test_send_tool_results.py::test_canonical_tool_result_converts_to_provider_wire[claude_code]`
- `test_send_tool_results.py::test_no_regime_forwards_unconverted_dataclass`

## Expected behavior

The three contract tests pass; constructing a session with `tools=None` is handled the same way `update_tools(None)` already handles it.

## Actual behavior / logs

```
TypeError: 'NoneType' object is not iterable
```

and, once that is fixed, the previously-unreachable second failure:

```
TypeError: _fake_invoke() got an unexpected keyword argument 'resume_session_id'
```

## LingTai / Python / OS versions

lingtai-kernel (baseline commit `2d8613f2`), Python 3.11, macOS

## Proposed fix

```python
# src/lingtai/llm/claude_code/adapter.py — ClaudeCodeChatSession.__init__
tools: list[FunctionSchema] | None,      # widened hint
...
# Normalized the same way ``update_tools`` already normalizes later updates,
# so ``_system_block_digest`` can always iterate ``_tools``.
self._tools = list(tools) if tools else []
```

```python
# tests/contracts/llm_conversation_input/regimes.py
def _fake_invoke(prompt, model, **kwargs):   # stays compatible as _invoke evolves
```

<details>
<summary>Full diff</summary>

```diff
diff --git a/src/lingtai/llm/claude_code/adapter.py b/src/lingtai/llm/claude_code/adapter.py
index e27c7dbe..8ee82392 100644
--- a/src/lingtai/llm/claude_code/adapter.py
+++ b/src/lingtai/llm/claude_code/adapter.py
@@ -309,14 +309,16 @@ class ClaudeCodeChatSession(ChatSession):
         adapter: "ClaudeCodeAdapter",
         model: str,
         system_prompt: str,
-        tools: list[FunctionSchema],
+        tools: list[FunctionSchema] | None,
         interface: ChatInterface,
         context_window: int,
     ) -> None:
         self._adapter = adapter
         self._model = model
         self._system_prompt = system_prompt
-        self._tools = tools
+        # Normalized the same way ``update_tools`` already normalizes later
+        # updates, so ``_system_block_digest`` can always iterate ``_tools``.
+        self._tools = list(tools) if tools else []
         self._interface = interface
         self._context_window = context_window
         # Remote state is an acceleration only. The canonical interface remains
diff --git a/tests/contracts/llm_conversation_input/regimes.py b/tests/contracts/llm_conversation_input/regimes.py
index 97aa0204..d6f7a415 100644
--- a/tests/contracts/llm_conversation_input/regimes.py
+++ b/tests/contracts/llm_conversation_input/regimes.py
@@ -348,7 +348,7 @@ def _build_claude_code() -> tuple[Any, Any]:
     # ``adapter._invoke(prompt, model)``. Mock only that seam; capture the prompt.
     calls: list[str] = []
 
-    def _fake_invoke(prompt, model):
+    def _fake_invoke(prompt, model, **kwargs):
         calls.append(prompt)
         return (
             {"action": "final", "text": "ok"},
```

</details>

## Verification

Reproduced all 3 failures on a clean git worktree at the baseline commit in a fresh venv. After the fix, the 3 target tests plus the full `tests/contracts/` and `tests/test_claude_code_adapter.py` suites — **223 tests** — pass. `git diff --check` clean.

`tests/test_architecture_documents.py` has one pre-existing unrelated failure (a stale `related_files` pointer to a missing `daemon_supervisor/supervisor.py`), confirmed present on the unmodified baseline via `git stash` — it predates and is untouched by this change.

The patch was built against `2d8613f2` but still **applies cleanly to a much later HEAD** with only a line offset (verified with `git apply --check`), so the surrounding code has not meaningfully drifted.

## Suggested follow-ups

1. Worth scanning for other constructors that rely on an external entry point to sanitize their arguments while a sibling method sanitizes its own.
2. `unittest.mock.create_autospec` (or equivalent) on these mock seams would surface signature drift at the point the real signature changes, rather than whenever the path first becomes reachable.

## Submission hygiene

- [x] I removed secrets, tokens, private paths that are not needed, and unrelated logs.

huangzesen:
Fixes #1238

## Problem

`ClaudeCodeChatSession.__init__` stored the `tools` argument as-is (`self._tools = tools`). When called with `tools=None`, `_system_block_digest` iterated `self._tools` and crashed with `TypeError: 'NoneType' object is not iterable`, breaking the 3 contract tests in `tests/contracts/llm_conversation_input/`. `update_tools` already normalizes this way (`list(tools) if tools else []`), so the constructor was inconsistent.

Additionally, the contract test mock `_fake_invoke(prompt, model)` had a signature incompatible with the real `_invoke(self, prompt, model, *, resume_session_id=None)`, which `send()` calls with the `resume_session_id` keyword.

## Changes

- `src/lingtai/llm/claude_code/adapter.py`: widen the `tools` type hint to `list[FunctionSchema] | None` and normalize in the constructor: `self._tools = list(tools) if tools else []`.
- `tests/contracts/llm_conversation_input/regimes.py`: make the fake `_fake_invoke(prompt, model, **kwargs)` accept the extra keyword.

## Validation

- Pre-fix reproduction (main HEAD): the 3 target tests fail with `TypeError: 'NoneType' object is not iterable` at `adapter.py:332`.
- Post-fix: all 3 target tests pass.
- Full contract + adapter suite: `pytest tests/contracts/ tests/test_claude_code_adapter.py` → **223 passed** in 2.06s.
- `tests/test_architecture_documents.py::test_every_tracked_file_climbs_the_anatomy_graph` fails on main before this change too (pre-existing: untracked test files in the anatomy graph and a missing `src/lingtai/daemon_supervisor/supervisor.py`); unrelated to this PR.

### DIFF
--- a/src/lingtai/llm/claude_code/adapter.py
+++ b/src/lingtai/llm/claude_code/adapter.py
@@ -225,7 +225,7 @@ class ClaudeCodeChatSession(ChatSession):
         adapter: "ClaudeCodeAdapter",
         model: str,
         system_prompt: str,
-        tools: list[FunctionSchema],
+        tools: list[FunctionSchema] | None,
         interface: ChatInterface,
         context_window: int,
         effort_argv: list[str] | None = None,
@@ -233,7 +233,9 @@ class ClaudeCodeChatSession(ChatSession):
         self._adapter = adapter
         self._model = model
         self._system_prompt = system_prompt
-        self._tools = tools
+        # Normalized the same way ``update_tools`` already normalizes later
+        # updates, so ``_system_block_digest`` can always iterate ``_tools``.
+        self._tools = list(tools) if tools else []
         self._interface = interface
         self._context_window = context_window
         # Frozen at construction, never re-read from mutable adapter state.
--- a/tests/contracts/llm_conversation_input/regimes.py
+++ b/tests/contracts/llm_conversation_input/regimes.py
@@ -359,7 +359,7 @@ def _build_claude_code() -> tuple[Any, Any]:
     # ``adapter._invoke(prompt, model)``. Mock only that seam; capture the prompt.
     calls: list[str] = []
 
-    def _fake_invoke(prompt, model):
+    def _fake_invoke(prompt, model, **kwargs):
         calls.append(prompt)
         return (
             {"action": "final", "text": "ok"},


huangzesen:

PR #1255, opened 2026-08-09
fix(kernel): agent session started_at is ISO-8601 in all rebuild tiers

### LINKED ISSUE #750: AgentSession.started_at format depends on rebuild tier: ISO-8601 from sqlite, raw epoch string from reverse/full scan
## Summary

`AgentSession.started_at` comes back in two different formats depending on which rebuild tier produced it: the sqlite tier returns an ISO-8601 UTC string (e.g. `"2026-07-05T08:41:12Z"`), while both fallback scan tiers return the raw epoch float stringified (e.g. `"1751704872.391"`). Any consumer of `started_at` therefore sees a format that depends on incidental runtime conditions (whether the `log.sqlite` sidecar was available), not on any contract. Both scan tiers should route through the existing `_iso_from_ts` helper, and a test should pin the format.

## Narrative

`src/lingtai_kernel/agent_session.py` rebuilds the current "agent mind segment" (the `AgentSession` dataclass, keyed by `molt_count`) from the durable trajectory at startup/refresh. The rebuild has three tiers, in preference order (see `rebuild_agent_session_from_events`): **Tier 1** queries the `logs/log.sqlite` sidecar; **Tier 2** does a bounded tail-first scan of `events.jsonl`; **Tier 3** forward-scans the whole `events.jsonl` as a last resort. All three tiers find the most recent `psyche_molt` boundary event and use its `ts` field as the session start.

That `ts` field is a wall-clock epoch float — events are logged in `src/lingtai_kernel/base_agent/__init__.py` (`_log`, around line 890) with `"ts": time.time()`. Tier 1 correctly converts it to a human-readable timestamp:

```python
# src/lingtai_kernel/agent_session.py:268 (_rebuild_via_sqlite)
started_at=_iso_from_ts(boundary_ts) if boundary_ts else None,
```

`_iso_from_ts` (line 445) produces `datetime.fromtimestamp(ts, tz=timezone.utc).isoformat().replace("+00:00", "Z")` — the same `...Z` shape used by `RuntimeSession.started_at` (via `_utc_now_iso`, line 60) and by manifest timestamps elsewhere in the kernel (e.g. `base_agent/lifecycle.py` writes `strftime("%Y-%m-%dT%H:%M:%SZ")`).

But the two fallback tiers stringify the raw float instead:

```python
# src/lingtai_kernel/agent_session.py:337 (_rebuild_via_reverse_scan)
started_at=str(boundary_ts) if boundary_ts is not None else None,
```

```python
# src/lingtai_kernel/agent_session.py:392 (_rebuild_via_full_scan)
started_at=str(boundary_ts) if boundary_ts is not None else None,
```

So the same agent, with the same `events.jsonl`, gets `started_at="2026-07-05T08:41:12.391000Z"` when the sqlite sidecar is healthy — and `started_at="1751704872.391"` when it is not (sidecar missing, sqlite disabled after an error, or a fresh checkout where the sidecar has not been built yet). The tier choice is deliberately invisible to consumers except via the diagnostic `rebuild_tier` field; `started_at` leaking the tier through its *format* defeats that design. A downstream consumer that renders the session start (status surfaces, dashboards, or anything reading the installed `AgentSession` off `Session.agent_session` in `src/lingtai_kernel/session.py`) either shows a raw epoch float to a human, or has to defensively parse two formats.

How we likely got here: Tier 1 was written against sqlite rows where `ts` is already coerced to a float (`_coerce_ts` in `services/logging.py`), so the author reached for `_iso_from_ts`. The scan tiers read raw JSONL where `ev.get("ts")` is `Any`, and `str(...)` was the low-friction way to satisfy the `str | None` field type without worrying about non-float values. That shortcut froze the epoch-float format into the fallback paths.

Better looks like: one format, everywhere — ISO-8601 UTC with the `Z` suffix, matching `RuntimeSession.started_at` — with the scan tiers coercing the raw JSONL `ts` to float before conversion, and returning `None` (as they already do) when the value is absent or unusable. A regression test should pin the format for all three tiers so a future tier addition cannot reintroduce the drift; today `tests/test_agent_session.py` never asserts anything about `AgentSession.started_at` (only `RuntimeSession.started_at` truthiness, line 42), so this path is entirely uncovered.

## Current behavior

- `_rebuild_via_sqlite` (`src/lingtai_kernel/agent_session.py:229-278`): sets `started_at=_iso_from_ts(boundary_ts) if boundary_ts else None` (line 268) → ISO-8601 `...Z` string, or `None` when there is no boundary (`boundary_ts == 0.0`).
- `_rebuild_via_reverse_scan` (lines 281-347): sets `started_at=str(boundary_ts) if boundary_ts is not None else None` (line 337), where `boundary_ts = ev.get("ts")` (line 318) is the raw JSONL value — an epoch float in practice → `"1751704872.391"`-style string.
- `_rebuild_via_full_scan` (lines 350-402): identical `str(boundary_ts)` at line 392, `boundary_ts = ev.get("ts")` at line 379.
- `_iso_from_ts` (lines 445-453) already exists in the same module and handles falsy input (`None`) and conversion failures (`OverflowError, OSError, ValueError` → `None`), but does **not** catch `TypeError` — relevant if a scan tier hands it a non-numeric raw JSONL value.
- Subtle secondary inconsistency: the sqlite tier maps a zero/absent boundary to `None` via `if boundary_ts`, while the scan tiers use `is not None` — a literal `ts: 0` boundary event would yield `None` in Tier 1 but `"0"` in Tiers 2/3.
- `tests/test_agent_session.py` has no assertion on `AgentSession.started_at` in any tier.

All changes are confined to `src/lingtai_kernel/agent_session.py` plus tests.

1. **Harden `_iso_from_ts` to accept raw JSONL values.** Change its signature to `def _iso_from_ts(ts: Any) -> str | None` and coerce defensively, so scan tiers can pass `ev.get("ts")` directly:

   ```python
   def _iso_from_ts(ts: Any) -> str | None:
       try:
           ts = float(ts or 0.0)
       except (TypeError, ValueError):
           return None
       if not ts:
           return None
       try:
           return datetime.fromtimestamp(ts, tz=timezone.utc).isoformat().replace(
               "+00:00", "Z"
           )
       except (OverflowError, OSError, ValueError):
           return None
   ```

   This keeps the existing behavior for Tier 1 (float in, `None` for `0.0`) and makes `None`, missing, string-numeric, and garbage `ts` values all safe.

2. **Route Tier 2 through the helper.** In `_rebuild_via_reverse_scan` (line 337) replace:

   ```python
   started_at=str(boundary_ts) if boundary_ts is not None else None,
   ```

   with:

   ```python
   started_at=_iso_from_ts(boundary_ts),
   ```

3. **Route Tier 3 through the helper.** Same one-line replacement in `_rebuild_via_full_scan` (line 392).

4. **Simplify Tier 1 (optional, for symmetry).** Line 268 can become `started_at=_iso_from_ts(boundary_ts)` since the hardened helper already returns `None` for falsy input; the `if boundary_ts else None` guard becomes redundant. This also resolves the `if boundary_ts` vs `is not None` asymmetry noted above in one direction (zero/absent boundary → `None` in all tiers).

5. **Back-compat check.** `AgentSession.started_at` is typed `str | None` and the field is rebuilt fresh at every startup/refresh (`rebuild_agent_session` → `install_agent_session`); it is not persisted anywhere in the epoch-string form, so there is no stored-data migration. Grep for consumers that might have adapted to the epoch-string format (`.started_at` on `AgentSession`) — as of this writing `Session.agent_session_token_usage` in `src/lingtai_kernel/session.py` does not include `started_at`, so the blast radius is minimal.

6. **Add tests** (see Testing plan) pinning the format across tiers.

## Risks and alternatives

- **Risk: a consumer silently depends on the epoch string.** Because the value only ever existed transiently in-process on the fallback tiers, and the primary (sqlite) tier already emits ISO, any consumer robust enough to work on Tier 1 already handles ISO. The change makes the fallback tiers match the primary tier, which is the least surprising direction.
- **Risk: `ts` values in old `events.jsonl` files that are not floats.** Handled by the hardened `_iso_from_ts` coercion (step 1), which degrades to `None` rather than raising — strictly better than today's `TypeError`-prone situation had someone naively swapped in the current helper.
- **Alternative: standardize on epoch floats instead of ISO.** Rejected: `RuntimeSession.started_at`, `_utc_now_iso`, the agent manifest (`lifecycle.py`), and `_iso_from_ts` all already commit to ISO-8601 `Z` strings; epoch strings are hostile to human-facing surfaces and would require changing the already-correct primary tier.
- **Alternative: change the field type to `float | None` epoch and format at render time.** More invasive (touches every consumer and the dataclass contract) for no benefit; the field is documented as a timestamp string.
- **Alternative: leave Tier 1 untouched (skip step 4).** Fine — steps 1-3 alone fix the bug; step 4 is a cleanup that unifies the zero-boundary edge case.

## Testing plan

Add to `tests/test_agent_session.py` (which already builds JSONL fixtures for tier tests):

1. `test_reverse_scan_started_at_is_iso8601` — write an `events.jsonl` containing a `psyche_molt` event with a known epoch `ts` (e.g. `1751704872.0`) followed by `llm_response` events; run `_rebuild_via_reverse_scan` (or the public `rebuild_agent_session_from_events` forced onto Tier 2 by omitting sqlite); assert `session.started_at == "2025-07-05T09:21:12Z"`-style exact ISO string (compute expected via `datetime.fromtimestamp(ts, tz=timezone.utc)`), and assert it ends with `"Z"` and does not equal `str(ts)`.
2. `test_full_scan_started_at_is_iso8601` — same fixture, force Tier 3 (small `max_scan_bytes` so the reverse scan defers, or call `_rebuild_via_full_scan` directly); same assertions.
3. `test_started_at_format_consistent_across_tiers` — build one fixture, run all three tiers (sqlite via an injected `sqlite_query_fn` returning a boundary row with the same `ts`), assert all three `started_at` values are identical.
4. `test_started_at_none_when_no_boundary` — fixture with only `llm_response` events (never molted); assert `started_at is None` on both scan tiers.
5. `test_iso_from_ts_handles_bad_input` — direct unit test: `_iso_from_ts(None) is None`, `_iso_from_ts(0) is None`, `_iso_from_ts("garbage") is None`, `_iso_from_ts("1751704872.0")` returns the ISO string (string-numeric coercion).

huangzesen:
Fixes #750

## Problem

`AgentSession.started_at` was formatted differently depending on which rebuild tier produced it: the sqlite tier (Tier 1) emitted ISO-8601 UTC (`2026-07-05T08:41:12Z`-style via `_iso_from_ts`), while the two fallback JSONL scan tiers (Tier 2 reverse scan, Tier 3 full scan) stringified the raw epoch float (`"1751704872.391"`). Consumers therefore saw a format that depended on incidental runtime conditions (whether the `log.sqlite` sidecar existed), not on any contract.

## Fix (all in `src/lingtai/kernel/agent_session.py`)

1. Hardened `_iso_from_ts` to accept raw JSONL `ts` values: signature is now `def _iso_from_ts(ts: Any) -> str | None`, coercing via `float(ts or 0.0)` and returning `None` on `TypeError`/`ValueError` (missing, `None`, garbage, or non-numeric values) instead of raising.
2. Tier 2 (`_rebuild_via_reverse_scan`) now sets `started_at=_iso_from_ts(boundary_ts)` instead of `str(boundary_ts)`.
3. Tier 3 (`_rebuild_via_full_scan`) now sets `started_at=_iso_from_ts(boundary_ts)` instead of `str(boundary_ts)`.
4. Tier 1 (`_rebuild_via_sqlite`) simplified to `started_at=_iso_from_ts(boundary_ts)`; the redundant `if boundary_ts else None` guard is gone, which also unifies the zero/absent-boundary edge case (all tiers map it to `None`).

No persisted data migration is needed: the field is rebuilt fresh at every startup/refresh, and no consumer reads `AgentSession.started_at` outside this module (verified by grep: `Session.agent_session_token_usage` does not surface it).

## Tests

Added 5 tests to `tests/test_agent_session.py`:

- `test_reverse_scan_started_at_is_iso8601`
- `test_full_scan_started_at_is_iso8601`
- `test_started_at_format_consistent_across_tiers` (sqlite + both scan tiers produce the identical ISO string for the same boundary `ts`)
- `test_started_at_none_when_no_boundary`
- `test_iso_from_ts_handles_bad_input` (`None`, `0`, `0.0`, `"garbage"`, `[]` -> `None`; `"1751704872.0"` -> ISO string)

Results:

```
$ python3 -m pytest tests/test_agent_session.py -q
22 passed in 0.14s

$ python3 -m pytest tests/test_agent_session.py tests/test_agent_session_wiring.py tests/test_session.py tests/test_services_logging.py -q
116 passed, 1 warning in 1.94s
```

(Pre-existing, unrelated collection errors in wechat/whatsapp test files come from an incompatible `mcp` package in the local site-packages and are untouched by this change.)

### DIFF
--- a/src/lingtai/kernel/agent_session.py
+++ b/src/lingtai/kernel/agent_session.py
@@ -293,7 +293,7 @@ def _rebuild_via_sqlite(
     row = agg[0] if agg else {}
     return AgentSession(
         molt_count=molt_count,
-        started_at=_iso_from_ts(boundary_ts) if boundary_ts else None,
+        started_at=_iso_from_ts(boundary_ts),
         boundary_source=boundary_source,
         boundary_offset=boundary_offset,
         api_calls=int(row.get("n") or 0),
@@ -362,7 +362,7 @@ def _rebuild_via_reverse_scan(
 
     return AgentSession(
         molt_count=molt_count,
-        started_at=str(boundary_ts) if boundary_ts is not None else None,
+        started_at=_iso_from_ts(boundary_ts),
         boundary_source=boundary_source,
         boundary_offset=None,
         api_calls=api_calls,
@@ -417,7 +417,7 @@ def _rebuild_via_full_scan(events_file: Path, molt_count: int) -> AgentSession:
 
     return AgentSession(
         molt_count=molt_count,
-        started_at=str(boundary_ts) if boundary_ts is not None else None,
+        started_at=_iso_from_ts(boundary_ts),
         boundary_source=boundary_source,
         boundary_offset=None,
         api_calls=api_calls,
@@ -458,7 +458,16 @@ def _read_tail_lines(path: Path, max_bytes: int) -> tuple[list[str], bool]:
     return lines, hit_start
 
 
-def _iso_from_ts(ts: float) -> str | None:
+def _iso_from_ts(ts: Any) -> str | None:
+    """Convert a wall-clock epoch (float, int, or numeric string) to ISO-8601 UTC.
+
+    Returns ``None`` for falsy, missing, or unusable input so callers can pass
+    raw JSONL ``ts`` values (``Any``) straight through without pre-checks.
+    """
+    try:
+        ts = float(ts or 0.0)
+    except (TypeError, ValueError):
+        return None
     if not ts:
         return None
     try:
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -18,6 +18,7 @@ from lingtai.kernel.agent_session import (
     TOKEN_EVENT,
     AgentSession,
     RuntimeSession,
+    _iso_from_ts,
     _rebuild_via_full_scan,
     _rebuild_via_reverse_scan,
     _rebuild_via_sqlite,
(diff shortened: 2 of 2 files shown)


huangzesen:

PR #1256, opened 2026-08-09
fix: never probe localhost for schemeless preset base_url (#741)

### LINKED ISSUE #741: check_connectivity silently probes localhost when a preset's base_url lacks a URL scheme
## Summary

`check_connectivity()` in `src/lingtai_kernel/preset_connectivity.py` parses the preset's `base_url` with `urlparse()` and passes `parsed.hostname` directly to `socket.create_connection()`. For a schemeless `base_url` such as `api.openai.com` or `myhost:8080` — an easy mistake in a hand-authored preset manifest — `parsed.hostname` is `None`, and `create_connection((None, 80))` resolves to **localhost**. The preset then reports a false `"ok"` if anything local listens on the port, or `"unreachable"` with a misleading connection-refused error that points at the wrong host.

## Narrative

The preset connectivity check exists precisely so an agent can trust the `presets` listing before swapping LLMs. The module docstring is explicit about this ("Caching connectivity status would let an agent confidently swap into a preset that went down"). The design invests heavily in freshness — no caching, parallel probes, a two-tier credential-then-network check — but the network tier trusts `urlparse()` more than `urlparse()` deserves.

A user hand-authors a preset manifest with an `llm.base_url` of `my-vllm-box:8000` or `api.openai.com` — both perfectly natural things to type, since many tools (curl, browsers, docker) accept schemeless host specs. The manifest loads fine: `src/lingtai_kernel/intrinsics/system/preset.py:294` forwards it verbatim into the connectivity spec:

```python
connectivity_specs.append({
    "provider": llm.get("provider"),
    "base_url": llm.get("base_url"),
    "api_key_env": llm.get("api_key_env"),
})
```

Then `check_connectivity()` does (`src/lingtai_kernel/preset_connectivity.py:146-151`):

```python
parsed = urlparse(url)
host = parsed.hostname
port = parsed.port or (443 if parsed.scheme == "https" else 80)

try:
    latency_ms = _probe_host(host, port, _PROBE_TIMEOUT_S)
```

Python's `urlparse` behaves in two surprising ways here, both verified on the current interpreter:

- `urlparse("api.openai.com")` → `scheme=''`, `hostname=None` (the whole string lands in `path`).
- `urlparse("myhost:8080")` → `scheme='myhost'`, `hostname=None`, `port=None` (the host is swallowed as a scheme).

In both cases `host` is `None` and `port` falls back to `80` (scheme is not `"https"`). `socket.create_connection((None, 80))` calls `getaddrinfo(None, 80)`, which — without `AI_PASSIVE` — resolves to the loopback address. So the probe silently tests **localhost:80** instead of the configured endpoint.

The two failure experiences are both bad:

1. **False positive.** Anything listening locally on port 80 (a dev web server, a reverse proxy, macOS's built-in httpd) makes the preset report `status: "ok"` with a suspiciously tiny `latency_ms`. The agent confidently swaps into a preset whose real endpoint may be down, unresolvable, or a typo — exactly the failure mode the no-caching design was built to prevent.
2. **Misleading negative.** With nothing local on the port, the user gets `status: "unreachable"` with an error like `[Errno 61] Connection refused` — which reads as "the API host refused me", sending them off to debug firewalls and provider outages when the actual problem is a missing `https://` in their manifest.

The design got here honestly: every URL in `_PROVIDER_DEFAULT_URLS` (lines 37-48) carries an `https://` scheme, so the happy path — provider defaults — never exercises the schemeless branch, and no test in `tests/test_preset_connectivity.py` passes a schemeless `base_url`. The gap only opens for explicit hand-authored `base_url` values, which is precisely the self-hosted/proxy use case where users type bare `host:port` strings.

Better looks like: normalize the URL before parsing (prepend `https://` when no scheme separator is present), and if the hostname is *still* unparseable, fail loudly with an explicit `invalid base_url` error rather than probing whatever `getaddrinfo(None, ...)` fancies.

## Current behavior

- `check_connectivity(provider, base_url, api_key_env)` (`src/lingtai_kernel/preset_connectivity.py:87-164`) resolves the probe URL via `_resolve_url()` (line 137), then at lines 146-148 does `urlparse(url)` / `parsed.hostname` / port fallback with no guard on `hostname is None`.
- `_probe_host(host, port, timeout)` (lines 51-62) passes the tuple straight to `socket.create_connection`, so `host=None` becomes a loopback connection.
- Concrete outcomes:
  - `base_url="api.openai.com"` → probes `localhost:80` → `"ok"` (false) or `"unreachable"` with `Connection refused` (misleading).
  - `base_url="myhost:8080"` → `urlparse` treats `myhost` as the scheme → probes `localhost:80` (note: not even port 8080).
- Callers affected: `check_many()` (same file, lines 167-182) and the preset listing in `src/lingtai_kernel/intrinsics/system/preset.py` (connectivity spec built around line 294), i.e. every `system(action='presets')` call.
- No existing test in `tests/test_preset_connectivity.py` covers schemeless or malformed `base_url` values.

1. **Add a URL-normalization helper** in `src/lingtai_kernel/preset_connectivity.py`, next to `_resolve_url()`:

   ```python
   def _parse_probe_target(url: str) -> tuple[str, int] | None:
       """Parse url into (host, port) for a TCP probe.

       Schemeless URLs (``api.openai.com``, ``myhost:8080``) are treated as
       https. Returns None if no hostname can be extracted.
       """
       if "://" not in url:
           url = f"https://{url}"
       parsed = urlparse(url)
       if not parsed.hostname:
           return None
       try:
           port = parsed.port  # raises ValueError on e.g. "host:notaport"
       except ValueError:
           return None
       return parsed.hostname, port or (443 if parsed.scheme == "https" else 80)
   ```

   Notes on the sketch:
   - The check must be `"://" not in url`, **not** `parsed.scheme == ""`, because `urlparse("myhost:8080")` returns a non-empty (bogus) scheme `myhost`. Prepending `https://` to the raw string yields `https://myhost:8080` → `hostname='myhost'`, `port=8080`, which is the user's intent.
   - `parsed.port` raises `ValueError` for non-numeric or out-of-range ports; catch it and treat as invalid rather than letting it escape `check_connectivity`.
   - Defaulting schemeless URLs to `https` (port 443) matches every entry in `_PROVIDER_DEFAULT_URLS` and the realistic intent of an API base URL.

2. **Use the helper in `check_connectivity()`**, replacing lines 146-148:

   ```python
   target = _parse_probe_target(url)
   if target is None:
       return {
           "status": "unreachable",
           "checked_at": datetime.now(timezone.utc).isoformat(),
           "latency_ms": None,
           "error": f"invalid base_url {url!r}: cannot determine host to probe",
       }
   host, port = target
   ```

   Keeping the existing `"unreachable"` status (rather than inventing a new status value) preserves the documented return shape in the docstring (lines 96-100) and requires no changes in `check_many()` or `src/lingtai_kernel/intrinsics/system/preset.py`, which consume the dict opaquely.

3. **Update the `check_connectivity` docstring** to mention that schemeless base URLs are probed as https and that a hostless URL yields `"unreachable"` with an `invalid base_url` error.

4. **Back-compat:** none needed. The return schema is unchanged; only behavior for previously-broken inputs changes (localhost probe → correct host probe, or explicit error). Well-formed `https://...` / `http://...` URLs take exactly the same path as before, including the http→port-80 fallback.

## Risks and alternatives

- **Guessing https for schemeless URLs could probe the wrong port** for a plaintext-http local server written as `myhost` with no port. In practice self-hosted endpoints are written as `host:port` (port preserved by the fix) and bare hostnames overwhelmingly mean https APIs. The failure mode of the guess (probing 443 instead of 80) is far milder than today's (probing localhost).
- **Alternative: reject schemeless URLs outright** with an `invalid base_url` error instead of normalizing. Simpler and unambiguous, but strictly worse UX: `api.openai.com` has exactly one plausible meaning, and hard-failing it turns a trivially recoverable typo into a broken preset. The plan reserves the hard error for genuinely unparseable inputs.
- **Alternative: validate `base_url` at preset load time** (in `load_preset` / manifest validation) rather than at probe time. Attractive long-term, but connectivity checks also receive URLs from other call paths (`check_many` specs), and load-time validation would not protect this function against future callers. Doing both is compatible; the probe-time guard is the minimal correct fix and can land first.
- **Regression risk is low:** the changed branch only triggers when `"://"` is absent or hostname/port extraction fails — inputs that today produce garbage results. All existing tests pass URLs with schemes.

## Testing plan

Add to `tests/test_preset_connectivity.py`, following the existing pattern of monkeypatching `_probe_host` to capture its arguments:

- `test_schemeless_base_url_probed_as_https` — `check_connectivity(provider=None, base_url="api.openai.com", api_key_env=None)` with a recording `_probe_host` stub; assert the probe received `("api.openai.com", 443)` and status is `"ok"`.
- `test_schemeless_host_port_base_url_keeps_port` — `base_url="myhost:8080"`; assert the probe received `("myhost", 8080)` (this is the `urlparse` scheme-swallowing case).
- `test_invalid_base_url_reports_error_without_probing` — `base_url="https://"` (and a parametrized case like `"http://host:notaport"`); assert status is `"unreachable"`, `error` contains `invalid base_url`, and the `_probe_host` stub was never called.
- `test_http_scheme_still_defaults_to_port_80` — regression guard: `base_url="http://myhost"` probes `("myhost", 80)`, confirming the normalization does not disturb explicit-scheme handling.
- Optionally, a stdlib-behavior canary `test_urlparse_schemeless_assumptions` asserting `urlparse("myhost:8080").hostname is None`, so a future Python change in `urlparse` semantics surfaces visibly.

huangzesen:
Fixes #741

## Problem

`check_connectivity()` parsed a preset's `base_url` with `urlparse()` and passed `parsed.hostname` straight to `socket.create_connection()`, with no guard on `hostname is None`. For a schemeless `base_url` — `api.openai.com` or `myhost:8080`, both natural to type in a hand-authored manifest — `urlparse` returns `hostname=None` (the string lands in `path`, or the host is swallowed as a bogus scheme). `create_connection((None, 80))` then resolves to **loopback** via `getaddrinfo(None, ...)`, so the probe silently tested `localhost:80` instead of the configured endpoint:

- **False positive:** anything listening locally on port 80 makes the preset report `"ok"` with a tiny `latency_ms` — the agent confidently swaps into a preset whose real endpoint may be down. Exactly the failure the module's no-caching design exists to prevent.
- **Misleading negative:** with nothing local, the user gets `"unreachable"` / `Connection refused`, which reads as "the API host refused me" and sends them debugging firewalls instead of a missing `https://`.
- The `http://host:notaport` case even let a `ValueError` from `parsed.port` escape `check_connectivity` entirely.

Verified on current main (`df2b523db810`) with a recording `_probe_host` stub:

```
base_url='api.openai.com'           probe=(None, 80)                     status=ok
base_url='myhost:8080'              probe=(None, 80)                     status=ok
base_url='http://host:notaport'     RAISED ValueError: Port could not be cast to integer value as 'notaport'
base_url='https://api.example.com'  probe=('api.example.com', 443)       status=ok
```

## Fix

A new `_parse_probe_target()` in `src/lingtai/kernel/preset_connectivity.py` normalizes a schemeless URL to `https` before parsing (the scheme test is `"://" not in url`, not `scheme == ""`, because `myhost:8080` yields a non-empty bogus scheme), catches the non-numeric-port `ValueError`, and returns `None` — **never a localhost fallback** — when no host can be extracted. `check_connectivity` then fails loud with an explicit `invalid base_url` `"unreachable"` result. The return schema is unchanged, so `check_many()` and `intrinsics/system/preset.py` need no changes; well-formed `http(s)://` URLs take exactly the same path as before (`http://host` → port 80 preserved).

After the fix:

```
base_url='api.openai.com'           probe=('api.openai.com', 443)        status=ok
base_url='myhost:8080'              probe=('myhost', 8080)               status=ok
base_url='https://'                 probe=None   status=unreachable error="invalid base_url 'https://': ..."
base_url='http://host:notaport'     probe=None   status=unreachable error="invalid base_url 'http://host:notaport': ..."
base_url='http://myhost'            probe=('myhost', 80)                 status=ok
```

## Validation

```
$ python -m pytest tests/test_preset_connectivity.py -q
23 passed in 0.93s

$ python -m pytest tests/test_presets.py tests/test_activate_preset.py -q
102 passed in 0.46s

$ git diff --check
(no output)
```

New regression tests: schemeless bare hostname → probed as https:443; `myhost:8080` → host+port preserved; hostless / non-numeric-port URLs → `unreachable` with `invalid base_url` and no probe call; explicit `http://` → port 80 preserved; plus a `urlparse` semantics canary.

### DIFF
--- a/src/lingtai/kernel/preset_connectivity.py
+++ b/src/lingtai/kernel/preset_connectivity.py
@@ -86,6 +86,30 @@ def _resolve_url(provider: str | None, base_url: str | None) -> str | None:
     return None
 
 
+def _parse_probe_target(url: str) -> tuple[str, int] | None:
+    """Parse a base_url into the (host, port) pair to TCP-probe.
+
+    Schemeless URLs (``api.openai.com``, ``myhost:8080``) are treated as
+    https. This matters because ``urlparse`` is hostile to them: it swallows
+    a bare ``host:port`` string as a bogus scheme and returns
+    ``hostname=None``, and ``socket.create_connection((None, ...))`` would
+    silently resolve to loopback — probing localhost instead of the
+    configured endpoint. Returns None when no hostname can be extracted
+    (e.g. ``https://`` or a non-numeric port) so the caller can fail loudly
+    rather than probing whatever ``getaddrinfo(None, ...)`` fancies.
+    """
+    if "://" not in url:
+        url = f"https://{url}"
+    parsed = urlparse(url)
+    if not parsed.hostname:
+        return None
+    try:
+        port = parsed.port  # raises ValueError on e.g. "host:notaport"
+    except ValueError:
+        return None
+    return parsed.hostname, port or (443 if parsed.scheme == "https" else 80)
+
+
 def check_connectivity(
     provider: str | None,
     base_url: str | None,
@@ -100,6 +124,11 @@ def check_connectivity(
          "checked_at": "<ISO timestamp>",
          "latency_ms": int (only on ok),
          "error": str | None}
+
+    Schemeless base URLs (``api.openai.com``, ``myhost:8080``) are probed as
+    https (port 443). A base URL with no extractable hostname yields
+    ``"unreachable"`` with an ``invalid base_url`` error rather than a
+    silent localhost probe.
     """
     # Local CLI-login providers (e.g. claude-code) have no network
     # endpoint and no API key — they authenticate through a local CLI/login
@@ -145,9 +174,15 @@ def check_connectivity(
             "error": f"no base_url and no default URL for provider {provider!r}",
         }
 
-    parsed = urlparse(url)
-    host = parsed.hostname
-    port = parsed.port or (443 if parsed.scheme == "https" else 80)
+    target = _parse_probe_target(url)
+    if target is None:
+        return {
+            "status": "unreachable",
+            "checked_at": datetime.now(timezone.utc).isoformat(),
+            "latency_ms": None,
+            "error": f"invalid base_url {url!r}: cannot determine host to probe",
+        }
+    host, port = target
 
     try:
         latency_ms = _probe_host(host, port, _PROBE_TIMEOUT_S)
--- a/tests/test_preset_connectivity.py
+++ b/tests/test_preset_connectivity.py
@@ -268,3 +268,94 @@ def test_kimi_code_missing_module_is_actionable(monkeypatch):
     assert result["status"] == "no_credentials"
     assert "kimi-code" in (result.get("error") or "")
     assert "no base_url" not in (result.get("error") or "")
+
+
+# ---------------------------------------------------------------------------
+# Schemeless / malformed base_url must never silently probe localhost
+# ---------------------------------------------------------------------------
+
+
+def test_schemeless_base_url_probed_as_https(monkeypatch):
+    """A bare hostname like `api.openai.com` is probed as https (port 443),
+    not resolved to localhost via getaddrinfo(None, ...)."""
+    from lingtai.kernel import preset_connectivity
+    captured = {}
+    def fake_probe(host, port, timeout):
+        captured["host"] = host
+        captured["port"] = port
+        return 12
+    with patch.object(preset_connectivity, "_probe_host", side_effect=fake_probe):
+        result = preset_connectivity.check_connectivity(
+            provider=None,
+            base_url="api.openai.com",
+            api_key_env=None,
+        )
+    assert captured == {"host": "api.openai.com", "port": 443}
+    assert result["status"] == "ok"
+    assert result["latency_ms"] == 12
+
+
+def test_schemeless_host_port_base_url_keeps_port(monkeypatch):
+    """`myhost:8080` is the urlparse scheme-swallowing case: the port must be
+    preserved and the host must not be dropped."""
+    from lingtai.kernel import preset_connectivity
+    captured = {}
+    def fake_probe(host, port, timeout):
+        captured["host"] = host
+        captured["port"] = port
+        return 7
+    with patch.object(preset_connectivity, "_probe_host", side_effect=fake_probe):
+        result = preset_connectivity.check_connectivity(
+            provider=None,
+            base_url="myhost:8080",
+            api_key_env=None,
+        )
+    assert captured == {"host": "myhost", "port": 8080}
+    assert result["status"] == "ok"
+
+
+@pytest.mark.parametrize("bad_url", ["https://", "http://host:notaport"])
+def test_invalid_base_url_reports_error_without_probing(bad_url):
+    """Hostless or non-numeric-port base URLs fail loud with an explicit
+    `invalid base_url` error — and never call the probe (no localhost
+    fallback, no escaping ValueError)."""
+    from lingtai.kernel import preset_connectivity
+    with patch.object(preset_connectivity, "_probe_host") as probe:
+        result = preset_connectivity.check_connectivity(
+            provider=None,
+            base_url=bad_url,
+            api_key_env=None,
+        )
+    assert result["status"] == "unreachable"
+    assert "invalid base_url" in result["error"]
+    probe.assert_not_called()
+
+
+def test_http_scheme_still_defaults_to_port_80(monkeypatch):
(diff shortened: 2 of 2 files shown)


huangzesen:

PR #1257, opened 2026-08-09
fix(retention): parse Z-suffixed finished_at as UTC regardless of host TZ

### LINKED ISSUE #729: Retention reporter misreads UTC "Z" timestamps as local time, shifting daemon ages by the host UTC offset
## Summary

`_parse_iso` in `src/lingtai_kernel/maintenance/retention.py` parses the daemon `finished_at` format `%Y-%m-%dT%H:%M:%SZ` into a **naive** `datetime` and then calls `.astimezone(timezone.utc)`, which interprets the naive value as **local time**. Since the daemon writer always emits UTC "Z" strings, every daemon run-dir age derived from `finished_at` is shifted by the host's UTC offset. On a UTC+8 host, `2026-06-05T12:00:00Z` is read as `2026-06-05T04:00:00Z` — an 8-hour error that can flip run dirs across the `older_than_days` cutoff, misclassifying them as retention candidates (or, on hosts west of UTC, keeping them protected longer than intended).

## Narrative

The retention reporter (`scan_retention`) walks each agent's `daemons/` directory and decides, per run dir, whether it is a deletion candidate, protected, or skipped. The age of a terminal daemon run is determined by `_daemon_state_and_age` (`src/lingtai_kernel/maintenance/retention.py:576-589`), and its **first-priority** age source is the `finished_at` field of `daemon.json`:

```python
finished_at = _parse_iso(data.get("finished_at") if isinstance(data, dict) else None)
if finished_at is not None:
    return state, finished_at, "daemon_finished_at", None
```

That field is written by the daemon runtime in `src/lingtai/core/daemon/run_dir.py:251-252`, which always produces a UTC timestamp with a literal `Z` suffix:

```python
@staticmethod
def _now_iso() -> str:
    return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
```

The parser on the reading side, `_parse_iso` (`src/lingtai_kernel/maintenance/retention.py:767-776`), tries two formats:

```python
for fmt in ("%Y-%m-%dT%H:%M:%SZ", "%Y-%m-%dT%H:%M:%S%z"):
    try:
        parsed = datetime.strptime(value, fmt)
        return parsed.astimezone(timezone.utc)
    except ValueError:
        continue
```

Here is the trap: the first format treats `Z` as a **literal character**, not a timezone designator, so `strptime` returns a naive `datetime`. Calling `.astimezone(timezone.utc)` on a naive datetime does not attach UTC — per Python semantics it assumes the naive value is in the **system local timezone** and converts from there. Reproduced directly:

```python
# TZ=Asia/Shanghai (UTC+8)
parsed = datetime.strptime("2026-06-05T12:00:00Z", "%Y-%m-%dT%H:%M:%SZ")
parsed.astimezone(timezone.utc)   # -> 2026-06-05 04:00:00+00:00  (WRONG, 8h early)
parsed.replace(tzinfo=timezone.utc)  # -> 2026-06-05 12:00:00+00:00  (correct)
```

Because the second format (`%z`) is only tried after the first raises `ValueError`, and the `Z`-literal format matches every string the writer produces, the buggy branch handles 100% of real `finished_at` values. (Ironically, since Python 3.7 `%z` accepts a literal `Z` and would parse these strings *correctly* as aware UTC — the dedicated `Z` format actively shadows the correct code path.)

The downstream effect: `scan_retention` computes `cutoff = now - timedelta(days=options.older_than_days)` (`retention.py:133`) and `_scan_daemons` compares `age_dt < cutoff` (`retention.py:359`) to decide between `newer_than_cutoff` protection and candidate status. On a host east of UTC (e.g. UTC+8), every daemon run looks up to 8 hours **older** than it is, so run dirs within 8 hours of the cutoff boundary are prematurely reported as `terminal_state_and_older_than_cutoff` candidates. On a host west of UTC, runs look **newer** than they are, so genuinely stale dirs are protected past the configured retention window; a just-finished daemon can even appear to have finished in the future (masked only by the `max(0.0, ...)` clamp in `_candidate`, `retention.py:607`).

The module is explicitly dry-run only ("no files were deleted, archived, moved, or rewritten", `retention.py:139`), so nothing is deleted directly — but the whole point of the report is that humans and agents act on it. The report's `age_seconds`, `age_days`, and `timestamp` fields (via `_iso(age_dt)`, `retention.py:617`) are all skewed too, so any consumer that trusts `daemon_finished_at`-sourced ages inherits the error.

How the design got here is easy to see: the two neighboring helpers, `_parse_mail_id_time` (`retention.py:747-752`) and `_parse_run_id_time` (`retention.py:755-764`), both parse naive timestamp formats and correctly stamp them with `.replace(tzinfo=timezone.utc)`. `_parse_iso` was presumably written to also handle explicit-offset strings via `%z`, and `.astimezone(timezone.utc)` is the right normalization *for aware values* — it was just applied uniformly to the naive `Z`-literal branch as well. Better looks like: naive parses get `.replace(tzinfo=timezone.utc)` (matching the documented writer contract and the sibling helpers), aware parses get `.astimezone(timezone.utc)`.

## Current behavior

- `_parse_iso` (`src/lingtai_kernel/maintenance/retention.py:767-776`) returns a UTC datetime shifted by the host's UTC offset for any input matching `%Y-%m-%dT%H:%M:%SZ` — i.e. for every `finished_at` the daemon runtime writes (`src/lingtai/core/daemon/run_dir.py:251-252`, plus the reconstruction paths in `src/lingtai/core/daemon/__init__.py` that reuse `DaemonRunDir._now_iso()` / `_utc_iso_from_timestamp`).
- `_daemon_state_and_age` (`retention.py:576-589`) propagates the skewed datetime with `age_source="daemon_finished_at"`, its highest-priority source, so the fallback `run_id`/`mtime` sources (which are computed correctly) never get a chance to mask the bug.
- `_scan_daemons` (`retention.py:359`) compares the skewed age against the cutoff, so terminal daemon run dirs within `|utc_offset|` of the `older_than_days` boundary are classified on the wrong side: false candidates east of UTC, false `newer_than_cutoff` protections west of it.
- Report output fields `age_seconds`, `age_days`, `timestamp` (`retention.py:604-617`) are skewed by the same offset for all `daemon_finished_at`-sourced entries.
- Existing tests in `tests/test_retention_report.py` write `finished_at` 60 days in the past against a 30-day cutoff (`tests/test_retention_report.py:53,65,92`), far outside any UTC offset, so the bug is invisible in CI unless the runner's TZ is manipulated and the fixture time sits near the boundary.

1. **Fix `_parse_iso`** in `src/lingtai_kernel/maintenance/retention.py` so naive parses are stamped as UTC and aware parses are normalized to UTC:

   ```python
   def _parse_iso(value: object) -> datetime | None:
       if not isinstance(value, str):
           return None
       for fmt in ("%Y-%m-%dT%H:%M:%SZ", "%Y-%m-%dT%H:%M:%S%z"):
           try:
               parsed = datetime.strptime(value, fmt)
           except ValueError:
               continue
           if parsed.tzinfo is None:
               return parsed.replace(tzinfo=timezone.utc)
           return parsed.astimezone(timezone.utc)
       return None
   ```

   This mirrors the writer contract (`_now_iso` always emits UTC `Z`) and matches the convention already used by `_parse_mail_id_time` (`retention.py:750`) and `_parse_run_id_time` (`retention.py:760-762`).

2. **Optional simplification (either is acceptable, keep one):** since the repo requires Python >= 3.11 (`pyproject.toml:13`), the whole loop can collapse to `datetime.fromisoformat`, which accepts both the `Z` suffix and explicit offsets since 3.11:

   ```python
   try:
       parsed = datetime.fromisoformat(value)
   except ValueError:
       return None
   if parsed.tzinfo is None:
       return parsed.replace(tzinfo=timezone.utc)
   return parsed.astimezone(timezone.utc)
   ```

   Note `fromisoformat` also accepts date-only and no-timezone strings, hence the explicit naive-means-UTC branch. If strictness about the exact two accepted shapes matters, prefer step 1's minimal fix.

3. **Audit for siblings.** `grep -rn "astimezone(timezone.utc)"` across `src/` and confirm no other call site applies it to a possibly-naive `strptime` result. In the current tree, `retention.py:773` is the only offender; `_iso` (`retention.py:811-814`) only receives aware datetimes from `_parse_*`/`_mtime_dt` and is fine once step 1 lands.

4. **No schema or migration work.** This is a read-side parsing fix; `daemon.json` contents, report JSON shape, and all category/reason strings are unchanged. Back-compat: reports generated on UTC hosts are already correct and stay byte-identical; reports on non-UTC hosts change to the correct values, which is the point.

## Risks and alternatives

- **Risk: naive-input semantics change.** Any hypothetical producer writing genuinely local-time strings with a `Z` suffix would now be read as UTC. But a `Z` suffix *means* UTC (RFC 3339), and the only producers in-tree (`DaemonRunDir._now_iso`, `_utc_iso_from_timestamp` in `src/lingtai/core/daemon/__init__.py`) emit real UTC — so this "risk" is precisely the bug being fixed.
- **Alternative: drop the `Z`-literal format and rely on `%z`.** Works on Python >= 3.7 because `%z` accepts `Z`, and is a one-line diff. Rejected as the primary plan because it fixes the bug by *accident of shadowing order* — a future reader could re-add a naive format ahead of it and silently reintroduce the bug. The explicit `tzinfo is None` branch documents the invariant.
- **Alternative: make the writer emit `+00:00` offsets instead of `Z`.** Rejected: it churns the on-disk `daemon.json` format, requires reading both old and new shapes forever, and touches far more surface than a read-side fix.
- **Alternative: compare in local time throughout.** Rejected: `run_id` and mail-id timestamps are UTC-encoded, `now` is taken in UTC, and mixing frames is how this class of bug proliferates.

## Testing plan

Add to `tests/test_retention_report.py` (the `_daemon` fixture at lines 48-66 already writes `finished_at` in the `Z` format):

1. `test_parse_iso_z_suffix_is_utc_regardless_of_local_tz` — unit test on `_parse_iso`: monkeypatch `TZ` to `Asia/Shanghai` (via `time.tzset` on POSIX, or parametrize over `TZ=UTC+8`/`TZ=UTC-8`; skip on Windows) and assert `_parse_iso("2026-06-05T12:00:00Z") == datetime(2026, 6, 5, 12, 0, 0, tzinfo=timezone.utc)` in every zone.
2. `test_parse_iso_explicit_offset_normalized_to_utc` — assert `_parse_iso("2026-06-05T20:00:00+08:00") == datetime(2026, 6, 5, 12, 0, 0, tzinfo=timezone.utc)` (guards the aware branch that must keep `astimezone`).
3. `test_daemon_finished_at_near_cutoff_not_shifted_by_host_tz` — integration test: create a terminal daemon run with `finished_at = NOW - timedelta(days=30) + timedelta(hours=4)` (4 hours *inside* the 30-day window), set `TZ` to a UTC+8 zone, run `scan_retention`, and assert the run dir is reported as protected with reason `newer_than_cutoff` and **not** counted in `classes["terminal_daemon_run"]["candidates"]`. Before the fix this asserts failure on the +8 host; after the fix it passes in any TZ.
4. `test_daemon_age_fields_match_finished_at` — assert the candidate's `timestamp` field round-trips the written `finished_at` string exactly and `age_days` equals the true UTC delta within tolerance, under a non-UTC `TZ`.
5. Keep `_parse_iso(None)`, `_parse_iso(123)`, and `_parse_iso("garbage")` returning `None` (regression guard if step 2's `fromisoformat` variant is chosen).

huangzesen:
Fixes #729.

## Problem

`_parse_iso` in `src/lingtai/kernel/maintenance/retention.py` parsed the daemon `finished_at` format `%Y-%m-%dT%H:%M:%SZ` into a **naive** datetime and then called `.astimezone(timezone.utc)`, which per Python semantics interprets the naive value as **local time**. Since the daemon writer always emits UTC "Z" strings, every daemon run-dir age derived from `finished_at` was shifted by the host's UTC offset (e.g. 8 hours early on a UTC+8 host), flipping run dirs across the `older_than_days` cutoff and skewing the report's `age_seconds`, `age_days`, and `timestamp` fields.

## Fix

Naive parses (the `Z`-literal branch, which matches every string the writer produces) are now stamped with `replace(tzinfo=timezone.utc)` instead of `astimezone(...)`; aware parses (explicit offsets) keep `astimezone(timezone.utc)` normalization. This mirrors the writer contract and the convention already used by `_parse_mail_id_time` / `_parse_run_id_time`. No schema or on-disk format changes.

## Tests

Added to `tests/test_retention_report.py` (5 new tests):

- `test_parse_iso_z_suffix_is_utc_regardless_of_local_tz` — unit test on `_parse_iso` under `TZ=Asia/Shanghai` and `TZ=America/New_York` (via `time.tzset()`)
- `test_parse_iso_explicit_offset_normalized_to_utc` — guards the aware branch that must keep `astimezone`
- `test_parse_iso_non_string_and_garbage_return_none` — regression guard
- `test_daemon_finished_at_near_cutoff_not_shifted_by_host_tz` — integration: a run 4h inside the 30-day cutoff stays `newer_than_cutoff` under a UTC+8 host (failed before the fix, passes after)
- `test_daemon_age_fields_match_finished_at` — candidate `timestamp` round-trips the written `finished_at`, `age_seconds`/`age_days` match the true UTC delta under a non-UTC TZ

Results (local, Python 3.14):

```
$ python3 -m pytest tests/test_retention_report.py -q
24 passed in 0.43s
```

Also ran the suite under `TZ=America/New_York` and `TZ=Asia/Tokyo`: 24 passed in each. `ruff check` passes on both changed files.

### DIFF
--- a/src/lingtai/kernel/maintenance/retention.py
+++ b/src/lingtai/kernel/maintenance/retention.py
@@ -1007,9 +1007,15 @@ def _parse_iso(value: object) -> datetime | None:
     for fmt in ("%Y-%m-%dT%H:%M:%SZ", "%Y-%m-%dT%H:%M:%S%z"):
         try:
             parsed = datetime.strptime(value, fmt)
-            return parsed.astimezone(timezone.utc)
         except ValueError:
             continue
+        if parsed.tzinfo is None:
+            # "Z" means UTC (RFC 3339); the daemon writer always emits UTC.
+            # Stamping instead of astimezone keeps the value independent of
+            # the host's local timezone (astimezone on a naive datetime would
+            # interpret it as local time and shift it by the UTC offset).
+            return parsed.replace(tzinfo=timezone.utc)
+        return parsed.astimezone(timezone.utc)
     return None
 
 
--- a/tests/test_retention_report.py
+++ b/tests/test_retention_report.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 import json
 import os
+import time
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 import pytest
 
 from lingtai.kernel.maintenance import RetentionOptions, TargetError, report_to_dict, scan_retention
+from lingtai.kernel.maintenance.retention import _parse_iso
 
 
 NOW = datetime(2026, 6, 27, 12, 0, 0, tzinfo=timezone.utc)
@@ -336,3 +338,68 @@ def test_json_report_is_deterministic(tmp_path):
     second = report_to_dict(scan_retention(agent, RetentionOptions(), _now=NOW))
 
     assert json.dumps(first, sort_keys=True) == json.dumps(second, sort_keys=True)
+
+
+def _set_tz(monkeypatch, tz: str) -> None:
+    """Switch the process local timezone for naive-datetime semantics."""
+    if not hasattr(time, "tzset"):
+        pytest.skip("time.tzset() unavailable on this platform")
+    monkeypatch.setenv("TZ", tz)
+    time.tzset()
+
+
+def test_parse_iso_z_suffix_is_utc_regardless_of_local_tz(monkeypatch):
+    _set_tz(monkeypatch, "Asia/Shanghai")
+    assert _parse_iso("2026-06-05T12:00:00Z") == datetime(
+        2026, 6, 5, 12, 0, 0, tzinfo=timezone.utc
+    )
+    _set_tz(monkeypatch, "America/New_York")
+    assert _parse_iso("2026-06-05T12:00:00Z") == datetime(
+        2026, 6, 5, 12, 0, 0, tzinfo=timezone.utc
+    )
+
+
+def test_parse_iso_explicit_offset_normalized_to_utc(monkeypatch):
+    _set_tz(monkeypatch, "Asia/Shanghai")
+    assert _parse_iso("2026-06-05T20:00:00+08:00") == datetime(
+        2026, 6, 5, 12, 0, 0, tzinfo=timezone.utc
+    )
+
+
+def test_parse_iso_non_string_and_garbage_return_none():
+    assert _parse_iso(None) is None
+    assert _parse_iso(123) is None
+    assert _parse_iso("garbage") is None
+
+
+def test_daemon_finished_at_near_cutoff_not_shifted_by_host_tz(tmp_path, monkeypatch):
+    # 4 hours inside the 30-day window: the run is 29d20h old, so it must be
+    # protected as newer_than_cutoff. Before the fix, a UTC+8 host read the
+    # Z timestamp 8 hours early and misclassified it as a candidate.
+    _set_tz(monkeypatch, "Asia/Shanghai")
+    agent = _agent(tmp_path)
+    _daemon(agent, finished_at=NOW - timedelta(days=30) + timedelta(hours=4))
+
+    data = _report(agent)
+
+    assert data["classes"]["terminal_daemon_run"]["candidates"] == 0
+    assert data["classes"]["terminal_daemon_run"]["protected"] == 1
+    reasons = {item["reason"] for item in data["protected"]}
+    assert "newer_than_cutoff" in reasons
+
+
+def test_daemon_age_fields_match_finished_at(tmp_path, monkeypatch):
+    _set_tz(monkeypatch, "Asia/Shanghai")
+    agent = _agent(tmp_path)
+    finished_at = NOW - timedelta(days=45)
+    run = _daemon(agent, finished_at=finished_at)
+
+    data = _report(agent)
+
+    candidates = [c for c in data["candidates"] if Path(c["path"]) == run]
+    assert len(candidates) == 1
+    candidate = candidates[0]
+    assert candidate["age_source"] == "daemon_finished_at"
+    assert candidate["timestamp"] == finished_at.strftime("%Y-%m-%dT%H:%M:%SZ")
+    assert candidate["age_seconds"] == pytest.approx(45 * 86400.0, abs=1.0)
+    assert candidate["age_days"] == pytest.approx(45.0, abs=0.01)


huangzesen:

PR #1258, opened 2026-08-09
fix(mail): guard listen() against double-start poll thread (#675)

### LINKED ISSUE #675: MailService.listen() is not idempotent: second call spawns second poll thread; existing RuntimeError guard in lifecycle.py is dead code
## Summary

`MailService.listen()` has no guard against being called twice. A second call spawns a second poll thread, causing double-dispatch of the same inbox messages. The caller in `lifecycle.py` attempts to guard against this with `except RuntimeError: pass`, but `listen()` never raises `RuntimeError` — the guard is dead code.

## Files

- `src/lingtai_kernel/services/mail.py` — `listen()` method
- `src/lingtai_kernel/base_agent/lifecycle.py` — lines 180-184

## Evidence

```python
# lifecycle.py L180-184 — caller guard (dead code)
try:
    self._mail_service.listen(...)
except RuntimeError:
    pass  # Already listening
```

The intent is clear: the author expected `listen()` to raise `RuntimeError` on re-entry. But `listen()` has no such check:

```python
# mail.py — listen() method
def listen(self, ...):
    # No guard against double-listen
    self._poll_thread = threading.Thread(...)
    self._poll_thread.start()
```

## Race scenario

1. Agent refreshes or restarts the mail service
2. `listen()` is called a second time
3. Two poll threads run concurrently
4. Both threads scan the inbox every 0.5s
5. Both threads try to dispatch the same incoming message
6. Double-dispatch: the agent processes the same email twice

## Impact

Double-dispatch of mail messages. The second processing is typically idempotent (reading an already-read message), but any side effects (tool calls, replies) would fire twice.

## Fix

Add a re-entry guard in `listen()` that raises `RuntimeError`:

```python
def listen(self, ...):
    if self._poll_thread is not None and self._poll_thread.is_alive():
        raise RuntimeError("MailService already listening")
    self._poll_thread = threading.Thread(...)
    self._poll_thread.start()
```

This makes the existing `except RuntimeError` guard in lifecycle.py work as intended.

huangzesen:
Fixes #675.

## Problem

`PosixFilesystemMailAdapter.listen()` (the production mail transport behind the Core `MailTransportPort`) has no re-entry guard: a second call unconditionally replaces `_poll_thread` and spawns a second poll loop. Both loops scan the same inbox on the same 0.5 s cadence, sharing `_seen` only through an unsynchronized set, so the same incoming message can be dispatched twice. The caller in `base_agent/lifecycle.py` wraps `listen()` in `except RuntimeError: pass` ("Already listening"), but `listen()` never raised — that guard was dead code.

## Fix

- `src/lingtai/adapters/posix/mail.py`: `listen()` now raises `RuntimeError("MailService already listening")` when a poll thread is already alive, before snapshotting `_seen` or touching `_poll_stop`. This is exactly the contract the issue describes and the lifecycle caller already expects; the `except RuntimeError: pass` guard in `lifecycle.py` now works as intended (no lifecycle change needed).
- `src/lingtai/kernel/mail_transport/__init__.py`: document the re-entry contract on `MailTransportPort.listen()` (raises on re-entry; `listen()` after `stop()` restarts delivery).
- `tests/test_filesystem_mail.py`: two new tests — re-entry while polling raises `RuntimeError` and leaves exactly one poll thread; `listen()` after `stop()` is allowed (refresh/restart path).

## Tests

- `pytest tests/test_filesystem_mail.py -k TestListen` → 6 passed (4 existing + 2 new)
- `pytest tests/test_filesystem_mail.py tests/test_mail_transport.py tests/test_services_mail.py` → 53 passed
- `pytest tests/test_agent.py -k mail` → 8 passed (incl. `test_mail_start_wires_listener`, which exercises the lifecycle listen wiring)

### DIFF
--- a/src/lingtai/adapters/posix/mail.py
+++ b/src/lingtai/adapters/posix/mail.py
@@ -197,7 +197,15 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         Existing messages are recorded in ``_seen`` so they are not
         re-delivered.  New directories that appear with a ``message.json``
         trigger *on_message*.
+
+        Re-entry guard: raises ``RuntimeError`` when a poll thread is already
+        running, so a second ``listen()`` cannot spawn a second poll thread
+        that double-dispatches the same messages.  The agent lifecycle
+        (``base_agent/lifecycle.py``) relies on this exception to treat a
+        repeated ``listen()`` as "already listening".
         """
+        if self._poll_thread is not None and self._poll_thread.is_alive():
+            raise RuntimeError("MailService already listening")
         # Snapshot existing inbox entries so we don't re-notify
         if self._inbox_dir.is_dir():
             for entry in self._inbox_dir.iterdir():
--- a/src/lingtai/kernel/mail_transport/__init__.py
+++ b/src/lingtai/kernel/mail_transport/__init__.py
@@ -47,6 +47,11 @@ class MailTransportPort(ABC):
         payloads are passed to ``on_message`` according to the adapter's
         delivery semantics; the Port does not promise end-to-end exactly-once
         delivery across process crashes or restarts.
+
+        Re-entry: calling ``listen()`` while delivery is already running raises
+        ``RuntimeError``; callers such as the agent lifecycle treat that as
+        "already listening" and ignore it. A fresh ``listen()`` after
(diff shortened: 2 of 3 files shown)


huangzesen:

PR #1228, opened 2026-08-08
chore(kernel): version examples, CI docs, dead i18n (staleness audit)

huangzesen:
## Summary

Part of the K-4 slice of the kernel-side staleness audit (`~21 findings`, P2).

- Bump stale version examples (`v0.16.4` / `v0.10.10` / `v0.17.1` → current `0.19.5` series) in `wheels.yml`, `RELEASING.md`, `README.md`, `setup.py`, `sync_gitee_mirror.py`, and `generate_release_manifest.py`.
- Swap `README.md`'s PyPI badge for a GitHub Releases badge — it contradicted `RELEASING.md`'s explicit "No PyPI publication" policy, and no PyPI publish step exists anywhere in CI.
- Correct `check_docs_governance.py`'s docstring claim that no PR-triggered pytest workflow exists — `kernel-windows-pr.yml` and `shell-windows-pr.yml` both already run pytest on `pull_request`.
- Refresh `docs.yaml`'s stale "244 candidates" tracked-markdown-file count to the current `304` (verified via `git ls-files | grep -c '\.md$'`, which matches the checker's own discovery logic).
- Check off `docs/plans/2026-06-25-fsutil-migration.md` and `docs/plans/2026-07-14-powershell-adapter-readiness.md` items that have since landed in production code (notification-store atomic writes, token-ledger JSONL helpers, Windows workdir-lease/refresh-watcher/avatar-launcher selector wiring), and correct stale `intrinsics/email`/`intrinsics/soul` paths to their current `tools/email`/`tools/soul` locations.
- Mark the `v0.15.2` release report's publish checklist as the completed release it now documents (release commits `95bb06ef`/`603e5d85`, tag `v0.15.2`, verified via `gh release view`).
- Sync `README.zh.md` and `README.wen.md`'s architecture entry-point tables with `README.md` — both were missing the `dev-guide-skill/SKILL.md` and `CONTRACT.md` rows plus the intro paragraph.
- Drop the dead i18n keys `system.molt_wiped` and `system.new_mail.no_subject` from `src/lingtai/kernel/i18n/{en,zh,wen}.json` — zero code references, superseded by the `tools/i18n` catalog. **Split into its own commit** so it can be reverted independently if a downstream consumer surfaces.

## Test plan

- [x] `python3 -m json.load` sanity check on all three edited i18n files
[...]

PR #1207, opened 2026-08-06
Refactor current reasoning-effort wiring into explicit seams

TZZheng:
Related: #1197

## What this PR does

This is the behavior-preserving first step in the live reasoning-effort plan. It makes the kernel's **current wiring** explicit without repairing or expanding any provider behavior.

The refactor gives the six approved responsibilities separate owners:

1. an immutable current-wiring route catalogue keyed by integration, provider/alias, endpoint identity, API compatibility, wire, model predicate, and CLI phase;
2. one main-agent omission-baseline resolver;
3. shared validation plus an explicit, currently empty normalization layer;
4. provider/integration-local mappers whose one construction result supplies the actual JSON/argv/env/session delta at the real construction point;
5. mechanically exact observability derived from that same deeply frozen result; and
6. immutable policy capture at the existing session-construction seam and original Python evaluation point.

The route-local owners cover the existing OpenAI Responses/Chat/Codex paths, DeepSeek and MiMo, Zhipu/GLM, OpenRouter, Anthropic/MiniMax/custom-Anthropic, both Gemini wires, Claude Code and Kimi Code first/resume commands, all four native-daemon handoffs, external-daemon startup argv, and every supported external-CLI ask/resume builder. Static catalogue facts remain descriptive; exact startup and ask evidence is produced by the callable construction owner.

## Behavior-preservation boundary

This PR deliberately preserves the current accepted/rejected sets, JSON/argv/env bytes, defaults, omissions, drops, coercions, known defects, validation error text/order, first/resume behavior, native-daemon forced `default`, startup-versus-ask phase differences, session policy capture timing, and pre-dispatch `llm_call` timing.

It does **not** add or repair provider effort behavior. It also does not add live get/set/clear, per-dispatch revisioning, persistence or lifecycle recovery, operator protocol, TUI controls, daemon-local live control, or any later #1197 node. The only additive runtime surface is the mechanically derived optional `llm_call.reasoning` evidence; uninstrumented routes retain their previous event shape.

## Test-first chronology

The unchanged Phase-A characterization fence was written before the refactor and contains 45 cases that pass on canonical main. It locks the existing route matrix, bytes, defaults, omissions, drops, coercions, phase differences, and preserved defects.

[...]

PR #1200, opened 2026-08-06
docs(windows): Git Bash/MSYS pitfall knowledge capture (PR-10)

huangzesen:
## Summary

Docs-only knowledge capture (PR-10) for Git Bash / MSYS pitfalls so future Windows-shell PRs do not rediscover them.

The kernel already has a Git Bash fallback dialect (`src/lingtai/adapters/windows/gitbash.py`) and an opt-in WSL adapter (`src/lingtai/adapters/windows/wsl.py`); the pitfall list from the Windows adapter research (`§1.1.3` / `§PR-10`) is now recorded where Windows adapter work will find it:

### Changes
- **`src/lingtai/adapters/windows/ANATOMY.md`** — new **Git Bash / MSYS pitfalls** subsection in `## Notes` covering, with one-line explanations and mitigations:
  - `MSYS_NO_PATHCONV=1` + `MSYS2_ARG_CONV_EXCL=*` for MSYS2 argument path-mangling
  - `/c/...` (MSYS drive-mount) path translation across the Git Bash ↔ native-Windows boundary
  - prepending `C:\Program Files\Git\usr\bin` to PATH for MSYS coreutils
  - the non-login `bash -c` coreutils gap (use `bash -lc`, as `gitbash.py` does)
  - the ASLR spawn-failure class (`0xc0000142` / `0xc0000005`) and its bounded-retry implication
  - WSL-bash-vs-Git-bash ambiguity — never assume which `bash` is on PATH; `discover_git_bash()` rejects the System32/SysWOW64 WSL launcher
  - `gitbash.py` / `wsl.py` added to `related_files`
- **`src/lingtai/tools/bash/ANATOMY.md`** — pointer line from the shell capability anatomy to the new Git Bash / MSYS notes in the Windows adapter ANATOMY.

### Validation
- Docs only: `git status` shows exactly the two `.md` files changed; no Python/test changes.
- `tests/test_architecture_documents.py`: the one failing test (`test_governed_child_contracts_have_reciprocal_anatomy_pairs`, stale `kernel/daemon_supervisor/supervisor.py` reference) fails identically on pristine `HEAD` — pre-existing, unrelated to this change. The related_files validation tests pass with the additions.

No merge performed.

PR #1194, opened 2026-08-05
fix(codex): bound token-expired recovery

ZigongXu:
## Summary

- recover one structural `401 / token_expired` failure before the first response event, without redrawing the bound Codex account
- validate account, generation, auth path, and rejected token under the account-context lock, then publish REST/WS credentials and create the retry request in the same transaction
- fail closed after the request-owned recovery budget is consumed, with exact kernel-owned replay-terminal flags, render-safe error handling, rollback/reset, and partial-stream priority
- add adversarial BaseAgent/session/auth regressions and update the capability contract plus ANATOMY navigation

## Safety properties

- any response event—including hidden reasoning—closes the replay boundary
- provider-owned marker storage/getters/setters/`__dict__` are never trusted; only exact `LLMReplayTerminalError` carries terminal flags
- partial output is checked before no-AED, so already-visible text/tool output ends the turn without replay
- secondary account telemetry sees only the original exception and cannot replace the replay-safe wrapper
- rollback snapshot construction, assistant publication, continuation state, response ID, token accounting, and account-success bookkeeping share one fail-closed recovered tail

## Validation

- focused auth/session/BaseAgent suite: **94 passed**
- exact repository-derived 23-file Codex/OpenAI/AED group: **511 passed**
- all seven changed Python files compile; `git diff --check` is clean
- adversarial probes for hostile `__dict__`/marker/render hooks, raising telemetry, and raising entry-ID hashing passed
- added-diff scan: no private path/ID, control character, credential-shaped string, or production literal credential
- architecture/anatomy checks: **1 failed, 11 passed**; the sole failure is the pre-existing missing `src/lingtai/kernel/daemon_supervisor/supervisor.py`, independently reproduced at the exact baseline

## Independent review

Final independent review: **PASS WITH NITS**. No P0/P1/P2 finding; all three last-round blockers were closed. The mechanical ANATOMY anchor nit was corrected before this commit. The remaining optional nit concerns only the reason string used when partial-stream termination suppresses post-turn save; it does not change control flow, persistence safety, or replay behavior.

Fixes Lingtai-AI/lingtai#791

PR #1259, opened 2026-08-09
fix(llm): disambiguate worker-raised TimeoutError in watchdog via future.done()

### LINKED ISSUE #738: llm_utils watchdog misclassifies a worker-raised TimeoutError as "future not done", causing a WorkerStillRunningError false positive (and a hot log-spam loop)
## Summary

On Python >= 3.11 (this repo's floor, `pyproject.toml` `requires-python = ">=3.11"`), `concurrent.futures.TimeoutError` **is** the builtin `TimeoutError`, and `socket.timeout` / `asyncio.TimeoutError` are aliases of it. The watchdog loop in `src/lingtai_kernel/llm_utils.py` uses `except TimeoutError:` around `future.result(timeout=...)` to mean "the future has not settled yet" — but that clause equally catches the case where the future **has** settled with a worker-raised `TimeoutError`. The result is a hot busy-loop that spams "LLM API not responding" warnings until `retry_timeout` expires, followed by a spurious `WorkerStillRunningError` that poisons the ChatInterface and forces the agent asleep, even though the worker finished and already ran its cleanup. The fix is to check `future.done()` inside the `except TimeoutError` blocks to disambiguate.

## Narrative

The LLM send path in `lingtai-kernel` deliberately runs every provider call on a worker thread and watches it from the main thread. `_send` (`src/lingtai_kernel/llm_utils.py:47-73`) polls the future in `_LLM_WARN_INTERVAL`-second slices so it can log periodic "not responding" warnings, and when the overall `retry_timeout` budget is exhausted it calls `_wait_for_worker_settle` (`src/lingtai_kernel/llm_utils.py:76-106`) to give the worker a `_WORKER_SETTLE_GRACE = 5.0` second window to finish its own HTTP-timeout cleanup (`drop_trailing` on the shared ChatInterface). If the worker is *still* alive after that grace, `_wait_for_worker_settle` raises `WorkerStillRunningError` — a deliberately scary signal meaning "the worker thread may still be mutating the shared ChatInterface, do not touch it." The AED loop in `src/lingtai_kernel/base_agent/turn.py` (around lines 648-666, and again at ~1132, ~1154, ~1222) treats that error fail-closed: it marks the interface poisoned, writes a worker-hang artifact, skips the chat save, puts the agent ASLEEP, and requests a forced refresh/relaunch. That is exactly the right response to a genuinely wedged worker.

The problem is how both functions detect "the future hasn't settled yet":

```python
# src/lingtai_kernel/llm_utils.py:63-73 (_send)
        try:
            return future.result(timeout=wait)
        except TimeoutError:
            elapsed = time.monotonic() - t0
            if elapsed >= retry_timeout:
                _wait_for_worker_settle(future, elapsed, agent_name)
                raise TimeoutError(f"LLM API call timed out after {elapsed:.0f}s")
            _logger.warning(
                "[%s] LLM API not responding after %.0fs...",
                agent_name, elapsed,
            )
```

This idiom was correct on Python 3.7 and earlier, where `concurrent.futures.TimeoutError` was a distinct class. Since Python 3.8 it is an alias of the builtin `TimeoutError`, and since 3.10/3.11 `socket.timeout` and `asyncio.TimeoutError` are aliases of it too. The repo requires Python >= 3.11, so `except TimeoutError:` catches **both** of these completely different situations:

1. `Future.result(timeout=wait)` gave up waiting — the worker is still running.
2. The worker function itself raised a `TimeoutError` (e.g. a raw `socket.timeout` bubbling out of a lower networking layer, an `asyncio.TimeoutError`, or any adapter that raises the builtin) — the future **is done**, and `Future.result` re-raised the worker's exception.

The provider adapters re-raise worker exceptions verbatim after running their cleanup (`except Exception: ... self._interface.drop_trailing(...); raise`), so a builtin `TimeoutError` from the worker arrives at `_send` unwrapped. When that happens, case 2 masquerades as case 1, and everything downstream goes wrong in sequence:

- **Hot loop with log spam.** Suppose the worker settles with a `socket.timeout` 30 seconds into a 300-second `retry_timeout`. On the next loop iteration, `future.result(timeout=wait)` raises the worker's `TimeoutError` *immediately* — a done future never waits. The `except` clause logs the "LLM API not responding" warning and loops again, and again, with no sleep in between. The main thread pegs a core and emits thousands of misleading warning lines for the remaining ~270 seconds, all while the worker thread is long dead.
- **Wrong grace-period verdict.** When `retry_timeout` finally expires, `_send` calls `_wait_for_worker_settle`, whose body has the same flaw (`src/lingtai_kernel/llm_utils.py:88-106`): `future.result(timeout=_WORKER_SETTLE_GRACE)` instantly re-raises the worker's `TimeoutError`, the `except TimeoutError:` clause logs "LLM worker thread still running after ... grace" and raises `WorkerStillRunningError`. Note the irony: the function already has an `except Exception: pass` fallback for "worker raised something other than timeout — that's fine, its except-block already ran drop_trailing", so a worker that raises `ValueError` is handled gracefully while a worker that raises `socket.timeout` — the *most likely* failure mode for an LLM HTTP call — is escalated to the poison path.
- **Heavyweight false-positive recovery.** The AED loop receives `WorkerStillRunningError` and does what it must for a real hang: poisons the interface, refuses retry, writes recovery artifacts, forces the agent ASLEEP, and requests a refresh/relaunch (`src/lingtai_kernel/base_agent/turn.py:648-666`). For a transient network timeout that the worker already cleaned up after (its `drop_trailing` ran, the ChatInterface is consistent), the correct outcome would have been an ordinary `TimeoutError` and an in-process AED retry. Instead the agent takes a full outage.

The design got here honestly: the two-timeout architecture (per-call HTTP timeout plumbed via `_SubmitFn` at `src/lingtai_kernel/llm_utils.py:109-134`, plus the main-thread watchdog) is sound, and the SDK-level timeout exceptions from the Anthropic/OpenAI clients are *not* builtin `TimeoutError` subclasses, so the common path never exposed the aliasing. But `llm_utils` is generic over any ChatInterface implementation, and the builtin-`TimeoutError` family (`socket.timeout`, `asyncio.TimeoutError`, hand-raised `TimeoutError` in custom adapters) is exactly the kind of exception a networking worker emits. Better looks like: the watchdog asks the future directly whether it settled (`future.done()`), and only interprets `TimeoutError` as "still waiting" when the answer is no.

## Current behavior

- `_send` (`src/lingtai_kernel/llm_utils.py:47-73`): if the submitted worker settles with a builtin `TimeoutError` (e.g. `socket.timeout`), the loop misreads it as "not responding", busy-loops logging one warning per iteration (no wait, since `future.result` on a done future returns/raises immediately) until `retry_timeout` elapses, then calls `_wait_for_worker_settle`.
- `_wait_for_worker_settle` (`src/lingtai_kernel/llm_utils.py:76-106`): `future.result(timeout=_WORKER_SETTLE_GRACE)` instantly re-raises the worker's `TimeoutError`, which is caught by `except TimeoutError:` and converted to `WorkerStillRunningError(elapsed=..., grace=..., agent_name=..., future=...)` — even though `future.done()` is `True`. Non-timeout worker exceptions are correctly swallowed by the `except Exception:` clause at lines 102-106.
- Downstream, the AED loop (`src/lingtai_kernel/base_agent/turn.py:648-666` and the tc-wake path at ~1222) treats the false `WorkerStillRunningError` as a poisoned-interface hang: no retry, agent ASLEEP, forced refresh — for a ChatInterface that is actually safe (the adapter's `except Exception` block already ran `drop_trailing` before re-raising).

Neither function consults `future.done()`; `grep` confirms no `done()` call anywhere in `src/lingtai_kernel/llm_utils.py`.

1. **Fix `_wait_for_worker_settle`** (`src/lingtai_kernel/llm_utils.py`). Inside the `except TimeoutError:` block, check `future.done()` first. If the future settled, the caught exception was the *worker's own* `TimeoutError`; its cleanup already ran, so treat it exactly like the existing `except Exception:` branch — return and let the caller raise its ordinary watchdog `TimeoutError`:

   ```python
   try:
       future.result(timeout=_WORKER_SETTLE_GRACE)
   except TimeoutError:
       if future.done():
           # The worker settled with its own TimeoutError (socket.timeout,
           # asyncio.TimeoutError, and concurrent.futures.TimeoutError are
           # all the builtin TimeoutError on >=3.11). Its except-block
           # already ran drop_trailing — same as any other worker error.
           return
       _logger.error(...)
       raise WorkerStillRunningError(...)
   except Exception:
       pass
   ```

2. **Fix `_send`**. In its `except TimeoutError:` block, check `future.done()` before interpreting the exception as "not responding". If the future is done, surface its settled outcome directly — `future.result(timeout=0)` re-raises the worker's exception (the common case) or returns a value (covers the benign race where the future settled between the wait expiring and the `done()` check):

   ```python
   try:
       return future.result(timeout=wait)
   except TimeoutError:
       if future.done():
           # Settled: either the worker raised its own TimeoutError, or it
           # finished during the race after our wait expired. Either way the
           # worker is gone and its cleanup ran; surface the real outcome.
           return future.result(timeout=0)
       elapsed = time.monotonic() - t0
       if elapsed >= retry_timeout:
           ...
       _logger.warning(...)
   ```

   A worker-raised `TimeoutError` now propagates out of `_send` immediately as an ordinary exception (no `WorkerStillRunningError`, no warning spam), and the AED transient-retry path handles it like any other provider failure.

3. **No changes needed in `base_agent/turn.py`** or `worker_recovery`: the poison/ASLEEP path remains correct for genuine hangs; this change only stops feeding it false positives.

4. **No schema/data changes, no migration.** The public surface (`send_with_timeout`, `send_with_timeout_stream`, `WorkerStillRunningError`) is unchanged; only the classification of one failure mode changes, from "poison + sleep" to "ordinary worker exception". That is a behavioral bug fix, not a compatibility break — callers already handle ordinary `TimeoutError` via the transient-retry branch.

5. **Optional hardening / docstring update**: note in the `_send` and `_wait_for_worker_settle` docstrings that `except TimeoutError` alone is ambiguous on >= 3.11 and `future.done()` is the discriminator, so the idiom is not reintroduced later.

## Risks and alternatives

- **Behavior change is the point, but review it**: a worker-raised `TimeoutError` previously ended in `WorkerStillRunningError` (agent asleep, refresh); afterward it becomes an ordinary transient error eligible for AED in-process retry. This is safe because the adapters' `except Exception` blocks run `drop_trailing` *before* re-raising, so by the time the future is done the shared ChatInterface is already reverted. Anyone writing a new adapter must preserve that invariant, but that requirement predates this fix.
- **Race between wait-expiry and settle**: `future.done()` can flip to `True` in the instant after `future.result(timeout=wait)` raised the wait timeout. The plan handles this deliberately — if `done()` is true for *either* reason, the settled outcome is retrievable without blocking (`future.result(timeout=0)`), and both interpretations converge on the correct answer.
- **Alternative A — restructure with `concurrent.futures.wait([future], timeout=wait)`**: `wait()` never raises, returning `(done, not_done)` sets, which eliminates the ambiguity entirely and removes exception-based control flow. This is arguably cleaner, but it is a larger rewrite of the loop for the same behavior; the `done()` check is a minimal, obviously-correct diff. Reasonable follow-up refactor.
- **Alternative B — wrap the worker callable to re-wrap `TimeoutError`** into a private marker exception before it reaches the future. Rejected: it changes the exception type observed by callers and logs, touches `_SubmitFn`, and solves at the wrong layer.
- **Alternative C — catch `concurrent.futures.TimeoutError` "specifically"**: does nothing, since it *is* the builtin `TimeoutError` on the supported Python versions — that aliasing is the root cause.
- **Edge case**: `WorkerStillRunningError` itself subclasses `TimeoutError` (`src/lingtai_kernel/llm_utils.py:27`). A worker function will never raise it in practice, and even if one did, the `done()` check would correctly classify it as a settled worker exception.

## Testing plan

Add to `tests/test_llm_utils.py` (which already covers the settle path, e.g. `test_wait_for_worker_settle_raises_when_future_still_running`):

1. `test_wait_for_worker_settle_returns_when_worker_raised_timeout` — build a `Future`, call `set_exception(TimeoutError("worker socket timeout"))`, call `_wait_for_worker_settle(future, elapsed=10.0, agent_name="t")`; assert it returns `None` and does **not** raise `WorkerStillRunningError`.
2. `test_send_surfaces_worker_timeout_immediately` — a worker fn that sleeps ~0.05 s and raises `socket.timeout` (i.e. builtin `TimeoutError`), `retry_timeout=30`; assert `_send` raises a plain `TimeoutError` that is **not** a `WorkerStillRunningError`, and that it does so promptly (wall-clock elapsed well under `retry_timeout`, e.g. < 2 s), proving the hot loop is gone.
3. `test_send_worker_timeout_produces_no_warning_spam` — same setup with `caplog`; assert zero (or at most one) "LLM API not responding" warnings are recorded.
4. `test_send_returns_value_when_future_settles_during_race` — monkeypatch or use a worker that completes with a value just after the first wait slice; assert `_send` returns the value (covers the `future.result(timeout=0)` race branch).
5. Regression: existing tests must still pass unchanged — `test_send_with_timeout_waits_for_worker_to_settle_after_timeout`, `test_wait_for_worker_settle_raises_when_future_still_running`, and `test_send_with_timeout_raises_worker_still_running_when_worker_never_settles` prove genuine hangs still escalate to `WorkerStillRunningError`, and the AED fail-closed tests in `tests/test_aed_recovery.py` prove the poison path is untouched.

huangzesen:
Fixes #738.

## Problem

On Python >= 3.11 (the repo floor), `concurrent.futures.TimeoutError`, `socket.timeout` and `asyncio.TimeoutError` are all the builtin `TimeoutError`. The watchdog in `src/lingtai/kernel/llm_utils.py` used `except TimeoutError:` around `future.result(timeout=...)` to mean "future not settled yet" — but that clause equally catches a **settled** future whose worker raised its own builtin `TimeoutError` (e.g. a raw `socket.timeout` from the HTTP layer). Consequences:

- `_send` busy-loops logging "LLM API not responding" warnings (no sleep, a done future never waits) until `retry_timeout` expires.
- `_wait_for_worker_settle` converts the worker's own `TimeoutError` into a spurious `WorkerStillRunningError`, which poisons the ChatInterface and forces the agent ASLEEP even though the worker already finished and ran its cleanup (`drop_trailing`).

## Fix

Check `future.done()` first inside both `except TimeoutError` blocks:

- `_send`: if the future is done, surface the settled outcome directly with `future.result(timeout=0)` (re-raises a worker-raised `TimeoutError` as an ordinary exception; returns the value in the wait-expiry race).
- `_wait_for_worker_settle`: if the future is done, return — the worker's own `TimeoutError` already ran its cleanup, same as the existing `except Exception` branch.

Genuine hangs still escalate to `WorkerStillRunningError` (existing tests unchanged). Docstrings note the ambiguity so the idiom is not reintroduced.

## Tests

4 new tests in `tests/test_llm_utils.py`:
- `test_wait_for_worker_settle_returns_when_worker_raised_timeout`
- `test_send_surfaces_worker_timeout_immediately` (prompt, <2s, not WorkerStillRunningError)
- `test_send_worker_timeout_produces_no_warning_spam` (caplog)
- `test_send_returns_value_when_future_settles_during_race`

Run on Python 3.14.6:
- `pytest tests/test_llm_utils.py` → **12 passed** (8 existing + 4 new)
- `pytest tests/test_aed_recovery.py tests/test_tc_wake_chat_not_ready.py` → **20 passed**
- `pytest tests/test_retroactive_history_compaction.py tests/test_llm_adapter_timeouts.py tests/test_llm_service.py` → **59 passed**

Total **91 passed**, 0 failures.

### DIFF
--- a/src/lingtai/kernel/llm_utils.py
+++ b/src/lingtai/kernel/llm_utils.py
@@ -50,7 +50,13 @@ def _send(
     retry_timeout: float,
     agent_name: str,
 ) -> LLMResponse:
-    """Send a message to the LLM. Single attempt with timeout."""
+    """Send a message to the LLM. Single attempt with timeout.
+
+    Note: on Python >= 3.11, ``except TimeoutError`` alone cannot distinguish a
+    wait-slice expiry from a worker-raised builtin ``TimeoutError`` (``socket
+    .timeout`` / ``asyncio.TimeoutError`` are aliases of it); ``future.done()``
+    is the discriminator and must stay part of this loop.
+    """
     future: Future = submit_fn()
     t0 = time.monotonic()
     while True:
@@ -63,6 +69,14 @@ def _send(
         try:
             return future.result(timeout=wait)
         except TimeoutError:
+            if future.done():
+                # The worker settled: either it raised its own TimeoutError
+                # (socket.timeout / asyncio.TimeoutError / concurrent.futures
+                # TimeoutError are all the builtin TimeoutError on >=3.11) or it
+                # finished during the race after our wait expired. Either way
+                # the worker is gone and its cleanup already ran — surface the
+                # real outcome instead of misreading it as "not responding".
+                return future.result(timeout=0)
             elapsed = time.monotonic() - t0
             if elapsed >= retry_timeout:
                 _wait_for_worker_settle(future, elapsed, agent_name)
@@ -84,10 +98,22 @@ def _wait_for_worker_settle(future: Future, elapsed: float, agent_name: str) -> 
     If the worker is still running after the grace period, raise a distinct
     WorkerStillRunningError. AED must not treat this as an ordinary timeout
     because the provider worker may still mutate the shared ChatInterface.
+
+    A worker that settled with its own builtin ``TimeoutError`` (``socket
+    .timeout`` / ``asyncio.TimeoutError`` / ``concurrent.futures.TimeoutError``
+    are all the same class on >= 3.11) is NOT still running: its except-block
+    already ran ``drop_trailing``. Use ``future.done()`` to tell the two apart
+    rather than ``except TimeoutError`` alone.
     """
     try:
         future.result(timeout=_WORKER_SETTLE_GRACE)
     except TimeoutError:
+        if future.done():
+            # The worker settled with its own TimeoutError (socket.timeout,
+            # asyncio.TimeoutError, and concurrent.futures.TimeoutError are all
+            # the builtin TimeoutError on >=3.11). Its except-block already ran
+            # drop_trailing — same as any other worker error.
+            return
         _logger.error(
             "[%s] LLM worker thread still running after %.0fs + %.0fs grace — "
             "interface state may be inconsistent. Refusing AED retry.",
--- a/tests/test_llm_utils.py
+++ b/tests/test_llm_utils.py
@@ -9,6 +9,7 @@ from lingtai.kernel.llm_utils import (
     track_llm_usage,
     execute_tools_batch,
     send_with_timeout,
+    _send,
     _SubmitFn,
     _wait_for_worker_settle,
 )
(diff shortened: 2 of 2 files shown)


huangzesen:
## Evidence: Kernel Windows PR contract run failure — not attributable to this PR

Run `31289999610` (job `93185431831`, step `Run Windows kernel contract tiers`) failed on the **unchanged** `tests/test_daemon_windows_supervisor.py::test_windows_adapter_spawns_real_detached_supervisor_reaching_capsule_read`: `AssertionError: condition not met within 60.0s`, summary `1 failed, 252 passed, 16 skipped` (log SHA cdc148d9...).

This PR (head `ec9100e6`, base `df2b523d`) changes only `src/lingtai/kernel/llm_utils.py` (+27/-1) and `tests/test_llm_utils.py` (+119/0); it does not touch the daemon/supervisor implementation or the failing test. The same unchanged test has recurred as a Windows runner failure in earlier PR triages (#1207, #1228, #1200/#1201, #1194). In the same window (02:18–02:24Z) 15+ sibling `Kernel Windows PR contract` runs on other dev4 branches **passed**, indicating the base is green and this is an intermittent unchanged-test failure rather than a PR-owned regression. Exact current-base control for this test is still absent.

Verdict: HOLD — not attributable to this PR; likely runner/intermittent flake of the unchanged test. No rerun performed.

huangzesen:

PR #1261, opened 2026-08-09
fix(prompt): make SystemPromptManager.set_order() functional (batch-preserving)

### LINKED ISSUE #746: SystemPromptManager.set_order() is a silent no-op — render path only reads class-level _BATCHES, never self._order
## Summary

`SystemPromptManager.set_order()` in `src/lingtai_kernel/prompt.py` is a silent no-op. Both `__init__` and `set_order()` write `self._order`, and the class docstring promises "Render order is configurable via set_order()", but `render()` delegates to `render_batches()`, which iterates only the class-level `_BATCHES` tuple and never consults `self._order`. A host that calls `set_order()` gets no effect and no error — the API lies.

## Narrative

Imagine a host embedding the kernel who wants their `rules` section rendered before `tools`, or who has added a custom section and wants it placed at a specific position rather than dumped into the anonymous "unordered" bucket. They read the class docstring at `src/lingtai_kernel/prompt.py:40-42`:

```
Render order is configurable via set_order(). Sections not in the order
list are rendered between the ordered sections and the tail. The last
name in the order list is always rendered last (typically 'context').
```

They call `mgr.set_order([...])`, re-render, and see the exact same output. No exception, no warning, no log line. The natural next step is to suspect their own code — maybe they mutated the wrong manager instance, maybe some cache is stale — and burn time debugging a bug that is actually in the kernel. The module-level docstring compounds the confusion by opening with "Sections are rendered in a configurable order" (`prompt.py:4`).

The code makes the failure mode obvious once you see all the pieces together. `set_order()` at `prompt.py:103-105` faithfully stores the list:

```python
def set_order(self, names: list[str]) -> None:
    """Set the render order. Last name is always rendered last."""
    self._order = list(names)
```

But `render_batches()` at `prompt.py:133-171` loops over the class attribute:

```python
for i, batch_names in enumerate(self._BATCHES):
    for name in batch_names:
        rendered = _render_entry(name)
        ...
```

`self._order` is written in exactly two places (`prompt.py:79`, `prompt.py:105`) and read in zero. A repo-wide grep of `src/` and `tests/` confirms no other module or test touches `set_order` or `_order`.

How did the design get here? The module clearly evolved. An earlier version rendered a flat, per-instance ordered list (`self._order`, seeded from `_DEFAULT_ORDER` at `prompt.py:56-75`), with a "last name renders last" tail convention — the docstring's mention of a `'context'` section (which no longer exists anywhere in the file) is a fossil of that era. Later, prompt-cache breakpoints were introduced: sections got grouped by mutation frequency into `_BATCHES` (`prompt.py:115-124`) so adapters can place `cache_control` markers at batch boundaries, and `render()` was rewritten as a thin join over `render_batches()` (`prompt.py:126-131`). The rewrite moved the source of truth from the instance's `_order` to the class's `_BATCHES`, but the old configurability API and its docstrings were left standing. The result is two parallel order definitions — `_DEFAULT_ORDER` and `_BATCHES` — that must be kept manually in sync (there is even a test guarding the sync: `test_default_order_and_batches_cover_the_same_sections_once` in `tests/test_prompt.py:41-45`), yet only one of them drives behavior.

What better looks like: instance-level batch state that `set_order()` actually reorders, with semantics that respect the cache-batching design (reordering must not silently merge batches, or it would destroy the stable-prefix caching that `_BATCHES` exists to provide). `_DEFAULT_ORDER` should be derived from `_BATCHES` structurally instead of duplicated by hand, and the fossil docstrings should be corrected.

## Current behavior

- `SystemPromptManager.__init__` (`src/lingtai_kernel/prompt.py:77-81`) initializes `self._order = list(self._DEFAULT_ORDER)`.
- `SystemPromptManager.set_order` (`prompt.py:103-105`) replaces `self._order`.
- `SystemPromptManager.render_batches` (`prompt.py:133-171`) iterates only the class attribute `_BATCHES`; `self._order` is never read anywhere in the codebase.
- `SystemPromptManager.render` (`prompt.py:126-131`) joins `render_batches()` output, so it inherits the same behavior.
- Net effect: calling `set_order()` changes nothing about rendering, and there is no error or warning. The class docstring (`prompt.py:40-42`) and module docstring (`prompt.py:4`) both claim configurable order; the class docstring additionally references a nonexistent `'context'` tail section.
- `_DEFAULT_ORDER` (`prompt.py:56-75`) and `_BATCHES` (`prompt.py:115-124`) are hand-maintained duplicates whose only consistency guarantee is a unit test.

The recommended fix wires order configurability into the batch-based render path while preserving cache-breakpoint semantics and current default output byte-for-byte.

1. **Make batches instance state.** In `SystemPromptManager.__init__` (`src/lingtai_kernel/prompt.py`), add:

   ```python
   self._batches: list[tuple[str, ...]] = [tuple(b) for b in self._BATCHES]
   ```

   Keep the class-level `_BATCHES` as the default template (external code and tests reference it).

2. **Point the render path at instance state.** In `render_batches()`, replace every use of `self._BATCHES` with `self._batches` — the loop at `prompt.py:154`, the list initialization at `prompt.py:143` (`[[] for _ in self._batches]`), and the `all_batched` set construction at `prompt.py:161`. No other logic changes; default behavior is identical because `self._batches` starts as a copy of `_BATCHES`.

3. **Reimplement `set_order()` with batch-preserving semantics.** Reordering must not collapse batch boundaries, or cache breakpoints lose their meaning. Semantics: each named section stays in the batch it currently belongs to, but sections within each batch are re-sorted by their position in `names`; names not currently in any batch are appended to the penultimate batch (matching the existing unordered-bucket rule at `prompt.py:160-169`); sections omitted from `names` keep their batch but sort after all named ones, preserving relative order.

   ```python
   def set_order(self, names: list[str]) -> None:
       """Reorder sections within their cache batches.

       Sections keep their batch (batch boundaries are cache breakpoints);
       within each batch, sections listed in `names` render in the given
       order, ahead of unlisted sections. Names not in any batch are
       appended to the penultimate batch.
       """
       index = {n: i for i, n in enumerate(names)}
       known = {n for batch in self._batches for n in batch}
       new_batches = [
           tuple(sorted(batch, key=lambda n: index.get(n, len(names))))
           for batch in self._batches
       ]
       extras = [n for n in names if n not in known]
       if extras:
           target = max(0, len(new_batches) - 2)
           new_batches[target] = new_batches[target] + tuple(extras)
       self._batches = new_batches
   ```

   Delete `self._order` entirely — it has no remaining purpose.

4. **Optionally add `set_batches()` for full control.** Hosts that genuinely need to move a section across batches (accepting the cache invalidation) get an explicit API:

   ```python
   def set_batches(self, batches: list[list[str]]) -> None:
       """Replace the cache-breakpoint batch layout entirely."""
       self._batches = [tuple(b) for b in batches]
   ```

5. **Derive `_DEFAULT_ORDER` from `_BATCHES`** so they cannot drift. Note this requires moving the `_BATCHES` definition above it (or defining `_DEFAULT_ORDER` after `_BATCHES` in the class body):

   ```python
   _DEFAULT_ORDER = [name for batch in _BATCHES for name in batch]
   ```

   Move the mutation-frequency commentary currently on `_DEFAULT_ORDER` (`prompt.py:45-55`) onto `_BATCHES`. Keep `_DEFAULT_ORDER` itself, since it is a useful flat view and is referenced by `tests/test_prompt.py:42`.

6. **Fix the docstrings.** Module docstring `prompt.py:4` and class docstring `prompt.py:40-42`: describe the batch-preserving `set_order()` semantics, remove the reference to the nonexistent `'context'` tail section, and state that unordered sections go to the penultimate batch (matching `render_batches()`'s actual documented behavior at `prompt.py:138-141`).

7. **Back-compat.** No in-tree caller invokes `set_order()` (only `src/lingtai_kernel/base_agent/__init__.py:394` constructs a `SystemPromptManager`, and it never reorders), so the signature can stay `set_order(names: list[str]) -> None` with new, actually-functional semantics. Default rendering is byte-identical, so existing prompt-cache behavior and the byte-identity tests (`test_batches_byte_identical_to_string`, `tests/test_prompt.py:231`) are unaffected.

## Risks and alternatives

- **Risk: silent semantic change for out-of-tree callers.** Any external host already calling `set_order()` today gets no-op behavior; after this change their call starts doing something. That is the point of the fix, but it should be called out in the changelog. Because reordering is constrained to within-batch, the blast radius is limited — no section can silently jump from the stable prefix batch into the volatile tail.
- **Risk: interaction with the unordered bucket.** `render_batches()` places sections not in any batch into the penultimate batch (`prompt.py:160-169`). Step 3 appends unknown names from `set_order()` into that same batch, keeping the two paths consistent; a test should pin this (see below).
- **Alternative A — delete the API.** Remove `set_order()`, `self._order`, and fix docstrings. Simplest and honest, and defensible given zero in-tree callers. Rejected as primary because the batching design genuinely benefits from a supported reorder/customize hook (hosts adding custom sections currently have no way to position them), and deleting a public method is a harder break than making it work.
- **Alternative B — flat-order rendering.** Make `set_order()` authoritative and derive batches by splitting the flat list at a recorded boundary. Rejected: boundary bookkeeping is fragile, and a flat list cannot unambiguously express multi-batch layouts; `set_batches()` covers that need explicitly.
- **Alternative C — `set_order()` collapses everything into one batch in the given order.** Maximally literal, but silently destroys cache breakpoints — exactly the kind of hidden behavior change this issue is about removing.

## Testing plan

Add to `tests/test_prompt.py`:

- `test_set_order_reorders_within_batch` — write `rules`, `brief`, `skills`; call `set_order(["skills", "rules", "brief"])`; assert `render()` contains `## skills` before `## rules` before `## brief`, and that batch membership is unchanged (`render_batches()[1]` contains all three, `render_batches()[0]` none).
- `test_set_order_does_not_move_sections_across_batches` — call `set_order(["pad", "principle"])` (names from different batches); assert `principle` still renders in batch 0 and `pad` in batch 1.
- `test_set_order_unknown_names_go_to_penultimate_batch` — include a custom name in `set_order()`, write that section, assert it renders in `render_batches()[0]` (the penultimate batch of two) in the requested position, matching where the unordered bucket would have put it.
- `test_set_order_omitted_sections_render_after_named_ones` — order only a subset; assert unlisted sections in the same batch render after the listed ones and keep their original relative order.
- `test_set_order_default_noop_is_byte_identical` — `set_order(list(SystemPromptManager._DEFAULT_ORDER))` produces byte-identical `render()` and `render_batches()` output to a fresh manager with the same sections.
- `test_set_batches_replaces_layout` (if step 4 is taken) — `set_batches([["rules"], ["tools"]])`; assert two batches with the given membership.
- Update `test_default_order_and_batches_cover_the_same_sections_once` (`tests/test_prompt.py:41`) — with `_DEFAULT_ORDER` derived from `_BATCHES` the set-equality assertion becomes structural; keep the test as a regression guard or tighten it to assert exact flattened equality.
- Confirm existing byte-identity and ordering tests (`test_batches_byte_identical_to_string`, `test_rules_renders_after_covenant_and_tools`, `test_meta_guidance_renders_between_procedures_and_comment`, `test_character_renders_after_identity_before_pad`) still pass unmodified, proving default behavior is unchanged.

huangzesen:
Fixes #746.

`SystemPromptManager.set_order()` was a silent no-op: both `__init__` and `set_order()` wrote `self._order`, but `render_batches()` only ever iterated the class-level `_BATCHES` tuple and never read `self._order` — while the module and class docstrings promised a configurable render order.

## What changed

- `SystemPromptManager` now keeps an instance-level batch layout (`self._batches`, initialized as a copy of the class `_BATCHES` template); `render_batches()` renders from it.
- `set_order()` is now functional with batch-preserving semantics: each section keeps its cache batch (batch boundaries remain cache breakpoints), listed names sort ahead of unlisted ones within each batch, and names not in any batch are appended to the penultimate batch — matching where `render_batches()` puts unordered sections. Default rendering is byte-identical.
- Added `set_batches()` as the explicit escape hatch for hosts that need to move a section across batches (accepting the resulting cache invalidation).
- `_DEFAULT_ORDER` is now derived structurally from `_BATCHES`, so the two can never drift again (the old hand-maintained duplicate is gone).
- Fixed module/class docstrings: removed the fossil `'context'` tail claim, documented the batch-preserving semantics, and corrected the stale Batch 2 listing in the module docstring (added `plugin`/`mcp`).

## Tests

Run with `PYTHONPATH=src pytest` (Python 3.14):

- `tests/test_prompt.py tests/test_prompt_catalog.py tests/test_prompt_section_definitions.py` → **155 passed**
- `tests/test_agent_meta_guidance.py` → **10 passed**
- New tests: `test_set_order_reorders_within_batch`, `test_set_order_does_not_move_sections_across_batches`, `test_set_order_unknown_names_go_to_penultimate_batch`, `test_set_order_omitted_sections_render_after_named_ones`, `test_set_order_default_noop_is_byte_identical`, `test_set_batches_replaces_layout`; the `_DEFAULT_ORDER`/`_BATCHES` sync test was tightened to exact flattened equality.
- Direct reproduction confirmed `set_order()` now changes rendered output (previously a no-op) and `set_order(list(SystemPromptManager._DEFAULT_ORDER))` stays byte-identical to a fresh manager.

### DIFF
--- a/src/lingtai/kernel/prompt.py
+++ b/src/lingtai/kernel/prompt.py
@@ -1,15 +1,16 @@
 """System prompt — section manager + builder.
 
 SystemPromptManager manages named sections of an agent's system prompt.
-Sections are rendered in a configurable order. The default order groups
-sections by mutation frequency so cache breakpoints can be placed between
-batches:
+Sections are rendered in a configurable order (set_order() reorders sections
+within their cache batches; set_batches() replaces the batch layout
+entirely). The default order groups sections by mutation frequency so cache
+breakpoints can be placed between batches:
 
     Batch 1 — resident prefix:
         principle (no header) → covenant → tools → substrate → procedures →
         meta_guidance → comment
     Batch 2 — rarely mutated (most stable first):
-        rules → brief → skills → knowledge → identity → character → pad
+        rules → brief → skills → plugin → mcp → knowledge → identity → character → pad
 
 `substrate` sits **right after tools** so it functions as the long-form
 companion to the schemas above it: tool schemas carry mechanical
@@ -37,53 +38,57 @@ class SystemPromptManager:
     Sections can be marked as protected (host-written, not overwritable by the LLM)
     or unprotected (LLM-writable at runtime).
 
-    Render order is configurable via set_order(). Sections not in the order
-    list are rendered between the ordered sections and the tail. The last
-    name in the order list is always rendered last (typically 'context').
+    Render order is configurable via set_order(), which reorders sections
+    within their cache-breakpoint batches: each section keeps the batch it
+    belongs to (batch boundaries are where adapters place cache_control
+    markers), while sections listed in `names` render ahead of unlisted ones
+    inside each batch. Names not in any batch are appended to the penultimate
+    batch — the same bucket where render_batches() places unordered sections —
+    never to the final tail batch. set_batches() replaces the batch layout
+    entirely for hosts that need to move sections across batches.
     """
 
-    # Default render order — grouped by mutation frequency. Sections in
+    # Cache-breakpoint batches — grouped by mutation frequency. Sections in
     # the same batch are adjacent so batch-boundary cache breakpoints in
     # the adapter can cover the whole stable prefix. Within each batch,
     # sections are ordered most-stable-first so later mutations invalidate
     # as little prior content as possible.
     #   Batch 1 (resident prefix):  principle, covenant, tools, substrate, procedures, meta_guidance, comment
-    #   Batch 2 (rarely-mutated):    rules, brief, skills, knowledge, identity, character, pad
+    #   Batch 2 (rarely-mutated):    rules, brief, skills, plugin, mcp, knowledge, identity, character, pad
+    # Resident kernel runtime guidance sits before operator/project comment
+    # so comment can remain the final stable prefix-layer instruction.
     # First entry (principle) is rendered without ## header (raw text).
     # `identity` is the mechanical section (name/nickname/manifest, written by
     # BaseAgent); `character` is the agent's self-authored identity from
     # system/lingtai.md (灵台) — distinct sections, character right after identity.
-    _DEFAULT_ORDER = [
-        # Batch 1 — immovable
-        "principle",
-        "covenant",
-        "tools",
-        "substrate",
-        "procedures",
-        # Resident kernel runtime guidance sits before operator/project comment
-        # so comment can remain the final stable prefix-layer instruction.
-        "meta_guidance",
-        "comment",
-        # Batch 2 — rarely mutated (most stable first)
-        "rules",
-        "brief",
-        "skills",
-        # Registered plugin packages and standalone MCP servers sit parallel to
-        # vanilla skills: each is a resident catalog field (written by the
-        # plugin/mcp capabilities, kept adjacent to skills so the agent sees the
-        # three capability namespaces side by side).
-        "plugin",
-        "mcp",
-        "knowledge",
-        "identity",
-        "character",
-        "pad",
-    ]
+    # Registered plugin packages and standalone MCP servers sit parallel to
+    # vanilla skills: each is a resident catalog field (written by the
+    # plugin/mcp capabilities, kept adjacent to skills so the agent sees the
+    # three capability namespaces side by side).
+    # Sections not listed here fall into the "unordered" bucket rendered just
+    # before the tail batch (see render_batches()).
+    _BATCHES: tuple[tuple[str, ...], ...] = (
+        (
+            "principle", "covenant", "tools", "substrate", "procedures",
+            "meta_guidance", "comment",
+        ),
+        (
+            "rules", "brief", "skills", "plugin", "mcp", "knowledge",
+            "identity", "character", "pad",
+        ),
+    )
+
+    # Flat view of the batch layout, derived structurally from _BATCHES so
+    # the two can never drift. Kept as a class attribute because tests and
+    # hosts reference it.
+    _DEFAULT_ORDER: list[str] = [name for batch in _BATCHES for name in batch]
 
     def __init__(self) -> None:
         self._sections: dict[str, dict] = {}
-        self._order: list[str] = list(self._DEFAULT_ORDER)
-        # First entry in order is rendered without ## header (raw text)
+        # Instance copy of the default batch layout; set_order()/set_batches()
+        # mutate this, never the class-level _BATCHES template.
+        self._batches: list[tuple[str, ...]] = [tuple(b) for b in self._BATCHES]
+        # First entry (principle) is rendered without ## header (raw text)
         self._raw_sections: set[str] = {"principle"}
 
     def write_section(self, name: str, content: str, protected: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


huangzesen:

PR #1263, opened 2026-08-09
fix(i18n): never raise on templates with literal braces when kwargs passed

### LINKED ISSUE #744: t() raises ValueError on templates containing literal braces when kwargs are passed
# t() raises ValueError on templates containing literal braces when kwargs are passed

## Summary

Both `t()` implementations (`src/lingtai_kernel/i18n/__init__.py` and `src/lingtai/i18n/__init__.py`) render templates with a bare `value.format_map(defaultdict(str, kwargs))` and no error handling. Several shipped templates contain literal, unescaped `{...}` (JSON-ish examples embedded in tool descriptions), so passing *any* kwarg to `t()` for those keys raises `ValueError: Invalid format specifier ...`. Today this is a landmine rather than a live crash only because the affected keys happen to be looked up without kwargs — any future kwarg, or any translator adding an unescaped `{`, turns a UI string lookup into an exception inside the agent loop.

## Narrative

The i18n layer is deliberately forgiving. Its docstring promises graceful degradation at every step: unknown language falls back to English, unknown key returns the key itself, and extra kwargs are silently ignored. That last promise is implemented with a `defaultdict(str)`:

```python
# src/lingtai_kernel/i18n/__init__.py:58-60
if kwargs:
    return value.format_map(defaultdict(str, kwargs))
return value
```

The `defaultdict` handles one failure mode — a *missing* placeholder — but `str.format_map` has a second, sharper failure mode: templates whose literal text happens to parse as a format field. The string tables have grown to include long tool descriptions that embed JSON-ish examples, and those examples contain raw braces. For example, `src/lingtai_kernel/i18n/en.json` key `soul.action_description` contains:

```
... invoking flow returns {status:'disabled', enabled:false} with an explanation ...
```

To `format_map`, `{status:'disabled', enabled:false}` is a replacement field named `status` with format spec `'disabled', enabled:false`. The `defaultdict` dutifully supplies `""` for `status`, and then formatting blows up on the bogus spec:

```
>>> t("en", "soul.action_description", foo="bar")
ValueError: Invalid format specifier ''disabled', enabled:false' for object of type 'str'
```

This reproduces today against the current tree. A quick audit that renders every shipped template with dummy kwargs finds **16 crashing templates**: 1 in `src/lingtai_kernel/i18n/en.json` (`soul.action_description`) and 5 in each of `src/lingtai/i18n/en.json`, `zh.json`, and `wen.json` (`knowledge.preamble`, `skills.preamble`, `daemon.tasks`, `daemon.tasks.backend_options`, `daemon.tasks.mcp`).

Nothing crashes in production *yet* because of a coincidence: the crashing keys are only ever looked up without kwargs. The single call site for the kernel-side key, `src/lingtai_kernel/intrinsics/soul/__init__.py:88`, is:

```python
"description": t(lang, "soul.action_description"),
```

and `t()` short-circuits past `format_map` entirely when `kwargs` is empty (`if kwargs:`). So the current safety of the system depends on an invisible, unenforced invariant — "keys whose templates contain literal braces must never be called with kwargs" — that is documented nowhere, checked nowhere, and violated by the very design intent of `t(lang, key, **kwargs)`. There are two easy ways to trip it:

1. A developer adds a kwarg to an existing call site (or adds a new call site with kwargs for one of the 16 keys). The code review looks completely innocuous; the crash only appears at runtime, in whatever code path builds tool schemas or preambles — i.e., inside the agent loop, where an unhandled `ValueError` from a *string lookup* is about the worst possible trade.
2. A translator edits `zh.json`/`wen.json` for a key that *is* called with kwargs (e.g. `system.current_time`, called as `t(lang, "system.current_time", time=..., ctx=...)`) and includes an unescaped `{` in the prose. Chinese translations of technical strings routinely keep English JSON examples verbatim, so this is likely, and nothing in CI would catch it.

The current design got here honestly: the string tables started as short interpolation templates (`"[Current time: {time} | context: {ctx}]"`) where `format_map` is exactly right, and later absorbed multi-paragraph tool documentation where braces are content, not syntax. The rendering function never grew to match. What better looks like: `t()` should keep its "never raise" contract end to end — if formatting fails, log a warning and return the raw template (a slightly odd string in a tool description is strictly better than a crash) — and CI should render every shipped template with dummy kwargs so brace bugs are caught at commit time, not in an agent's event loop.

## Current behavior

- `src/lingtai_kernel/i18n/__init__.py`, function `t()` (lines 44–60): resolves the key with fallbacks, then `return value.format_map(defaultdict(str, kwargs))` with no `try/except` when kwargs are present.
- `src/lingtai/i18n/__init__.py`, function `t()` (lines 57–66): identical pattern (duplicated implementation).
- `t("en", "soul.action_description", foo="bar")` raises `ValueError: Invalid format specifier ''disabled', enabled:false' for object of type 'str'`.
- Same class of failure for `knowledge.preamble`, `skills.preamble`, `daemon.tasks`, `daemon.tasks.backend_options`, `daemon.tasks.mcp` in all three `src/lingtai/i18n/*.json` tables (errors include `Invalid format specifier` and `unmatched '{' in format spec`).
- `tests/test_i18n.py` exercises substitution and fallback but never renders a brace-bearing template with kwargs, so nothing catches this.

1. **Harden kernel `t()`** — `src/lingtai_kernel/i18n/__init__.py`. Extract the formatting into a small helper and wrap it:

   ```python
   import logging
   log = logging.getLogger(__name__)

   def _render(value: str, kwargs: dict) -> str:
       try:
           return value.format_map(defaultdict(str, kwargs))
       except (ValueError, KeyError, IndexError) as e:
           log.warning("i18n: template render failed (%s); returning raw template", e)
           return value
   ```

   In `t()`, replace `return value.format_map(defaultdict(str, kwargs))` with `return _render(value, kwargs)`. Include the key in the warning message (pass `key` into `_render` or log at the call site) so a translator can find the offending template. `KeyError`/`IndexError` are included for positional fields like `{0}` and attribute/index access like `{a.b}` / `{a[0]}`, which `defaultdict` does not cover.

2. **Harden capability `t()`** — `src/lingtai/i18n/__init__.py`. Apply the identical change. Simplest is to duplicate `_render` (the two modules are intentionally parallel); alternatively, have `lingtai.i18n` import `_render` from `lingtai_kernel.i18n` since it already imports `register_strings` from there — prefer the import to avoid a third divergence point.

3. **Optionally fix the source templates.** For the 16 known templates, either escape braces (`{{status:'disabled', enabled:false}}`) or leave them as-is once `t()` is safe. Recommendation: leave the JSON untouched in this change — escaping doubles braces in three languages, is easy to get wrong, and the raw-template fallback renders them correctly anyway. If a template later needs *both* literal braces and substitutions, that key's template must escape its literal braces; the exhaustive-render test below will confirm it renders.

4. **Add an exhaustive-render test** — `tests/test_i18n.py`. Iterate every `*.json` under both `src/lingtai_kernel/i18n/` and `src/lingtai/i18n/`, render every key through the *public* `t()` with a dummy kwarg, and assert a `str` comes back with no exception:

   ```python
   def test_every_shipped_template_renders_with_dummy_kwargs():
       import json
       from pathlib import Path
       import lingtai_kernel.i18n as ki18n
       import lingtai.i18n as li18n
       for mod in (ki18n, li18n):
           for path in Path(mod.__file__).parent.glob("*.json"):
               table = json.loads(path.read_text(encoding="utf-8"))
               for key in table:
                   result = mod.t(path.stem, key, dummy="x")
                   assert isinstance(result, str)
   ```

   This test fails on the current tree (16 templates) and passes after step 1–2, and it permanently guards against translators introducing unescaped braces.

5. **Back-compat**: none needed. The function signature, key-resolution order, and all currently-succeeding renders are unchanged; only the previously-raising paths now return the raw template. No schema or data migration.

## Risks and alternatives

- **Silent corruption risk**: swallowing the exception means a template with a genuinely broken placeholder (e.g. typo `{tim}` for `{time}`) renders with an empty string or raw braces instead of failing loudly. Mitigations: the logged warning names the key, and the exhaustive-render test plus existing substitution tests (`test_stuck_revive_uses_err_desc_placeholder` already asserts placeholders are consumed) catch template/call-site drift in CI. For UI strings, "slightly wrong text plus a warning" beats "ValueError in the agent loop" every time.
- **Alternative: escape all literal braces in the JSON files.** Fixes the data but not the class of bug — the next translated string or new key reintroduces it, and doubled braces in three languages are hard to review. Worth doing selectively (step 3) but insufficient alone.
- **Alternative: switch to a non-colliding placeholder syntax** (e.g. `string.Template` with `$var`, or a custom `%%var%%`). Safest long-term but requires rewriting every template in six JSON files across three languages plus any downstream tables registered via `register_strings()`, and breaks third-party language packs. Disproportionate for the problem.
- **Alternative: validate templates at load time in `_load()`.** Catches bad templates early but still needs a per-render guard for kwargs-dependent failures, and adds startup cost. The try/except at render time is strictly simpler and covers both.
- **Keeping two copies of `t()` in sync** is a pre-existing maintenance hazard this issue makes visible; sharing `_render` (step 2) reduces, not increases, divergence.

## Testing plan

- `test_every_shipped_template_renders_with_dummy_kwargs` (see plan step 4): renders all keys of all six shipped JSON tables through both public `t()` functions with a dummy kwarg; asserts a `str` returns and nothing raises. Red on current tree, green after the fix.
- `test_literal_brace_template_returns_raw_on_kwargs`: `t("en", "soul.action_description", foo="bar")` returns the raw template unchanged (still contains `{status:'disabled', enabled:false}`) instead of raising.
- `test_literal_brace_render_logs_warning`: with `caplog`, assert the failed render emits a warning that includes the key name.
- `test_no_kwargs_path_unchanged`: `t("en", "soul.action_description")` (no kwargs) returns the template verbatim, as today.
- `test_substitution_still_works_after_hardening`: `t("en", "system.current_time", time="T", ctx="CTX") == "[Current time: T | context: CTX]"` — the happy path is untouched.
- Mirror of the first two tests against `lingtai.i18n.t` using one of its crashing keys (e.g. `daemon.tasks`) to cover the second implementation.

huangzesen:
Fixes #744

## Problem

`t()` renders templates with a bare `value.format_map(defaultdict(str, kwargs))` and no error handling. Templates are translator-authored prose that may embed literal, unescaped `{...}` (e.g. JSON-ish examples in tool descriptions). Passing *any* kwarg for such a key makes `str.format_map` mis-parse the literal braces as a format field and raises `ValueError: Invalid format specifier ...` (or `IndexError`/`AttributeError` for positional / attribute-access fields) — a crash inside the agent loop for what should be a pure string lookup.

## Current state on main

The two `t()` implementations cited in the issue were consolidated during the namespace restructure: `src/lingtai/tools/i18n/` now only *registers* strings into the kernel cache (no `format_map`), and the remaining kernel `t()` at `src/lingtai/kernel/i18n/__init__.py:76` still has the unguarded `format_map`. The 16 previously-crashing shipped templates no longer exist (tool descriptions moved to package glossaries), so nothing crashes today — but the landmine is fully live: any template with literal braces or positional fields + kwargs still raises, and nothing in CI would catch a translator adding an unescaped `{`.

Reproduced on current main (`df2b523d`):

```
>>> t('en', 'test.literal_brace', foo='bar')   # template: returns {status:'disabled', enabled:false} on failure
ValueError: Invalid format specifier ''disabled', enabled:false' for object of type 'str'
>>> t('en', 'test.positional', foo='bar')       # template: first {0} second {1}
ValueError: Format string contains positional fields
>>> t('en', 'test.attr_access', a='x')          # template: value {a.b}
AttributeError: 'str' object has no attribute 'b'
```

## Fix

- `src/lingtai/kernel/i18n/__init__.py`: extract rendering into `_render()` which catches `(ValueError, KeyError, IndexError, AttributeError)`, logs a warning naming the key, and returns the raw template unchanged. `t()` keeps its "never raise" contract end to end: unknown language → English fallback, unknown key → key itself, bad template → raw template + warning.
- `tests/test_i18n.py`:
  - `TestLiteralBraceTemplates`: literal-brace, positional, and attribute-access templates return the raw template instead of raising; no-kwargs path unchanged; failed render logs a warning containing the key; happy-path substitution still works.
  - `TestEveryShippedTemplateRenders.test_every_shipped_template_renders_with_dummy_kwargs`: renders every key of every shipped `*.json` catalog (kernel + tools, en/zh/wen) through the public `t()` with a dummy kwarg — permanent guard against translators introducing unescaped braces.
  - Importing `lingtai.tools.i18n` in the test module also fixes 3 pre-existing failures in `TestOperationalFallbackToEnglish` that occurred when `test_i18n.py` ran standalone (tools catalog was never registered).

## Tests

```
$ python3 -m pytest tests/test_i18n.py -q
22 passed in 0.33s
```

Broader i18n-touching suite (97 tests across 14 files): `97 passed, 2 failed` — the 2 failures (`test_docs_governance.py`) are pre-existing on clean main (verified by stashing the change: they fail identically without this fix).

No schema or data migration; function signature and all currently-succeeding renders are unchanged.

### DIFF
--- a/src/lingtai/kernel/i18n/__init__.py
+++ b/src/lingtai/kernel/i18n/__init__.py
@@ -13,7 +13,11 @@ key returns the key itself.
 from __future__ import annotations
 
 import json
+import logging
+from collections import defaultdict
 from pathlib import Path
+
+log = logging.getLogger(__name__)
 
 _DIR = Path(__file__).parent
 _CACHE: dict[str, dict[str, str]] = {}
@@ -58,14 +62,33 @@ def register_strings(lang: str, strings: dict[str, str]) -> None:
     table.update(strings)
 
 
+def _render(value: str, kwargs: dict, key: str) -> str:
+    """Render *value* with *kwargs*, falling back to the raw template.
+
+    Templates are written by translators and can contain literal braces (e.g.
+    JSON-ish examples embedded in prose) that ``str.format_map`` would
+    mis-parse as format fields. Rendering must never raise: a slightly odd
+    string in a tool description is strictly better than a ValueError inside
+    the agent loop. On failure, log a warning naming the key so the offending
+    template can be found, and return the raw template unchanged.
+    """
+    try:
+        return value.format_map(defaultdict(str, kwargs))
+    except (ValueError, KeyError, IndexError, AttributeError) as e:
+        log.warning(
+            "i18n: template render failed for key %r (%s); returning raw template",
+            key,
+            e,
+        )
+        return value
+
+
 def t(lang: str, key: str, **kwargs) -> str:
     """Translate a key. Falls back to English, then returns the key itself.
 
     Extra kwargs not referenced in the template are silently ignored
     (needed because en/zh templates may use different subsets of kwargs).
     """
-    from collections import defaultdict
-
     table = _load(lang)
     value = table.get(key)
     if value is None and lang != "en":
@@ -73,5 +96,5 @@ def t(lang: str, key: str, **kwargs) -> str:
     if value is None:
         return key
     if kwargs:
-        return value.format_map(defaultdict(str, kwargs))
+        return _render(value, kwargs, key)
     return value
--- a/tests/test_i18n.py
+++ b/tests/test_i18n.py
@@ -1,5 +1,14 @@
 """Tests for lingtai.kernel.i18n."""
-from lingtai.kernel.i18n import t
+import json
+import logging
+from pathlib import Path
+
+import lingtai.tools.i18n  # noqa: F401  (registers the tool catalog into the kernel cache)
+from lingtai.kernel.i18n import register_strings, t
+
+
+_KERNEL_I18N_DIR = Path(__file__).resolve().parents[1] / "src" / "lingtai" / "kernel" / "i18n"
+_TOOLS_I18N_DIR = Path(__file__).resolve().parents[1] / "src" / "lingtai" / "tools" / "i18n"
 
 
 class TestT:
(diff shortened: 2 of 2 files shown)


huangzesen:
Superseded by #1333. Closing as superseded.

huangzesen:
Superseded by #1333. Closing as superseded.

PR #1265, opened 2026-08-09
fix(mcp): give HTTPMCPClient stale-session recovery and blank-error formatting (#740)

### LINKED ISSUE #740: HTTPMCPClient lacks the stale-session recovery and blank-error formatting that MCPClient got for issues #104/#34
# HTTPMCPClient lacks the stale-session recovery and blank-error formatting that MCPClient got for issues #104/#34

## Summary

`MCPClient.call_tool` (stdio transport) wraps each tool call in a try/except that formats blank-`str()` exceptions via `_format_exception`, detects dead transports via `_is_stale_resource_error`, and recovers with a `restart()` + one retry. `HTTPMCPClient.call_tool` (`src/lingtai/services/mcp.py:412-456`) got none of this: it calls `future.result(timeout=timeout)` bare, the class has no `restart()`, and `is_connected()` cannot even detect a dropped HTTP/SSE stream. One dropped remote stream therefore poisons the client permanently — every subsequent call raises a raw (often message-less) exception until something re-creates the client object.

## Narrative

Issues #104 and #34 taught this codebase a painful lesson about the stdio MCP bridge: when an MCP transport dies underneath a live `ClientSession`, the next `call_tool` raises anyio's `ClosedResourceError` — an exception whose `str()` is *empty*. Before the fix, that produced the infamous blank `{"status": "error", "message": ""}` in agent tool results, and because nothing ever tore the session down, the error repeated forever. The fix landed in `MCPClient` as three cooperating pieces in `src/lingtai/services/mcp.py`: `_format_exception` (lines 61–72, renders `ClassName: message`, falling back to the class name when the message is empty), `_is_stale_resource_error` (lines 74–88, matches `ClosedResourceError` by class name plus closed/broken-transport substrings), and `restart()` (lines 117–138, which closes and fully resets the latched startup state — `_ready`, `_error`, `_closed`, `_session`, `_loop`, `_thread`, the context managers — so a subsequent `start()` is a real reconnect rather than a lie). `call_tool` ties them together (lines 209–243): attempt, and on a stale error, restart and retry exactly once, always returning a non-blank `{"status": "error", ...}` dict rather than raising.

`HTTPMCPClient`, the sibling class for remote streamable-HTTP servers, never received this treatment. Its `call_tool` is the pre-#104 shape:

```python
# src/lingtai/services/mcp.py:443-444
future = asyncio.run_coroutine_threadsafe(_call(), self._loop)
result = future.result(timeout=timeout)
```

No try/except, no `_format_exception`, no restart. The only #104 artifact it received is a borrowed `MCPClient._format_exception` in `_run_loop` (line 495) — and that covers *connect-time* failures only, not tool calls. The module's own documentation acknowledges the gap explicitly: `src/lingtai/services/ANATOMY.md` (line 66) ends its stale-recovery paragraph with "`HTTPMCPClient` ... has no stale-resource restart (stdio is the reported transport)." That was a scoping decision at fix time, not a claim that HTTP streams cannot die.

But HTTP/SSE streams die *more* readily than stdio pipes, not less. The remote server restarts, a proxy or NAT idles out the long-lived SSE connection, the laptop sleeps and wakes, the network blips. `HTTPMCPClient` instances are long-lived in three places: `Agent.connect_mcp_http` (`src/lingtai/agent.py:910-957`) registers per-tool handlers that close over one client for the life of the agent; `ZhipuSearchService` (`src/lingtai/services/websearch/zhipu.py:42`) keeps a persistent client across searches; and the daemon's task-scoped registrations (`src/lingtai/core/daemon/__init__.py:1583-1616`). When the stream drops, here is what a user or agent experiences today: the next tool call raises a raw `ClosedResourceError` out of the handler. Because `str()` of that exception is empty, whatever catches it upstream logs or displays an essentially blank error — the exact failure mode issue #104 fixed for stdio. Worse, the failure is sticky in a way the lazy-start path cannot heal: `HTTPMCPClient.is_connected()` (lines 404–410) checks only `_session is not None and self._loop.is_running() and not self._closed`. A dropped stream leaves the event loop running and `_session` set, so `is_connected()` keeps returning `True`, the lazy `self.start()` in `call_tool` (line 419) never fires, and every subsequent call fails identically. Even `zhipu.py`'s defensive `_ensure_client()` (lines 31–46), which checks `is_connected()` before reusing the client, is defeated by this — it happily reuses a zombie. Nothing recovers until a full agent refresh (or service `close()`) discards the client object.

There is a second, quieter casualty: observability. `MCPClient.call_tool` converts every failure into a result dict *before* appending to `_activity_log`, so failed calls are visible in `get_activity_log()`. In `HTTPMCPClient.call_tool` the exception propagates at line 444, before the activity-log append at lines 446–454 — failed HTTP tool calls leave no trace in the client's own debugging log.

Better looks like symmetry: both bridges already advertise "the same synchronous call_tool() interface" (module docstring, lines 3–7). The `_attempt` / classify / `restart()` / retry-once pattern is transport-agnostic — only the fields reset by `restart()` differ (`_transport_cm` vs `_stdio_cm`, plus HTTP's `_read_stream`/`_write_stream` set in `_async_connect`, lines 504–518). The recovery logic should be written once and shared, so the two classes cannot drift apart again.

## Current behavior

- `MCPClient.call_tool` (`src/lingtai/services/mcp.py:153-256`): wraps `_attempt()` in try/except; non-stale errors return `{"status": "error", "message": _format_exception(exc)}` (never blank); stale errors (per `_is_stale_resource_error`, lines 74–88) trigger `restart()` (lines 117–138) and one retry; all outcomes, including failures, are recorded in `_activity_log`.
- `HTTPMCPClient.call_tool` (`src/lingtai/services/mcp.py:412-456`): calls `future.result(timeout=timeout)` with no exception handling. Any transport failure propagates as a raw exception — for anyio's `ClosedResourceError`, one with an empty message.
- `HTTPMCPClient` has no `restart()` method. Its `start()` (lines 386–393) early-returns when `is_connected()` is true and never clears latched `_ready`/`_error`/`_closed` state, so even a caller that wanted to recover manually cannot do it correctly without reimplementing `restart()`'s field-reset dance.
- `HTTPMCPClient.is_connected()` (lines 404–410) returns `True` for a session whose underlying stream is closed, so the lazy-start branch in `call_tool` (lines 418–419) and `ZhipuSearchService._ensure_client` (`src/lingtai/services/websearch/zhipu.py:31-46`) both fail to detect the dead client.
- Failed HTTP tool calls never reach `_activity_log` (append at lines 446–454 is only reached on success).
- Tests in `tests/test_mcp_closed_resource_restart.py` cover the recovery pattern for `MCPClient` only.

1. **Extract the shared recovery machinery in `src/lingtai/services/mcp.py`.** Move `_format_exception` and `_is_stale_resource_error` to module-level functions (`_format_exception(exc)`, `_is_stale_resource_error(exc)`). Keep the existing `MCPClient._format_exception` / `MCPClient._is_stale_resource_error` staticmethods as thin aliases delegating to the module functions — `HTTPMCPClient._run_loop` (line 495) and the existing tests reference them via the class, so no call sites break.

2. **Add a shared recovery mixin** (or small base class) used by both clients, containing the classify/restart/retry logic currently inlined in `MCPClient.call_tool` (lines 209–243):

   ```python
   class _MCPRecoveryMixin:
       _client_kind = "MCP"  # "HTTP MCP" in HTTPMCPClient

       def _call_with_recovery(self, name: str, attempt) -> dict:
           try:
               return attempt()
           except Exception as exc:
               formatted = _format_exception(exc)
               if not _is_stale_resource_error(exc):
                   return {"status": "error", "message": formatted}
               logger.warning(
                   "%s tool %s hit stale resource (%s); restarting and retrying once",
                   self._client_kind, name, formatted)
               try:
                   self.restart()
               except Exception as restart_exc:
                   return {"status": "error", "message":
                           f"{formatted}: session closed; restart failed: "
                           f"{_format_exception(restart_exc)}"}
               try:
                   return attempt()
               except Exception as retry_exc:
                   return {"status": "error", "message":
                           f"{_format_exception(retry_exc)}: session closed; "
                           "restarted once but retry failed"}
   ```

   Rewire `MCPClient.call_tool` to call `self._call_with_recovery(name, _attempt)`, preserving its current messages (keep the exact "MCP session closed" wording if churn in error strings matters to downstream matchers — grep for consumers first).

3. **Add `HTTPMCPClient.restart()`**, mirroring `MCPClient.restart()` (lines 117–138) with the HTTP field set:

   ```python
   def restart(self) -> None:
       self.close()
       self._ready.clear()
       self._error = None
       self._closed = False
       self._session = None
       self._read_stream = None
       self._write_stream = None
       self._loop = None
       self._thread = None
       self._transport_cm = None
       self._session_cm = None
       self.start()
   ```

   Note: `_read_stream`/`_write_stream` are currently created only in `_async_connect` (line 512), not in `__init__` (lines 365–384). Add them to `__init__` for symmetry so `restart()` never touches undefined attributes.

4. **Rewrite `HTTPMCPClient.call_tool`** to the `MCPClient` shape: hoist the existing `_call` coroutine into a local `_attempt()` closure ending in `future.result(timeout=timeout)`, then `result = self._call_with_recovery(name, _attempt)`, then the activity-log append. This also fixes the observability gap — failed calls now land in `_activity_log` because they are converted to result dicts before the append.

5. **Back-compat on raise-vs-return.** Today `HTTPMCPClient.call_tool` *raises* on transport failure; after this change it returns `{"status": "error", ...}` like `MCPClient`. Audit the three call sites: `agent.py` tool handlers (line 938–941) pass the dict through to the LLM — strictly better than an unhandled exception; `zhipu.py:59-70` already handles both the exception path and the `status == "error"` dict path, so it keeps working; the daemon task registrations wrap handlers similarly. Document the unified contract in both classes' docstrings.

6. **Update docs**: fix `src/lingtai/services/ANATOMY.md` line 66 (remove "it has no stale-resource restart") and the module docstring to state both clients share the recovery behavior.

7. **Optional hardening (separate commit):** consider making `restart()` also fire when `list_tools` hits a stale stream, and/or a `ping`-based `is_connected()` probe. Not required for parity; keep this issue scoped to `call_tool`.

## Risks and alternatives

- **Retry of non-idempotent tools.** `_is_stale_resource_error`'s substring markers ("closed", "stream", "transport") are broad, and HTTP error messages (httpx, SSE) mention "stream" more often than stdio errors do, so the HTTP client may restart+retry on errors where the original request actually reached the server. Mitigations: the retry is bounded at one, matches the accepted stdio behavior, and MCP tool calls in this codebase (web search, LICC) are effectively idempotent. If this proves noisy, tighten the marker list for the HTTP client — but start with parity.
- **Behavior change from raise to error-dict** (step 5). A caller that relied on catching exceptions from `HTTPMCPClient.call_tool` would now get a dict. All three in-repo call sites tolerate or prefer the dict; the module has always advertised interface parity with `MCPClient`, which returns dicts.
- **Blocking reconnect.** `restart()` → `start()` blocks up to 30 s on `_ready.wait`. Same bound as the stdio client and as initial connection; acceptable for a synchronous bridge.
- **Alternative: full common base class** for lifecycle + loop + activity log (the two classes are ~80% duplicated). Cleaner long-term but a much larger diff with more regression surface; the mixin delivers the fix with minimal churn and can be a stepping stone.
- **Alternative: copy-paste the pattern into `HTTPMCPClient`.** Rejected — divergent copies are exactly how this gap arose; shared code makes the next fix land in both clients automatically.
- **Alternative: recover at call sites** (e.g., make `zhipu.py` recreate the client on failure). Rejected — there are at least three call sites, and `agent.py`'s per-tool handlers close over the client instance, making external replacement awkward; the client owns its transport and should own its recovery.

## Testing plan

Extend `tests/test_mcp_closed_resource_restart.py` (its `ClosedResourceError` stand-in, `_FakeFuture`, and `_Loop` fakes are reusable), parametrizing over both clients where the fakes allow:

- `test_http_call_tool_restarts_and_retries_once_on_stale_error` — first `_attempt` raises `ClosedResourceError`; assert `restart()` was called exactly once and the retry's success dict is returned.
- `test_http_call_tool_non_stale_error_returns_formatted_dict_not_raise` — a `ValueError("boom")` attempt returns `{"status": "error", "message": "ValueError: boom"}` and does *not* restart or raise.
- `test_http_call_tool_empty_message_error_surfaces_class_name` — an exception with empty `str()` yields a non-empty message containing the class name (the #104 regression guard, now for HTTP).
- `test_http_call_tool_failed_retry_returns_helpful_error` — both attempts raise; assert the message mentions the retry failure class and that a restart happened, and nothing propagates.
- `test_http_restart_resets_startup_state_so_start_cannot_lie` — mirror of the existing `MCPClient` test: set `_closed=True`, a stale `_error`, and a set `_ready`; after `restart()` (with a monkeypatched `start`), assert all latched fields are cleared, including `_transport_cm`, `_read_stream`, `_write_stream`.
- `test_http_call_tool_failure_recorded_in_activity_log` — after a failed call, `get_activity_log()` contains an entry whose `result["status"] == "error"` (guards the observability fix in step 4).
- `test_recovery_logic_shared_not_duplicated` — assert both classes resolve `_call_with_recovery` to the same function object (guards against future drift).

Run the existing suite (`tests/test_mcp_closed_resource_restart.py`, `tests/test_mcp_capability.py`) to confirm the `MCPClient` refactor onto the mixin preserves current messages and behavior.

huangzesen:
Fixes #740.

## Problem
`HTTPMCPClient.call_tool` was the pre-#104 shape: a bare `future.result(timeout=timeout)` with no exception handling, no `restart()`, and an `is_connected()` that keeps returning `True` after a dropped HTTP/SSE stream. One dropped remote stream therefore poisoned a long-lived client: the next call raised a raw (often message-less) exception — the exact blank-error failure mode issue #104 fixed for stdio — and failed calls never reached `_activity_log`.

## Fix
- **`HTTPMCPClient.restart()`** — close + reset all latched startup/session fields (incl. `_client`/`_http_client`/`_transport_cm`/`_session_cm`) so the subsequent `start()` is a real reconnect, not an early-return lie.
- **`call_tool` rewritten with the stdio recovery shape**: non-stale failures return a non-blank `{"status": "error", "message": "ClassName[: msg]"}` dict instead of raising (issue #104 guard); stale/closed streams `restart()` the transport for a future call and return an ambiguous, non-retryable error dict — the submitted call is never replayed (HTTP non-retry is contractual, matching the stdio default `retry_policy="never"`); a timeout after submission is reported ambiguous without retry; all outcomes incl. failures are recorded in `_activity_log`.
- **`get_activity_log()`** added to `HTTPMCPClient` so failed calls are observable.
- Reuses `MCPClient._format_exception` / `_is_stale_resource_error` / `_ambiguous_call_error` (the module established borrowing pattern), so the #104 formatting/stale-detection logic stays single-sourced.
- Docs: `ANATOMY.md` stale-recovery and HTTP non-retry bullets updated; module docstring now states both clients recover a dropped transport.

`HTTPMCPClient.call_tool` keeps its signature (no `retry_policy` parameter), preserving the asserted HTTP non-retry contract.

## Tests
`tests/test_mcp_closed_resource_restart.py` — 7 new HTTP tests: stale error → restart once without replay + ambiguous error dict; non-stale error → formatted dict, no restart, no raise; empty-message error → class name surfaced; post-submission timeout → ambiguous, cancelled; restart failure → ambiguous envelope naming restart error; failure recorded in `get_activity_log()`; `restart()` resets startup/session state incl. HTTP fields.

All green:
- `tests/test_mcp_closed_resource_restart.py`: 30 passed
- `tests/test_mcp_capability.py` + `test_mcp_sdk_v2_contract.py` + `test_mcp_structured_result.py`: 122 passed
- `test_mcp_v2_adapter_metadata.py` + `test_mcp_client_close_kills_subprocess.py` + `test_mcp_config_helper.py`: 44 passed
- `test_telegram_task_card.py`: 28 passed
- `ruff check` on changed files: clean


huangzesen:

PR #1269, opened 2026-08-09
fix(refresh): single-flight refresh so concurrent requests cannot spawn duplicate watchers

huangzesen:
## Summary

Fixes Lingtai-AI/lingtai#624 (blocking): concurrent refresh requests for the same agent directory spawn duplicate relaunch watchers that lock each other and exhaust recovery.

`_perform_refresh` (kernel lifecycle) had no per-agent single-flight guard: a second refresh request while one was pending saw `.refresh.taken` already on disk and spawned its own watcher anyway (e.g. a heartbeat-detected `.refresh` racing a `system(action='refresh')` tool call, or the AED preset-fallback racing either). Sibling watchers then each launched `lingtai run <dir>` children that mutually rejected one another with `error: another lingtai agent is already running`, exhausting both 12-attempt retry loops and publishing a permanent refresh failure; the agent stayed offline until an external restart.

## Fix

Make `_perform_refresh` single-flight per agent process:

- A per-agent `threading.Lock` + `_refresh_started` claim, taken at the very start of `_perform_refresh` before any side effect (chat-history save, handshake rename, watcher spawn).
- A second refresh while one is in flight is coalesced: logged `refresh_skipped` with `reason=refresh_already_in_progress` and **no second watcher is spawned**.
- The slot is released on every failure exit (no launch command, missing `RefreshWatcherPort`, ack-write failure, spawn exception) so a failed refresh never wedges a live agent.

Rationale: refresh is terminal for the current process — the watcher relaunches a NEW `lingtai run` child and `_shutdown` is set so `.agent.lock` releases — so at most one refresh operation may ever run per process lifetime. All refresh paths (heartbeat, intrinsic tool call, AED preset fallback, worker-recovery) funnel through `_perform_refresh`, so one guard covers them all.

## Tests

New regression tests in `tests/test_perform_refresh_handshake.py`:

- `test_two_refresh_requests_coalesce_to_one_watcher` — a second `_perform_refresh` on the same agent logs `refresh_skipped(refresh_already_in_progress)` and spawns no second watcher.
- `test_concurrent_refresh_requests_spawn_exactly_one_watcher` — barrier-driven two-thread variant (heartbeat thread vs LLM worker thread); exactly one watcher spawn.

Verified the new tests **FAIL on unpatched main** (`assert 2 == 1` — two watchers spawned, deterministically reproducing #624) and **PASS with the fix**.

Full local runs (Python 3.13):

- `tests/test_perform_refresh_handshake.py`: 44 passed
- `tests/test_refresh_watcher_process.py tests/test_deep_refresh.py tests/test_lifecycle_daemon_shutdown.py`: 87 passed
- `tests/test_cli_worker_poison_recovery.py tests/test_system.py`: 39 passed
- `tests/test_aed_recovery.py tests/test_base_agent.py tests/test_cli.py`: 62 passed

## Follow-ups (not in this PR)

- Stable operation ID / initiator attribution in refresh-watcher events (issue item 3).
- Typed `refresh_in_progress`/`agent_stopping` tool result instead of the unstructured `RuntimeError: MCP client has been closed` on dispatch after stop (issue item 4).
- Explicit cancel/transfer/handoff disposition for child-daemon records during refresh (issue item 7).

### DIFF
--- a/src/lingtai/kernel/base_agent/lifecycle.py
+++ b/src/lingtai/kernel/base_agent/lifecycle.py
@@ -716,6 +716,29 @@ def _perform_refresh(
     ``_shutdown`` / ``_cancel_event`` ourselves so the watcher's second
     phase can complete.
     """
+    # Single-flight guard (lingtai#624): refresh is terminal for this
+    # process — the watcher relaunches a NEW `lingtai run <dir>` child and
+    # we set `_shutdown` so `.agent.lock` releases — so at most ONE refresh
+    # operation may ever run per process lifetime. Two near-concurrent
+    # refresh requests (heartbeat-detected `.refresh` racing a
+    # `system(action='refresh')` tool call, or the AED preset-fallback
+    # racing either) must coalesce into one watcher; otherwise sibling
+    # watchers each launch a `lingtai run <dir>` child that the other
+    # rejects with "another lingtai agent is already running", and both
+    # retry loops exhaust MAX_ATTEMPTS, failing recovery permanently.
+    guard = getattr(agent, "_refresh_singleflight_lock", None)
+    if guard is None:
+        guard = threading.Lock()
+        agent._refresh_singleflight_lock = guard
+    with guard:
+        if getattr(agent, "_refresh_started", False):
+            agent._log("refresh_skipped", reason="refresh_already_in_progress")
+            return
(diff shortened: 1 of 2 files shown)


huangzesen:

PR #1271, opened 2026-08-09
fix(whatsapp): dedup inbound webhook replays by stable wamid (#733)

### LINKED ISSUE #733: WhatsApp: ingest_webhook has no replay/dedup guard, so Meta webhook retries create duplicate inbox entries and agent wakes
# WhatsApp: `ingest_webhook` has no replay/dedup guard — Meta webhook retries create duplicate inbox entries and duplicate agent wakes

## Summary

`WhatsAppManager.ingest_webhook` stores every inbound `message` event under a fresh `uuid4()` directory and fires `on_inbound(..., wake=True)` unconditionally, keyed by nothing stable — even though every event carries a globally stable upstream `wamid` (`ev["message_id"]`). Meta re-delivers webhook payloads whenever it does not receive a timely HTTP 200, so a single user message can be landed in the inbox multiple times and wake the agent repeatedly. The WeChat manager already solves this exact failure mode with a persisted `_stable_key` / `_is_replay` / `_record_seen` guard (`inbox_seen.json`); WhatsApp needs the analogous guard in front of `_store_message` / `on_inbound`.

## Narrative

The WhatsApp channel addon receives inbound messages via a Meta Cloud API webhook. `WhatsAppWebhookServer` (`src/lingtai/mcp_servers/whatsapp/webhook_server.py`) verifies the `X-Hub-Signature-256` signature, resolves the account alias from `metadata.phone_number_id`, and hands the payload to the manager:

```python
# src/lingtai/mcp_servers/whatsapp/webhook_server.py, do_POST (~lines 132-133)
events = outer.manager.ingest_webhook(alias, payload)
self._write(HTTPStatus.OK, json.dumps({"status": "ok", "events": len(events)}), "application/json")
```

`ingest_webhook` then lands each `message` event and pings the host:

```python
# src/lingtai/mcp_servers/whatsapp/manager.py, ingest_webhook (~lines 375-388)
for ev in events:
    if ev.get("kind") == "message":
        wa_id = ev.get("wa_id") or "unknown"
        wamid = ev.get("message_id") or f"local-{uuid4()}"
        msg = {"id": self._compound(alias, wa_id, wamid), ...}
        self._store_message(alias, "inbox", msg)
        if self.on_inbound:
            ...
            self.on_inbound({..., "wake": True})
```

and `_store_message` mints a brand-new local identity on every call:

```python
# src/lingtai/mcp_servers/whatsapp/manager.py, _store_message (~lines 140-146)
d = self._account_dir(alias) / folder / str(uuid4())
d.mkdir(parents=True, exist_ok=True)
```

Nothing in this path ever asks "have I already seen this wamid?". That matters because Meta's webhook delivery is at-least-once: if the receiver is slow, temporarily down, the tunnel (ngrok/Cloudflare) hiccups, or the response is non-200, Meta retries the same notification — with the same `messages[].id` (wamid) — with backoff, potentially for hours. Retries are the *normal* recovery mechanism, not an edge case. Every retry that eventually reaches `ingest_webhook` produces:

1. a new `whatsapp/<alias>/inbox/<uuid4>/message.json` with identical content but a different local directory name, and
2. a fresh `on_inbound` LICC notification with `wake: True`, waking the agent again for a message it may have already read and replied to.

The user experience is bad on both ends. The human sends one WhatsApp message; the agent gets woken two or three times, sees "duplicate" unread entries in `action=check` / `action=read`, and may reply twice (the compound id `alias:wa_id:wamid` is identical across copies, so `mark_read` at least hits the same upstream wamid — but the local inbox still shows N copies, and `search`/`check` results are polluted). There is also a subtler amplification loop: if the agent host is busy and the webhook HTTP response is delayed past Meta's timeout, Meta retries — and the retry itself creates more inbox entries and more wakes, making the host busier.

The current design likely got here because the WhatsApp addon was scaffolded as a straightforward store-and-notify pipeline, and the retry behavior of Meta's webhooks only bites under real-world latency. The repository has already been through this exact lesson once: the WeChat manager grew a documented replay guard after a production incident where a stale `get_updates` cursor re-delivered a backlog. Its comment block reads:

```python
# src/lingtai/mcp_servers/wechat/manager.py (~lines 406-411)
# Replay guard. ... Those re-deliveries arrive with identical upstream ids but
# would otherwise be landed under brand-new local UUIDs and counted as
# fresh unread. Detect them by their stable upstream signature and skip
# the second landing entirely — no new inbox entry, no LICC wake.
```

WeChat implements this with `_stable_key()` (prefer upstream id, fall back to a content signature), `_is_replay()`, `_record_seen()` with a bounded FIFO (`SEEN_KEYS_MAX = 5000`), atomic persistence to `inbox_seen.json`, and a crash-safety ordering rule ("record the signature only after the inbox write has succeeded"). WhatsApp is actually the *easier* case: Meta assigns a globally stable `wamid` to every message and repeats it verbatim on retries, so the stable key is essentially free. Better looks like: `ingest_webhook` checks a persisted seen-set keyed on `(alias, wa_id, wamid)` before `_store_message`, skips both the store and the wake on a hit, and records the key only after the store succeeds — mirroring the WeChat semantics so both channel addons behave identically under upstream redelivery.

Note also that the webhook server is a `ThreadingHTTPServer`, so a retry can arrive *concurrently* with the original delivery; the guard must be lock-protected (the WeChat implementation already holds `self._lock` inside `_is_replay`/`_record_seen`).

## Current behavior

- `src/lingtai/mcp_servers/whatsapp/manager.py` — `WhatsAppManager.ingest_webhook` (lines ~375-388): for every `kind == "message"` event it unconditionally calls `_store_message(alias, "inbox", msg)` and `self.on_inbound({..., "wake": True})`. No seen-key check exists anywhere in the class; the only lock is `_contacts_lock`, used solely for contacts.
- `src/lingtai/mcp_servers/whatsapp/manager.py` — `_store_message` (lines ~140-146): every call creates a new `uuid4()` directory, so identical redelivered events become distinct inbox entries.
- `src/lingtai/mcp_servers/whatsapp/webhook.py` — `extract_events` already surfaces the stable upstream id as `ev["message_id"]` (from `msg.get("id")`, the wamid) plus `wa_id`, `timestamp`, `type`, `text`.
- `src/lingtai/mcp_servers/whatsapp/webhook_server.py` — `do_POST` returns 200 only after `ingest_webhook` completes; any slow/failed response triggers Meta redelivery of the same payload, which the manager re-lands.
- Contrast: `src/lingtai/mcp_servers/wechat/manager.py` — `_stable_key` (~line 947), `_is_replay` (~line 987), `_record_seen` (~line 992), `_save_seen` persisting `inbox_seen.json`, `SEEN_KEYS_MAX = 5000` (~line 56), state reload in `_load_state` (~lines 908-920), all exercised by `tests/test_wechat_inbound_replay.py`. There is no WhatsApp equivalent and no WhatsApp replay test.

All changes are in `src/lingtai/mcp_servers/whatsapp/manager.py` unless noted. Mirror the WeChat design (names, file format, eviction policy) so maintainers deal with one dedup idiom, not two.

1. **Add module-level constant** `SEEN_KEYS_MAX = 5000` with a comment mirroring the WeChat one (bounded FIFO window; Meta retry windows are far smaller than 5000 messages).

2. **Extend `WhatsAppManager.__init__`** with guard state and a lock, and load persisted state:
   ```python
   self._seen_keys: dict[str, str] = {}   # stable_key -> first stored_at (or first local dir name)
   self._seen_order: list[str] = []       # FIFO for eviction
   self._seen_lock = threading.Lock()
   self._load_seen()
   ```
   Persist per-manager (not per-account) at `self.root / "inbox_seen.json"`, or per-account at `self._account_dir(alias) / "inbox_seen.json"` — per-manager is simpler and matches WeChat's single-file approach; keys are already namespaced by alias (see step 3), so one file is safe.

3. **Add `_stable_key(alias, ev) -> str | None`** (staticmethod or method):
   ```python
   @staticmethod
   def _stable_key(alias: str, ev: dict[str, Any]) -> str | None:
       wamid = ev.get("message_id")
       if wamid:
           return f"{alias}|{ev.get('wa_id') or 'unknown'}|mid:{wamid}"
       return None  # or content fallback, see below
   ```
   The wamid is Meta-assigned and stable across retries, so it is the primary key. Optionally add a WeChat-style content fallback for events with no `message_id` — `f"{alias}|{wa_id}|content:{ev.get('timestamp')}:{sha256(text)[:16]}"` — but if omitted, `None` must mean "no dedup, always land" (never dedup on a `local-<uuid4>` placeholder, which would be a new key every time anyway; the important thing is to *compute the key from the event before* the `wamid = ev.get("message_id") or f"local-{uuid4()}"` fallback runs).

4. **Add `_is_replay(key)` and `_record_seen(key, local_id)`**, lock-protected, with FIFO eviction beyond `SEEN_KEYS_MAX`, plus `_save_seen()` / `_load_seen()` using atomic write (tmp file + `os.replace`, same as WeChat's `_atomic_write`; `_save_contacts` in this file already uses the tmp-then-replace pattern). `_load_seen` must tolerate a missing or corrupt file by degrading to an empty guard (log a warning, never crash `__init__`).

5. **Wire the guard into `ingest_webhook`** before storing/waking:
   ```python
   for ev in events:
       if ev.get("kind") != "message":
           continue
       stable_key = self._stable_key(alias, ev)
       if stable_key is not None and self._is_replay(stable_key):
           log.info("WhatsApp inbound replay suppressed: %s", stable_key)
           continue  # no new inbox entry, no wake
       ...existing store...
       if stable_key is not None:
           self._record_seen(stable_key, local_id)
       ...existing on_inbound...
   ```
   Ordering rule (copy the WeChat comment): record the seen key only **after** `_store_message` succeeds, so a crash mid-write cannot mark a message seen without landing it. Recording before `on_inbound` is correct: a crash between store and wake loses at most one wake for a message that *is* in the inbox, whereas recording after `on_inbound` could double-wake on a crash — matching WeChat, which records before its `_on_inbound` call. To store `local_id`, have `_store_message` return the directory name (it currently returns only the msg dict; either return a `(msg, dir_name)` tuple internally or add the dir name into the stored/returned dict as e.g. `local_id`).

6. **Record provenance in the stored message** (optional but matches WeChat): add `"stable_key": stable_key` to the `msg` dict written in `ingest_webhook`, so suppressed duplicates can be traced to their original landing.

7. **`status` events**: they also carry stable ids and are redelivered, but today they are not stored and do not wake, so no guard is needed; leave them out of the seen-set to avoid growing it with delivery receipts.

8. **Back-compat / migration**: none needed. Absence of `inbox_seen.json` means an empty guard; existing inbox entries are unaffected (worst case, a retry of a message landed before the upgrade creates one final duplicate). The file format should carry `{"version": 1, "order": [...], "keys": {...}}` like WeChat's for future evolution.

9. **Optional follow-up (separate concern)**: consider whether `webhook_server.py` should respond 200 *before* heavy ingest work to shrink the retry window. With the dedup guard in place this becomes a latency optimization rather than a correctness requirement, so it can be deferred.

## Risks and alternatives

- **False-positive suppression (dropping a real message)**: the wamid is unique per message per Meta, and the key is additionally namespaced by `alias` and `wa_id`, so two genuinely distinct messages cannot collide. The FIFO cap (5000) bounds memory/disk; Meta's retry horizon is far shorter than 5000 inbound messages for any realistic deployment.
- **Crash-ordering**: recording seen-before-store would drop messages on crash; the plan mandates store-then-record, accepting the benign failure mode (rare duplicate after a crash between store and record) over the dangerous one.
- **Concurrency**: `ThreadingHTTPServer` can run two `do_POST` handlers at once, so the original and a retry can race. The lock in `_is_replay`/`_record_seen` closes the window except for the instant between the check and the record; to fully close it, `_is_replay`+`_record_seen` could be fused into a single atomic `_claim(key) -> bool` under one lock acquisition (check-and-insert). Recommended: implement `_claim` and call it before `_store_message`, then treat "claimed but store failed" by removing the key in an `except` path — or accept WeChat's simpler check/record split for consistency. Either is acceptable; the issue prefers the fused `_claim` since the WhatsApp server is threaded while WeChat's poller is single-consumer.
- **Alternative: dedup by scanning existing inbox for the compound id** (`alias:wa_id:wamid` is already stored in `msg["id"]`). Rejected: O(n) directory scan per inbound message, unbounded as the inbox grows, and it cannot distinguish "seen and later archived/deleted" from "never seen".
- **Alternative: dedup in `webhook_server.py`** (hash the raw body). Rejected: retries can arrive in re-batched payloads (one entry retried inside a different envelope), and the manager can be driven by other callers/tests; the guard belongs at the ingest boundary where per-message stable ids are available.
- **Alternative: shared dedup helper extracted from WeChat**. Attractive long-term (a small `_inbox_seen.py` in `src/lingtai/mcp_servers/` with the key-store/FIFO/atomic-persist logic), but WeChat's `_stable_key` is protocol-specific; extracting only the store is a reasonable stretch goal, not a blocker.

## Testing plan

Add `tests/test_whatsapp_inbound_replay.py`, modeled on `tests/test_wechat_inbound_replay.py`:

- `test_duplicate_webhook_delivery_lands_once` — build a `WhatsAppManager` with a tmp `working_dir` and a recording `on_inbound`; call `ingest_webhook(alias, payload)` twice with an identical Meta-shaped payload (same `messages[].id`). Assert exactly one directory under `whatsapp/<alias>/inbox/`, exactly one `on_inbound` call, and that the second call still returns the extracted events list (return value unchanged).
- `test_distinct_wamids_both_land` — two payloads differing only in `messages[].id`; assert two inbox entries and two wakes.
- `test_same_text_different_message_ids_not_deduped` — same `wa_id` and identical `text` but different wamids must both land (guards against over-aggressive content keying).
- `test_seen_state_survives_restart` — ingest once, construct a *new* `WhatsAppManager` over the same `working_dir`, re-ingest the same payload; assert no new inbox entry and no wake (verifies `inbox_seen.json` persistence and reload).
- `test_corrupt_seen_file_degrades_gracefully` — write garbage to `inbox_seen.json`, construct the manager; assert no exception and that ingest still works (guard resets to empty).
- `test_missing_message_id_never_suppressed` — payload whose message lacks `id`; ingest twice; assert both land (or, if the content fallback is implemented, assert the documented fallback behavior instead).
- `test_seen_keys_fifo_eviction` — set/monkeypatch `SEEN_KEYS_MAX` small (e.g. 3), ingest 4 distinct messages, then replay the first; assert it lands again (evicted) while replaying the newest is still suppressed.
- `test_record_after_store_ordering` — monkeypatch `_store_message` to raise; assert the key was **not** recorded (a subsequent successful ingest of the same payload lands the message).
- Concurrency smoke test (optional): fire the same payload from two threads through `ingest_webhook`; assert exactly one inbox entry and one wake.
- Extend `tests` that exercise the webhook server end-to-end, if any are added later, to POST the same signed body twice and assert a single inbox landing.

huangzesen:
Fixes #733.

## Problem
Meta Cloud API webhook delivery is at-least-once. Whenever the receiver is slow, down, or answers non-200, Meta re-delivers the same notification - with the same stable `messages[].id` (wamid) - on backoff, potentially for hours. `WhatsAppManager.ingest_webhook` landed every delivery under a fresh local UUID and woke the agent each time, creating duplicate inbox entries and duplicate agent wakes (and a busy-host amplification loop). The WeChat manager already solves this with a persisted replay guard; WhatsApp had no equivalent and no replay test.

## Change
All in `src/lingtai/mcp_servers/whatsapp/manager.py`, mirroring the WeChat manager's `inbox_seen.json` idiom (one dedup pattern across channel addons):

- `SEEN_KEYS_MAX = 5000` bounded FIFO window constant.
- `_stable_key(alias, ev)` - stable key from the upstream wamid, namespaced by account alias and `wa_id`; returns `None` (no dedup) when the event carries no stable id, never keying on a local `local-<uuid4>` placeholder.
- Lock-protected `_is_replay` / `_record_seen` with FIFO eviction, `_save_seen` (atomic tempfile + `os.replace`) and `_load_seen` (missing/corrupt file degrades to an empty guard, never crashes boot).
- `ingest_webhook` checks the seen-set before storing/waking, skips replays entirely (no new inbox entry, no LICC wake), and records the key only after the inbox write succeeds - crash-safe ordering: a crash mid-write can never mark a message seen without landing it. Stored messages carry `stable_key` provenance for traceability.
- `_store_message` now returns `(msg, local_id)` so the guard records the local landing directory; `_send`/`_reply` unpack the tuple.

`status` events are not stored or woken today, so they stay out of the seen-set.

## Tests
New `tests/test_whatsapp_inbound_replay.py` (9 tests):

- duplicate webhook delivery lands once (1 inbox entry, 1 wake, return value unchanged)
- distinct wamids both land
- same text + different wamids are NOT deduped (no content-key false positives)
- seen state survives restart (inbox_seen.json reload)
- corrupt inbox_seen.json degrades gracefully
- missing message_id is never suppressed
- FIFO eviction at SEEN_KEYS_MAX
- store failure does NOT record the key (retry lands the message)
- stored message keeps stable_key provenance

Results:
- `pytest tests/test_whatsapp_inbound_replay.py` - 9 passed
- `pytest tests/test_whatsapp_notification_metadata.py tests/test_whatsapp_toolfamily_ltpv2.py tests/test_wechat_inbound_replay.py` - 55 passed
- `ruff check` on the new test file - clean (manager.py carries the same 12 pre-existing violations on base main; this PR adds none)

Note: this implements the check-then-record split from the issue's detailed plan (step 5), exactly mirroring the WeChat manager. The fused `_claim` alternative mentioned in the issue's risks section remains a possible follow-up; the split never drops a message, which is the crash-ordering invariant the issue prioritizes.

### DIFF
--- a/src/lingtai/mcp_servers/whatsapp/manager.py
+++ b/src/lingtai/mcp_servers/whatsapp/manager.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import re
+import tempfile
 import threading
 from datetime import datetime, timezone
 from importlib import resources
@@ -49,6 +51,12 @@ _CS_WINDOW_NOTE = (
 _CONVERSATION_CONTEXT_MESSAGES = 10
 _STRUCTURED_MESSAGE_TEXT_CAP = 500
 
+# Bounded FIFO window for the inbound replay guard (inbox_seen.json). Sized
+# well above a single Meta webhook retry horizon: Meta re-delivers the same
+# wamid with backoff, so a live retry never falls outside this window while it
+# could still be re-delivered. Bounds memory/disk for the persisted seen-set.
+SEEN_KEYS_MAX = 5000
+
 DESCRIPTION = "WhatsApp Cloud API client for LingTai. Official Meta API only; no WhatsApp Web bridge."
 
 # Public callers receive the strict LTP-v2 family schema. Manager dispatch
@@ -73,6 +81,15 @@ class WhatsAppManager:
         self.config_source = config_source
         self._last_verified_at = _utcnow()
         self._contacts_lock = threading.Lock()
+        # Inbound replay guard state (see _stable_key/_is_replay/_record_seen).
+        # Meta webhooks are at-least-once: on a slow or non-200 response the
+        # same payload (same wamid) is redelivered with backoff. Persisted to
+        # root/inbox_seen.json so the guard survives restarts; keys are
+        # namespaced by account alias, so one file covers all accounts.
+        self._seen_keys: dict[str, str] = {}
+        self._seen_order: list[str] = []
+        self._seen_lock = threading.Lock()
+        self._load_seen()
 
     def _account_alias(self, alias: str | None) -> str:
         if alias:
@@ -127,13 +144,19 @@ class WhatsAppManager:
             raise ValueError("message_id must be account:wa_id:wamid")
         return parts[0], parts[1], parts[2]
 
-    def _store_message(self, alias: str, folder: str, msg: dict[str, Any]) -> dict[str, Any]:
+    def _store_message(self, alias: str, folder: str, msg: dict[str, Any]) -> tuple[dict[str, Any], str]:
+        """Write ``msg`` to a fresh local inbox/sent directory.
+
+        Returns ``(stored_msg, local_id)`` where ``local_id`` is the local
+        directory name (the replay guard records it so a suppressed duplicate
+        can be traced to its original landing).
+        """
         d = self._account_dir(alias) / folder / str(uuid4())
         d.mkdir(parents=True, exist_ok=True)
         msg = dict(msg)
         msg.setdefault("stored_at", _utcnow())
         (d / "message.json").write_text(json.dumps(msg, ensure_ascii=False, indent=2), encoding="utf-8")
-        return msg
+        return msg, d.name
 
     def _iter_messages(self, alias: str, folder: str | None = None) -> list[dict[str, Any]]:
         base = self._account_dir(alias)
@@ -204,7 +227,7 @@ class WhatsAppManager:
         try:
             response = self._client(alias).post_message(payload)
             wamid = (((response.get("messages") or [{}])[0]).get("id") or f"local-{uuid4()}")
-            stored = self._store_message(alias, "sent", {"id": self._compound(alias, to, wamid), "wa_id": to, "message_id": wamid, "text": args.get("text"), "payload": payload, "response": response, "direction": "outgoing"})
+            stored, _ = self._store_message(alias, "sent", {"id": self._compound(alias, to, wamid), "wa_id": to, "message_id": wamid, "text": args.get("text"), "payload": payload, "response": response, "direction": "outgoing"})
             return {"status": "sent", "message_id": stored["id"], "response": response}
         except Exception as e:
             return {"status": "error", "error": str(e), "note": _CS_WINDOW_NOTE}
@@ -222,7 +245,7 @@ class WhatsAppManager:
         try:
             response = self._client(alias).post_message(payload)
             new_id = (((response.get("messages") or [{}])[0]).get("id") or f"local-{uuid4()}")
-            stored = self._store_message(alias, "sent", {"id": self._compound(alias, wa_id, new_id), "wa_id": wa_id, "message_id": new_id, "reply_to": wamid, "text": args.get("text"), "payload": payload, "response": response, "direction": "outgoing"})
+            stored, _ = self._store_message(alias, "sent", {"id": self._compound(alias, wa_id, new_id), "wa_id": wa_id, "message_id": new_id, "reply_to": wamid, "text": args.get("text"), "payload": payload, "response": response, "direction": "outgoing"})
             return {"status": "sent", "message_id": stored["id"], "response": response}
         except Exception as e:
             return {"status": "error", "error": str(e), "note": _CS_WINDOW_NOTE}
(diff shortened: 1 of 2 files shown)


huangzesen:
Superseded by #1332. Closing as superseded.

huangzesen:
Superseded by #1332. Closing as superseded.

PR #1272, opened 2026-08-09
fix: snapshot adapter cache scan in get_adapter fast path (#739)

### LINKED ISSUE #739: LLMService.get_adapter fast path iterates the adapter cache without the lock — races with concurrent adapter creation
## Summary

`LLMService.get_adapter` (`src/lingtai/llm/service.py:275-321`) has a lock-free "fast path" that iterates `self._adapters.items()` when `base_url is None`, while another thread holding `self._adapter_lock` can concurrently execute `self._adapters[cache_key] = adapter` (line 320). A concurrent insert during that iteration raises `RuntimeError: dictionary changed size during iteration`, which surfaces as a spurious LLM failure in daemon and capability code paths. Additionally, the `base_url=None` "any adapter for this provider" match is nondeterministic when multiple base_urls are cached for one provider.

## Narrative

`LLMService` keeps a per-service adapter cache so the same provider + base_url pair reuses one `LLMAdapter` instance instead of constructing a new HTTP client on every call. The cache is a plain dict keyed by `(provider, base_url)` and guarded by a `threading.Lock` for *writes* (`src/lingtai/llm/service.py:221-222`):

```python
self._adapters: dict[tuple[str, str | None], LLMAdapter] = {}
self._adapter_lock = threading.Lock()
```

`get_adapter` was written with the classic double-checked-locking shape: try a cheap lock-free read first, and only take the lock when the adapter needs to be created. For the exact-key hit that is fine — a single `cache_key in self._adapters` membership test followed by a `[]` lookup is atomic-enough in CPython. But the fast path does more than a point lookup. When the caller passes no `base_url`, it *scans the whole dict* (`src/lingtai/llm/service.py:290-294`):

```python
# When no base_url specified, find any cached adapter for this provider
if base_url is None:
    for (p, _url), adapter in self._adapters.items():
        if p == provider:
            return adapter
```

Iterating a dict is not an atomic operation. If any other thread inserts a new key while this loop is between steps, CPython raises `RuntimeError: dictionary changed size during iteration`. And there *is* a writer that can run concurrently: the slow path inside the lock does `self._adapters[cache_key] = adapter` at line 320. The lock protects writers from each other, but the fast-path reader deliberately does not take the lock, so nothing protects the scan from a concurrent insert.

How does someone actually hit this today? The daemon path is the clearest example. `src/lingtai/core/daemon/__init__.py:185` calls `service.get_adapter(provider)` — no `base_url` — from daemon-scoped work, while agent/capability code paths call `get_adapter` on the same shared `LLMService` from other threads. The first time a second provider (or a provider + explicit base_url combination) is requested, the slow path inserts a new cache entry. If a daemon thread happens to be in the middle of the fast-path scan at that moment, the scan blows up with a `RuntimeError` that has nothing to do with the LLM call itself. The window is small — adapter creation only inserts once per `(provider, base_url)` — which makes this the worst kind of bug: it essentially never reproduces in tests, and in production it shows up as a one-off, unexplainable "LLM failure" right around agent startup or the first use of a fallback provider, exactly when several threads race to warm the cache.

There is a second, quieter defect in the same lines. The `base_url=None` fast path returns *any* cached adapter whose provider matches, in dict insertion order. If two adapters are cached for the same provider with different base_urls (e.g. OpenRouter and a local vLLM endpoint — the exact scenario the cache key was designed for, per the docstring at lines 278-280), which one a `base_url=None` caller gets depends on which happened to be created first. That is nondeterministic across process runs and surprising: the constructor seeds the cache with the service's *configured* `(self._provider, base_url)` entry (line 223), and a `base_url=None` lookup for the primary provider should deterministically prefer that configured adapter, or at minimum apply a stable, documented rule.

The fix is small. Dict point reads can stay lock-free, but the scan must either operate on a snapshot (`list(self._adapters.items())` — the race disappears because the snapshot is taken atomically) or move under the lock. Snapshotting keeps the fast path lock-free and is a one-line change; the same loop inside the lock (lines 301-304) is already safe but should get the same deterministic-preference treatment. Better still, factor the provider-match scan into one helper so the fast path and the locked path cannot drift apart, and give it a deterministic rule: exact `(provider, None)` key first, then the entry matching the provider's configured default base_url, then first match in sorted-key order.

## Current behavior

- `src/lingtai/llm/service.py:288-289` — lock-free exact-key check: safe (atomic point read).
- `src/lingtai/llm/service.py:291-294` — lock-free iteration over `self._adapters.items()`: **unsafe**. A concurrent `self._adapters[cache_key] = adapter` at line 320 (executed under `self._adapter_lock` by another thread) during this iteration raises `RuntimeError: dictionary changed size during iteration`. The exception propagates to the caller as a failed LLM operation.
- `src/lingtai/llm/service.py:301-304` — the same scan repeated inside the lock: safe from the writer at line 320 (mutual exclusion), but shares the nondeterminism below.
- Both scans return the *first* adapter whose provider matches, in dict insertion order — nondeterministic from the caller's perspective when multiple base_urls are cached for one provider.
- Callers that hit the `base_url=None` scan path: `src/lingtai/core/daemon/__init__.py:185` and the `provider`-override branches in `src/lingtai/llm/service.py:381`, `:423`, `:451`.

1. **Add a private scan helper on `LLMService`** in `src/lingtai/llm/service.py`, so the fast path and locked path share one implementation and one deterministic rule:

   ```python
   def _find_provider_adapter(self, provider: str) -> LLMAdapter | None:
       """Deterministically pick a cached adapter for *provider* (any base_url).

       Preference order:
       1. the (provider, None) entry,
       2. the entry matching the provider's configured default base_url,
       3. first match by sorted cache key (stable tie-break).

       Snapshots the cache so it is safe to call without _adapter_lock.
       """
       items = list(self._adapters.items())  # atomic snapshot; safe vs concurrent insert
       matches = {url: a for (p, url), a in items if p == provider}
       if not matches:
           return None
       if None in matches:
           return matches[None]
       defaults = self._get_provider_defaults(provider)
       default_url = defaults.get("base_url") if defaults else None
       if default_url in matches:
           return matches[default_url]
       return matches[min(k for k in matches if k is not None)]
   ```

   The key correctness point is `list(self._adapters.items())`: materializing the items is a single C-level operation in CPython, so a concurrent insert can no longer interleave with the iteration. (On free-threaded builds the per-object dict lock makes this snapshot safe as well.)

2. **Rewrite the fast path in `get_adapter`** (replace lines 287-294):

   ```python
   # Fast path — atomic point read, then snapshot-based scan; no lock needed
   adapter = self._adapters.get(cache_key)
   if adapter is not None:
       return adapter
   if base_url is None:
       adapter = self._find_provider_adapter(provider)
       if adapter is not None:
           return adapter
   ```

   Note the switch from `in` + `[]` to a single `.get()` — the two-step form is itself a (much smaller) TOCTOU pattern; one atomic read is strictly better.

3. **Rewrite the locked double-check** (replace lines 298-304) to reuse the same helper:

   ```python
   with self._adapter_lock:
       adapter = self._adapters.get(cache_key)
       if adapter is not None:
           return adapter
       if base_url is None:
           adapter = self._find_provider_adapter(provider)
           if adapter is not None:
               return adapter
       ...
   ```

4. **Document the deterministic rule** in the `get_adapter` docstring (lines 276-283): when `base_url is None` and multiple adapters are cached for the provider, the `(provider, None)` entry wins, then the provider's configured default base_url, then sorted-key order. Since the constructor (line 223) seeds `(self._provider, base_url)`, callers of `get_adapter(self._provider)` on a service configured with an explicit base_url will now deterministically get that configured adapter via rule 2 — previously they got it only by insertion-order accident.

5. **No schema, data-structure, or back-compat changes.** The cache stays `dict[tuple[str, str | None], LLMAdapter]`; the public signature of `get_adapter` is unchanged. Behavior changes only in (a) never raising the spurious `RuntimeError`, and (b) which adapter a `base_url=None` call returns when several are cached — previously unspecified, so no documented contract is broken.

## Risks and alternatives

- **Alternative: take the lock in the fast path too.** Simplest to reason about, but it serializes every `get_adapter` call across all sessions of a service, including the overwhelmingly common cache-hit case; `get_adapter` sits on the hot path of every completion (`create_session`, lines 348/381/423/451). The snapshot approach keeps reads lock-free with a one-line change, which is why it wins.
- **Alternative: copy-on-write dict** (writers build a new dict under the lock and atomically rebind `self._adapters`). Also correct and lock-free for readers, but more machinery than needed for a dict that gains at most a handful of entries per process lifetime.
- **Alternative: rely on the GIL and call it a non-issue.** The GIL does not make dict *iteration* atomic — the interpreter can switch threads between loop steps, and `RuntimeError: dictionary changed size during iteration` is precisely the guard CPython raises for this. It also ages badly under free-threaded (PEP 703) Python.
- **Risk: the deterministic preference changes which adapter some callers receive.** Any caller that previously depended on insertion order was depending on a race; still, review call sites (`src/lingtai/core/daemon/__init__.py:185` and the `provider`-override branches in `service.py`) to confirm the new preference (explicit `None` entry, then configured default base_url) matches intent. For the seeded primary-provider entry it does.
- **Risk: snapshot staleness.** The fast path may miss an entry inserted a microsecond earlier and fall through to the lock, where the double-check finds it. That is the normal double-checked-locking outcome — correct, merely occasionally slower — and can also return an adapter created concurrently for the same provider, which is exactly what the old code intended.

## Testing plan

Add `tests/test_llm_service_adapter_cache.py` (or extend the existing service tests using helpers from `tests/_service_helpers.py`):

1. **`test_get_adapter_scan_race_with_concurrent_insert`** — regression test for the crash. Build an `LLMService` with a stub adapter registry and a key resolver that returns a dummy key. Pre-seed `service._adapters` with many entries for provider `"a"` (forcing a long scan), then run two thread groups under a `threading.Barrier`: readers looping `service.get_adapter("a")` (no base_url) and writers repeatedly inserting/creating adapters for fresh `(provider, base_url)` keys (e.g. via `get_adapter("b", f"http://x/{i}")` with a factory that yields new objects). Assert no thread raises `RuntimeError`. On the unfixed code with a scan over a few thousand entries and aggressive writers, this fails within a few hundred iterations; keep the loop count high enough to be a meaningful canary but bounded (<2s).
2. **`test_get_adapter_prefers_none_key_entry`** — cache `("p", None)` and `("p", "http://b")`; assert `get_adapter("p")` returns the `None`-keyed adapter regardless of insertion order (parametrize both orders).
3. **`test_get_adapter_prefers_configured_default_base_url`** — no `("p", None)` entry; provider defaults carry `base_url="http://b"`; cache `("p", "http://a")` and `("p", "http://b")` in both orders; assert `get_adapter("p")` returns the `"http://b"` adapter.
4. **`test_get_adapter_sorted_tiebreak_is_stable`** — no `None` entry, no matching default; cache `("p", "http://z")` then `("p", "http://a")`; assert the `"http://a"` adapter is returned (sorted-key rule), in both insertion orders.
5. **`test_get_adapter_exact_key_still_wins`** — with multiple entries cached, `get_adapter("p", "http://a")` returns exactly the `("p", "http://a")` adapter (guards against the helper accidentally shadowing the point lookup).
6. **`test_get_adapter_single_creation_under_contention`** — existing double-checked-locking guarantee: N threads call `get_adapter("newprov")` simultaneously with a counting factory; assert the factory ran exactly once and all threads got the same instance.

huangzesen:
Fixes #739.

## Root cause

`LLMService.get_adapter` (`src/lingtai/llm/service.py`) had a lock-free fast path that, when the caller passes no `base_url`, scans the live `self._adapters` dict (`for ... in self._adapters.items()`). The slow path inserts `self._adapters[cache_key] = adapter` under `self._adapter_lock` from another thread; a concurrent insert landing mid-iteration raises `RuntimeError: dictionary changed size during iteration`, surfacing as a spurious LLM failure exactly when several threads race to warm the cache.

## Fix (smallest correct scope)

- Add `_find_provider_adapter`: one shared, snapshot-based provider scan (`list(self._adapters.items())` — a single C-level operation, safe against concurrent inserts) with a documented deterministic preference: (1) the `(provider, None)` entry, (2) the entry matching the provider's configured default base_url, (3) first match in sorted-key order. This also removes the insertion-order nondeterminism for `base_url=None` lookups when several base_urls are cached for one provider (the second defect in #739).
- Fast path and locked double-check in `get_adapter` now reuse that helper; point reads switched from `in` + `[]` to a single atomic `.get()`.
- No schema / signature / back-compat changes. Public behavior changes only in (a) never raising the spurious RuntimeError, and (b) a deterministic (previously unspecified) pick for `base_url=None` multi-entry lookups.

## Validation

Failing-first regression test (`tests/test_llm_service_adapter_cache.py`), run against the unfixed code — reproduces the exact crash from #739 (3/3 runs):

```
AssertionError: get_adapter raced on concurrent insert: [RuntimeError('dictionary changed size during iteration'), ...]
```

After the fix — all 6 new tests pass, race test stable across repeats, under 2s:

```
6 passed in 1.83s
race test: passed 3/3 repeat runs
```

Existing tests for the touched module — unaffected:

```
tests/test_llm_service.py -q          -> 15 passed
tests/test_services_integration.py -q -> 14 passed, 2 skipped (1 pre-existing env failure: test_duckduckgo, ModuleNotFoundError: ddgs)
```

Pre-existing, unrelated env failures in the test venv (also present on pristine main): `test_llm_identity_headers.py` (4, missing adapter deps) and `test_llm_adapter_timeouts.py` (collection, missing anthropic).

Note: an earlier attempt at this crash fix (#787, fable-5/issue-739) was closed unmerged; this PR implements the full #739 plan including the deterministic `_find_provider_adapter` rule and the complete test suite.

### DIFF
--- a/src/lingtai/llm/service.py
+++ b/src/lingtai/llm/service.py
@@ -313,29 +313,36 @@ class LLMService(LLMServiceABC):
         with different base URLs (e.g. OpenRouter vs local vLLM) gets separate
         adapter instances.
 
+        When *base_url* is None and multiple adapters are cached for
+        *provider*, the returned adapter is chosen deterministically: the
+        ``(provider, None)`` entry wins, then the entry whose base_url matches
+        the provider's configured default, then the first match in sorted-key
+        order (see ``_find_provider_adapter``).
+
         Raises RuntimeError if the API key for *provider* is not configured.
         """
         provider = provider.lower()
         cache_key = (provider, base_url)
 
-        # Fast path — no lock needed for reads of an already-cached adapter
-        if cache_key in self._adapters:
-            return self._adapters[cache_key]
-        # When no base_url specified, find any cached adapter for this provider
+        # Fast path — atomic point read, then snapshot-based scan; no lock needed
+        adapter = self._adapters.get(cache_key)
+        if adapter is not None:
+            return adapter
         if base_url is None:
-            for (p, _url), adapter in self._adapters.items():
-                if p == provider:
-                    return adapter
+            adapter = self._find_provider_adapter(provider)
+            if adapter is not None:
+                return adapter
 
         # Slow path — lock to prevent duplicate adapter creation
         with self._adapter_lock:
             # Double-check after acquiring lock
-            if cache_key in self._adapters:
-                return self._adapters[cache_key]
+            adapter = self._adapters.get(cache_key)
+            if adapter is not None:
+                return adapter
             if base_url is None:
-                for (p, _url), adapter in self._adapters.items():
-                    if p == provider:
-                        return adapter
+                adapter = self._find_provider_adapter(provider)
+                if adapter is not None:
+                    return adapter
 
             # Need to create a new adapter — check API key first
             api_key = self._key_resolver(provider)
@@ -353,6 +360,28 @@ class LLMService(LLMServiceABC):
             adapter = self._create_adapter(provider, api_key, effective_base_url)
             self._adapters[cache_key] = adapter
             return adapter
+
+    def _find_provider_adapter(self, provider: str) -> LLMAdapter | None:
+        """Deterministically pick a cached adapter for *provider* (any base_url).
+
+        Preference order:
+        1. the (provider, None) entry,
+        2. the entry matching the provider's configured default base_url,
+        3. first match by sorted cache key (stable tie-break).
+
+        Snapshots the cache so it is safe to call without _adapter_lock.
+        """
+        items = list(self._adapters.items())  # atomic snapshot; safe vs concurrent insert
+        matches = {url: adapter for (p, url), adapter in items if p == provider}
+        if not matches:
+            return None
+        if None in matches:
+            return matches[None]
+        defaults = self._get_provider_defaults(provider)
+        default_url = defaults.get("base_url") if defaults else None
+        if default_url in matches:
+            return matches[default_url]
+        return matches[min(url for url in matches if url is not None)]
 
     def _get_provider_defaults(self, provider_name: str) -> dict | None:
         """Get defaults for a provider from the injected provider_defaults dict."""
--- a/tests/test_llm_service_adapter_cache.py
+++ b/tests/test_llm_service_adapter_cache.py
@@ -1,0 +1,216 @@
+"""Regression tests for LLMService.get_adapter adapter-cache concurrency.
+
+Issue lingtai-kernel#739: the lock-free fast-path scan of ``self._adapters``
+(``base_url=None``) raced with a concurrent insert from the locked slow path,
+raising ``RuntimeError: dictionary changed size during iteration``. The scan is
+now snapshot-based and follows a documented deterministic preference:
+
+1. the ``(provider, None)`` entry,
+2. the entry matching the provider's configured default base_url,
+3. first match by sorted cache key (stable tie-break).
+"""
+
+from __future__ import annotations
+
+import sys
+import threading
+
+from lingtai.llm.service import LLMService
+
+
+def _register_identity_adapter(provider: str) -> None:
+    """Register a hermetic factory returning a fresh opaque object per call."""
+    LLMService.register_adapter(provider, lambda **kwargs: object())
+
+
+def _make_service(
+    provider: str = "p",
+    *,
+    base_url: str | None = None,
+    defaults: dict | None = None,
+) -> LLMService:
+    """Build an LLMService whose adapter cache the test controls exactly.
+
+    The constructor seeds one ``(provider, base_url)`` entry; tests that need
+    a precise cache shape call :meth:`_make_service` and then replace
+    ``svc._adapters`` contents themselves.
+    """
+    _register_identity_adapter(provider)
+    svc = LLMService(
+        provider=provider,
+        model="m",
+        api_key="k",
+        base_url=base_url,
+        key_resolver=lambda p: "k",
+        provider_defaults=defaults or {},
+    )
+    svc._adapters.clear()
+    return svc
+
+
+# ---------------------------------------------------------------------------
+# Regression: lock-free fast-path scan vs concurrent insert (the #739 crash)
+# ---------------------------------------------------------------------------
+
+
+def test_get_adapter_scan_race_with_concurrent_insert():
+    """A base_url=None scan must never raise under a concurrent cache insert.
+
+    On the unfixed code this reproduces ``RuntimeError: dictionary changed
+    size during iteration``: the fast path iterates the live dict while the
+    slow path inserts a fresh ``(provider, base_url)`` key under the lock.
+    """
+    _register_identity_adapter("race-other")
+    _register_identity_adapter("race-b")
+    svc = LLMService(
+        provider="race-other",
+        model="m",
+        api_key="k",
+        key_resolver=lambda p: "k",
+    )
+    svc._adapters.clear()
+    # Every reader scan walks thousands of NON-matching entries before hitting
+    # the single match at the end, so each get_adapter("race-a") call does a
+    # long full-dict iteration (the unfixed fast-path scan is the race window).
+    for i in range(1500):
+        svc._adapters[("race-other", f"http://x/{i}")] = object()
+    svc._adapters[("race-a", "http://end")] = object()
+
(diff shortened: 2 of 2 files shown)


huangzesen:

PR #1277, opened 2026-08-09
fix(llm): forward attribute writes through _GatedSession proxy (#724)

### LINKED ISSUE #724: _GatedSession proxy swallows attribute writes: pre_request_hook and session metadata never reach the inner adapter session
# _GatedSession proxy swallows attribute writes

## Summary

`_GatedSession` (`src/lingtai/llm/base.py:19-77`) forwards attribute *reads* to the inner adapter session via `__getattr__` but has no `__setattr__`, so attribute *writes* land on the proxy's own `__dict__` and never reach the inner session. Since `max_rpm` defaults to `60` (`src/lingtai_kernel/config.py:81`), every session in a default-configured process is gated, which means `BaseAgent`'s `pre_request_hook` install and `LLMService.create_session`'s `session_id` / `_agent_type` / `_tracked` assignments are silently swallowed: the adapter's internal `self.pre_request_hook` check stays `None`, and `get_state()` reports `session_id=""` with default metadata for every gated session.

## Narrative

The rate-gating proxy was hoisted out of the MiniMax adapter so that every provider gets requests-per-minute gating "by inheritance": each adapter's `create_chat` passes its concrete session through `LLMAdapter._wrap_with_gate` (`src/lingtai/llm/base.py:100-110`), which wraps it in `_GatedSession` whenever a gate is configured. All five adapters do this today (`src/lingtai/llm/anthropic/adapter.py:766`, `src/lingtai/llm/openai/adapter.py:2051`, `src/lingtai/llm/gemini/adapter.py:745` and `:781`, `src/lingtai/llm/claude_code/adapter.py:463`, and MiniMax by inheritance). Because the default agent config sets `max_rpm: int = 60` (`src/lingtai_kernel/config.py:81`, `0` disables), the proxy is not an exotic opt-in — it wraps essentially every production chat session.

The proxy's author knew writes would land on the proxy and reasoned it was harmless. The docstring says so explicitly (`src/lingtai/llm/base.py:23-27`):

```
Read-only attribute access falls through to the inner
session via __getattr__; attribute writes (e.g. ``chat.session_id =
...`` from LLMService.create_session) land on the proxy itself, which
is fine because subsequent reads of those names hit the proxy first
and never reach __getattr__.
```

That reasoning is only half true. It holds for *external* readers going through the proxy (`chat.session_id` after `chat.session_id = x` does return `x`). It fails for any code path where the **inner session reads its own attribute internally** — i.e. `self.<attr>` inside a method of the concrete session. Two such paths exist today and both are load-bearing.

**Path 1: the pre-request hook.** `BaseAgent._install_pre_request_hook` (`src/lingtai_kernel/base_agent/__init__.py:1048-1055`) installs the tc_inbox drain hook by assignment:

```python
if not hasattr(self._chat, "pre_request_hook"):
    return
...
self._chat.pre_request_hook = lambda _iface: self._drain_tc_inbox_for_hook()
```

The `hasattr` guard passes — `__getattr__` forwards to the inner session, whose `ChatSession` base class declares `pre_request_hook: Callable | None = None` (`src/lingtai_kernel/llm/base.py:155`) — so the assignment proceeds and lands on the proxy's `__dict__`. But the hook is *fired* from inside the concrete adapter's `send()` / `send_stream()`, on the inner `self`:

```python
# src/lingtai/llm/anthropic/adapter.py:389-390 (same pattern at :472-473)
if self.pre_request_hook is not None:
    self.pre_request_hook(self._interface)
```

The identical check appears in `src/lingtai/llm/openai/adapter.py:1389`, `:1564`, `:1806`, `:1838`, `:3404`, `src/lingtai/llm/gemini/adapter.py:163`, `:398`, `:445`, and `src/lingtai/llm/claude_code/adapter.py:221`. `self` there is the inner session, whose `pre_request_hook` is still the class default `None`. The hook installed through the proxy never fires. One honest caveat on severity: per `src/lingtai/llm/ANATOMY.md` (pre-request hook section), the mid-turn tc_inbox drain is currently dormant in production — after the `.notification/` redesign the queue is empty at send time and ACTIVE notifications defer to the post-turn IDLE path. So today this is a broken contract rather than visible data loss. But the hook is still installed on every session (including sessions rebuilt by AED recovery, which the install docstring explicitly calls out), the ANATOMY docs describe it as a live convention until "Phase 3" removes it, and any future producer that relies on the documented send-time drain will silently get nothing on gated sessions — which is all default sessions.

**Path 2: session identity and tracking metadata.** `LLMService.create_session` (`src/lingtai/llm/service.py:395-402`) stamps the session after creation:

```python
if tracked:
    chat.session_id = _generate_session_id()
    chat._agent_type = agent_type
    chat._tracked = True
    self._sessions[chat.session_id] = chat
else:
    chat.session_id = ""
    chat._tracked = False
```

All three writes land on the proxy. `ChatSession.get_state` (`src/lingtai_kernel/llm/base.py:254-267`) reads `self.session_id`, `self._agent_type`, and `self._tracked` — and because the proxy defines no `get_state`, the call resolves through `__getattr__` to the *inner* session's bound method, so `self` is the inner session and the returned state is `{"session_id": "", ..., "agent_type": "", "tracked": True}` regardless of what the service assigned. Any consumer of `get_state()` (session persistence, debugging dumps, external tooling) sees identity-less, wrongly-tracked state for every gated session. Note the inversion: an *untracked* session (`_tracked = False` written to the proxy) still reports `tracked: True` from the inner default.

The friction is already visible in the test suite: `tests/integration_test_compaction.py:64` carries a workaround comment — "Handle _GatedSession proxy — reach inner session if needed". A transparent proxy should never require callers to reach around it.

Better looks like: writes through the proxy behave exactly like reads — they go to the inner session, except for the proxy's two private slots. That makes the proxy transparent in both directions, fixes both paths above with no call-site changes, and removes the trap for every future attribute the kernel or service sets on a session.

## Current behavior

- `_GatedSession.__init__` (`src/lingtai/llm/base.py:30-34`) stores `_inner` and `_gate` via `object.__setattr__`; the class defines `__getattr__` (`:75-77`) but no `__setattr__`.
- `LLMService.create_session` (`src/lingtai/llm/service.py:395-402`) writes `session_id`, `_agent_type`, `_tracked` onto the proxy; the inner session keeps class defaults `""` / `""` / `True` (`src/lingtai_kernel/llm/base.py:126-129`).
- `ChatSession.get_state` (`src/lingtai_kernel/llm/base.py:254-267`) executes with the inner session as `self` (resolved via proxy `__getattr__`), so it returns the stale defaults — including `session_id: ""` for tracked sessions and `tracked: True` for untracked ones.
- `BaseAgent._install_pre_request_hook` (`src/lingtai_kernel/base_agent/__init__.py:1055`) writes `pre_request_hook` onto the proxy; every adapter `send`/`send_stream` reads `self.pre_request_hook` on the inner session (e.g. `src/lingtai/llm/anthropic/adapter.py:389`, `src/lingtai/llm/openai/adapter.py:1389`, `src/lingtai/llm/gemini/adapter.py:163`, `src/lingtai/llm/claude_code/adapter.py:221`) and finds `None`, so the hook never fires on gated sessions.
- Reads through the proxy of previously written names return the proxy-local values, which masks the divergence — externally the session *looks* configured.

1. **Add `__setattr__` to `_GatedSession`** in `src/lingtai/llm/base.py`, forwarding everything except the proxy's own two slots:

   ```python
   _PROXY_SLOTS = frozenset({"_inner", "_gate"})

   def __setattr__(self, name, value):
       if name in _PROXY_SLOTS:
           object.__setattr__(self, name, value)
       else:
           setattr(self._inner, name, value)
   ```

   Define `_PROXY_SLOTS` at module level (or as a class attribute) so `__init__` and `__setattr__` share one source of truth. With this in place, the `object.__setattr__` calls in `__init__` (`src/lingtai/llm/base.py:33-34`) can stay as-is (belt and braces) or become plain assignments — either works because `_inner`/`_gate` route to `object.__setattr__`.

2. **Add `__delattr__` for symmetry** (small, prevents the same trap for `del chat.attr`):

   ```python
   def __delattr__(self, name):
       if name in _PROXY_SLOTS:
           object.__delattr__(self, name)
       else:
           delattr(self._inner, name)
   ```

3. **Fix the docstring** at `src/lingtai/llm/base.py:20-28`: replace the "writes land on the proxy itself, which is fine" paragraph with a statement that writes forward to the inner session (except `_inner`/`_gate`), because the inner session reads attributes like `pre_request_hook`, `session_id`, `_agent_type`, and `_tracked` internally (adapter `send()` paths, `ChatSession.get_state`).

4. **Check the `interface` property interaction.** `_GatedSession.interface` (`src/lingtai/llm/base.py:36-38`) is a getter-only property. Today `proxy.interface = x` raises `AttributeError`; with the new `__setattr__` it forwards to the inner session instead. No repo code assigns `.interface` on a session, and forwarding is the semantically correct behavior for a transparent proxy, so no extra handling is needed — but note it in the docstring/PR description.

5. **No schema or migration work.** The proxy is a transient in-process object; nothing persists proxy `__dict__` contents. Back-compat: external readers that previously saw proxy-local values still see the same values after the fix (writes now live on the inner session and `__getattr__` forwards reads there), so behavior only changes where it was wrong — internal reads by the inner session.

6. **Optionally simplify the workaround** at `tests/integration_test_compaction.py:64` ("reach inner session if needed") once the proxy is transparent for writes; verify whether that unwrap is still needed for its specific purpose before removing.

7. **Do not change** `BaseAgent._install_pre_request_hook` or `LLMService.create_session` — their write-through-the-session style is exactly what the proxy should support.

## Risks and alternatives

- **Risk: code that deliberately parked private state on the proxy.** A repo-wide search shows only `_inner`/`_gate` (set in `__init__`) as intentional proxy-local state; all other writes (`session_id`, `_agent_type`, `_tracked`, `pre_request_hook`) are meant for the session. The slot allowlist preserves the two real proxy fields.
- **Risk: name shadowing on reads.** The proxy defines `send`, `send_stream`, `interface`, and the comment/hook forwarders as real class attributes; if someone assigned e.g. `chat.send = fn`, the write would now reach the inner session but reads through the proxy would still hit the proxy's method. No such assignment exists in the repo; this is a theoretical edge, unchanged in spirit from today.
- **Alternative A — unwrap at call sites** (make `create_session` and `_install_pre_request_hook` write to `chat._inner` when present): spreads proxy-awareness across the kernel and service layers, breaks encapsulation, and every future write site is a new landmine. Rejected.
- **Alternative B — replicate hook firing on the proxy** (have `_GatedSession.send` invoke a proxy-level `pre_request_hook` before delegating): fixes only the hook, not `get_state()` metadata, and creates two places where the hook contract lives (the wire-state contract documented at `src/lingtai_kernel/llm/base.py:131-155` is deliberately adapter-internal — the hook must fire *after* the message is committed to the canonical interface, which only the adapter knows). Rejected.
- **Alternative C — subclass `ChatSession` instead of duck-typed proxy**: larger refactor, must mirror the ABC's abstract surface, and doesn't inherently fix write forwarding. Rejected.
- The chosen `__setattr__` fix is ~10 lines, keeps the proxy fully transparent in both directions, and fixes both known paths plus all future ones.

## Testing plan

Add to `tests/test_max_rpm_plumbing.py` (which already builds `_GatedSession` via `_wrap_with_gate` and asserts wrapping at `:71`), or a new `tests/test_gated_session_proxy.py`:

1. `test_setattr_forwards_to_inner_session` — wrap a fake `ChatSession` with a gate; set `proxy.pre_request_hook = fn`; assert `inner.pre_request_hook is fn` and `"pre_request_hook" not in object.__getattribute__(proxy, "__dict__")`.
2. `test_proxy_slots_stay_on_proxy` — after construction, assert `proxy.__dict__` contains exactly `_inner` and `_gate`, and that reassigning `proxy._gate` does not touch the inner session.
3. `test_pre_request_hook_fires_inside_gated_send` — fake inner session whose `send()` calls `self.pre_request_hook(self._interface)` when set (mirroring `src/lingtai/llm/anthropic/adapter.py:389-390`); install the hook through the proxy; call `proxy.send("hi")`; assert the hook ran exactly once. This test fails on current main.
4. `test_get_state_reflects_service_assigned_identity` — set `proxy.session_id = "abc"`, `proxy._agent_type = "worker"`, `proxy._tracked = True`; assert `proxy.get_state()["session_id"] == "abc"` and `metadata["agent_type"] == "worker"`. Fails on current main (returns `""` / `""`).
5. `test_untracked_session_reports_untracked` — set `proxy.session_id = ""`, `proxy._tracked = False`; assert `proxy.get_state()["metadata"]["tracked"] is False`. Fails on current main (inner default `True`).
6. `test_create_session_identity_survives_gating` — service-level: `LLMService.create_session` with a gated fake adapter; assert the returned chat's `get_state()["session_id"]` matches `chat.session_id` and the session is retrievable via `LLMService.get_session` under that id.
7. `test_delattr_forwards` (if step 2 of the plan is included) — set then `del proxy.some_attr`; assert it is removed from the inner session.

huangzesen:
## Summary

- Add `__setattr__` (and `__delattr__` for symmetry) to `_GatedSession` (`src/lingtai/llm/base.py`) so attribute writes forward to the inner adapter session, keeping only the proxy's own `_inner`/`_gate` slots (`_PROXY_SLOTS`, one shared source of truth) on the proxy.
- Update the `_GatedSession` docstring and the `src/lingtai/llm/ANATOMY.md` "Gated sessions" note to state the proxy is now transparent in both directions.
- Add `tests/test_gated_session_proxy.py` — seven regression tests, five of which fail on pre-fix main.
- Update `tests/test_max_rpm_plumbing.py::test_gated_session_routes_send_through_gate`, which had encoded the buggy proxy-local write contract (asserted `inner.session_id` untouched); it now asserts the transparent write-forwarding behavior.

## Root cause

`_GatedSession` forwarded attribute *reads* to the inner session via `__getattr__` but defined no `__setattr__`, so attribute *writes* landed on the proxy's own `__dict__` and never reached the inner session. Because default `max_rpm` is `60` (`0` disables), essentially every production session is gated, so this trapped every default-configured process.

The old docstring reasoned the writes-on-proxy behavior was "fine because subsequent reads hit the proxy first." That holds only for *external* readers going through the proxy. It fails wherever the inner session reads its own attribute internally (`self.<attr>` inside its methods), and two such paths exist:

- **Session identity / tracking (still broken on main).** `LLMService.create_session` assigns `session_id` / `_agent_type` / `_tracked`; `ChatSession.get_state()` reads them on the inner `self`, so `get_state()` reported `session_id=""` / `agent_type=""` / `tracked=True` for every gated session — even inverting an *untracked* session's `tracked` flag. This PR fixes this path.
- **Pre-request hook (already fixed on main).** `BaseAgent._install_pre_request_hook` assigns `pre_request_hook`, and every adapter fires it as `self.pre_request_hook` from inside `send()`/`send_stream()`. Main already carries the named getter/setter property (commit `3effb442`) that forwards the hook to the inner session; the new `__setattr__` makes that property belt-and-braces rather than a special case.

A prior attempt (PR #772, closed unmerged) proposed the same `__setattr__` approach; only the narrow property fix landed on main. This PR completes the transparency contract at the owning layer (the proxy), with no call-site changes — closing the whole class of bug for every future attribute the kernel or service sets on a session.

## Validation

Run with the repo venv interpreter (worktree `src` on `PYTHONPATH` via pytest's `pythonpath = ["src"]`).

Fail-first (stash the `base.py` fix, keep the new test), pre-fix `_GatedSession`:

```
$ .venv/bin/python -m pytest tests/test_gated_session_proxy.py -q
FAILED tests/test_gated_session_proxy.py::test_setattr_forwards_to_inner_session
FAILED tests/test_gated_session_proxy.py::test_get_state_reflects_service_assigned_identity
FAILED tests/test_gated_session_proxy.py::test_untracked_session_reports_untracked
FAILED tests/test_gated_session_proxy.py::test_delattr_forwards_to_inner_session
FAILED tests/test_gated_session_proxy.py::test_create_session_identity_survives_gating
5 failed, 2 passed in 0.11s
```

(The 2 pre-fix passes are the slot-invariant guard and `test_pre_request_hook_fires_inside_gated_send`, which already passes via the named property landed in `3effb442`.)

After the fix:

```
$ .venv/bin/python -m pytest tests/test_gated_session_proxy.py -q
7 passed in 0.09s
```

Adjacent gate/service/session suites (all green):

```
$ .venv/bin/python -m pytest tests/test_max_rpm_plumbing.py tests/test_llm_service.py \
    tests/test_session.py tests/test_api_gate.py tests/test_agent_session.py \
    tests/test_gated_session_proxy.py -q
105 passed, 1 warning in 3.61s

$ git diff --check
(clean)
```

`tests/test_tc_inbox*.py` show the same pre-existing failures with and without this change (environment-related, unrelated to the proxy).

## Non-goals

- No changes to `BaseAgent._install_pre_request_hook` or `LLMService.create_session` — their write-through-the-session style is exactly what the proxy should now support.
- No schema/migration work: the proxy is a transient in-process object; nothing persists its `__dict__`.

Fixes #724


huangzesen:

PR #1278, opened 2026-08-09
fix(refresh): serialize watcher system.json merge with the notification store lock

### LINKED ISSUE #742: Cross-process read-modify-write race on .notification/system.json between agent and refresh watcher
## Summary

`.notification/system.json` is merged (read → append event → cap at 20 → write) by two different **processes**: the agent itself and the refresh watcher subprocess. All in-agent writers serialize on a per-agent `threading.Lock`, but a `threading.Lock` is invisible to the watcher process, whose generated code does its own unguarded read-modify-write of the same file. A watcher publishing `refresh_failed_permanent` concurrently with an in-agent system-event burst (e.g. mail bounces) can silently lose events in either direction. A cross-process advisory file lock (`fcntl.flock` / `msvcrt.locking`, both already used elsewhere in this repo) around both merge paths closes the gap.

## Narrative

The system notification channel is deliberately a single file that multiplexes event types. `_enqueue_system_notification` documents the concurrency contract it relies on (`src/lingtai_kernel/base_agent/messaging.py:151-155`):

```
The merge is read-modify-write on the same file, so concurrent
arrivals (e.g. a burst of bounces) need a per-agent lock to avoid
losing writes.  The lock is initialized by ``BaseAgent``; only
``system.json`` needs it because ``email.json`` and ``soul.json``
recompute full state on every publish (no merge).
```

That lock is `agent._system_notification_lock`, a plain `threading.Lock` created in `src/lingtai_kernel/base_agent/__init__.py:503`. Every in-process writer honors it: the enqueue merge (`base_agent/messaging.py:186-231`), large-result reminder clearing (`notifications.py:368-373`), system-event dismissal (`notifications.py:987-989`), and the goal nudge channel writer (`nudge/goal.py:126`). Within one Python process this is airtight.

The problem is that the agent process is not the only writer. When a refresh is dispatched, `_perform_refresh` spawns a detached watcher **subprocess** whose Python source is generated inline in `src/lingtai_kernel/base_agent/lifecycle.py`. When all relaunch attempts are exhausted, the watcher makes the dead agent diagnosable by appending a high-priority `refresh_failed_permanent` event to the very same `system.json`, via its own `_append_system_notification` (`base_agent/lifecycle.py:897-935`):

```python
"def _append_system_notification(meta):\n"
"    notif_dir = os.path.join(wd, '.notification')\n"
"    target = os.path.join(notif_dir, 'system.json')\n"
"    current = {}\n"
"    try:\n"
"        with open(target, encoding='utf-8') as f:\n"
"            current = json.load(f)\n"
...
"    events_list.append({...})\n"
"    events_list = events_list[-20:]\n"
...
"    _write_json_atomic(target, payload)\n"
```

This is the same read-modify-write shape as the in-agent merge, but the watcher cannot see the agent's `threading.Lock` — thread locks do not cross process boundaries. The only protection here is `_write_json_atomic` (tmp + `os.replace`, `lifecycle.py:860-866`), which prevents *torn reads* but does nothing about *lost updates*: whichever writer replaces the file last wins, and the loser's event vanishes.

It is tempting to dismiss this as "the agent is dead when the watcher publishes, so there is no concurrency." That is not reliable. `_publish_refresh_failed_permanent` fires after `MAX_ATTEMPTS` relaunch attempts, and the aliveness signal the watcher uses is heartbeat freshness (`is_alive()` requires heartbeat age < 30s, `lifecycle.py:985-987`). A relaunched agent can have started far enough to run its heartbeat-driven notification producers (mail bounce events, worker-hang notifications, nudges) and then stall — a hung event loop with a stale heartbeat is still a live process whose threads can be mid-merge on `system.json` at the moment the watcher does its read → append → replace. The interleaving `watcher reads → agent merges bounce event + writes → watcher writes` drops the bounce event; the reverse ordering drops the `refresh_failed_permanent` event — which is precisely the event whose whole purpose is "make the dead agent diagnosable rather than silently gone."

Losses are especially hard to notice because the file is capped at the 20 most recent events (`messaging.py:207-208`, mirrored at `lifecycle.py:925`): an occasionally missing event looks exactly like normal cap-driven eviction, so nobody files a bug — the agent just never learns about a bounce, or the human never sees why the refresh died.

The current design got here honestly: the per-agent thread lock was added when all `system.json` writers lived inside the agent process, and the watcher's terminal-failure publisher was added later as self-contained generated code that could not import agent state (by design — it must outlive the agent). Nobody re-audited the lock's scope when the second process started merging the file.

Better looks like what the repo already does elsewhere for cross-process exclusion: `src/lingtai_kernel/workdir.py:15-30` defines `_lock_fd`/`_unlock_fd` over `fcntl.flock` (POSIX) and `msvcrt.locking` (Windows) for `.agent.lock`, and `maintenance/retention.py:702-714` uses `fcntl.flock` for its own guard. The fix is a small shared helper that takes an exclusive advisory lock on a sidecar lock file for the duration of any `system.json` merge, used by both the in-agent writers and the watcher's generated code.

## Current behavior

- `src/lingtai_kernel/base_agent/__init__.py:503` — `self._system_notification_lock: threading.Lock = threading.Lock()`; process-local only.
- `src/lingtai_kernel/base_agent/messaging.py:133-251` — `_enqueue_system_notification` holds that lock across `collect_notifications(...)` read → event append → 20-cap → `publish_notification(...)` write.
- `src/lingtai_kernel/notifications.py:339-373` (`clear_large_result_reminders`), `notifications.py:987-989` (system-channel dismiss), and `src/lingtai_kernel/nudge/goal.py:126` — all take the same thread lock for their merges/rewrites.
- `src/lingtai_kernel/base_agent/lifecycle.py:897-935` (`_append_system_notification` in the generated watcher script) and `lifecycle.py:936-961` (`_publish_refresh_failed_permanent`) — read-modify-write the same file from a **separate process** with no lock of any kind. `_write_json_atomic` (`lifecycle.py:860-866`) makes the final write atomic but the read→write window is unguarded.
- `src/lingtai_kernel/notifications.py:260-276` (`publish`) — atomic tmp+rename write; same property, same limitation.
- Result: a lost-update race between the agent process and the watcher process. Loss is silent and masked by the 20-event cap.

1. **Add a cross-process lock helper in `src/lingtai_kernel/notifications.py`** (this module already owns the channel-file protocol and is importable by both sides):

   ```python
   from contextlib import contextmanager

   _SYSTEM_LOCK_FILENAME = ".system.merge.lock"  # not *.json → invisible to collect_notifications

   @contextmanager
   def system_events_file_lock(workdir: Path, *, timeout: float = 5.0):
       """Exclusive cross-process lock for system.json read-modify-write merges.

       Locks a sidecar file (never system.json itself: publish() swaps the
       inode via os.replace, which would defeat an flock on the data file).
       Blocking-with-timeout; on timeout, proceeds unlocked (fail-open, today's
       behavior) and yields False so callers can log the degradation.
       """
       notif_dir = workdir_layout(workdir).notification_dir
       notif_dir.mkdir(exist_ok=True)
       lock_path = notif_dir / _SYSTEM_LOCK_FILENAME
       # open + retry non-blocking _lock_fd (reuse workdir.py's fcntl/msvcrt
       # pair) until acquired or deadline; yield acquired-bool; always unlock/close.
   ```

   Reuse (or lift into a tiny shared module) the `_lock_fd`/`_unlock_fd` platform pair from `src/lingtai_kernel/workdir.py:15-30` rather than re-implementing the win32 branch.

2. **Wrap every in-agent merge with the file lock, keeping the thread lock as the outer guard** (lock ordering rule: `threading.Lock` outer → file lock inner, documented in the helper docstring to prevent deadlock):
   - `base_agent/messaging.py` `_enqueue_system_notification`: inside `with lock:` add `with system_events_file_lock(agent._working_dir):` around the `collect_notifications` → `publish_notification` span (`messaging.py:188-231`).
   - `notifications.py` `clear_large_result_reminders` (`_do_clear`), the `channel == "system"` dismiss branch (`notifications.py:987-989`), and `nudge/goal.py:126` — same pattern. Since these all already funnel their file rewrite through `_rewrite_system_events_locked` / `publish`, the cleanest shape is to take the file lock at the same call sites that currently take the thread lock.

3. **Guard the watcher's merge in the generated script** (`base_agent/lifecycle.py`, `_append_system_notification`). The watcher script is self-contained by design, so inline a minimal equivalent rather than importing the kernel helper:

   ```python
   "def _locked_system_merge(fn):\n"
   "    lock_path = os.path.join(wd, '.notification', '.system.merge.lock')\n"
   "    fd = open(lock_path, 'a+')\n"
   "    deadline = time.time() + 5\n"
   "    got = False\n"
   "    while time.time() < deadline:\n"
   "        try:\n"
   "            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB); got = True; break\n"
   "        except OSError:\n"
   "            time.sleep(0.1)\n"
   "    try:\n"
   "        return fn()\n"   # read → append → cap → _write_json_atomic
   "    finally:\n"
   "        if got: fcntl.flock(fd, fcntl.LOCK_UN)\n"
   "        fd.close()\n"
   ```

   Wrap the existing body of `_append_system_notification` in this. Fail-open on timeout (log `refresh_failed_permanent_lock_timeout` through the watcher's redacting `log()`) — the watcher must never hang forever on a lock held by a wedged agent, and unlocked-after-timeout is exactly today's behavior. Guard the `import fcntl` for win32 (fall back to `msvcrt.locking`, mirroring `workdir.py`, or skip locking with a logged marker).

4. **Keep the lock file out of every channel-reading surface.** `collect_notifications` globs `*.json` and filters through `is_channel_allowed` (`notifications.py:234-258`), so a non-`.json` sidecar is already invisible; verify the `.notification` directory fingerprint function (`notifications.py:~225`) also only hashes `*.json` so lock-file mtime churn cannot trigger spurious notification syncs. Add the sidecar name to any workdir-hygiene/purge allowlists if one exists.

5. **Back-compat / migration:** none needed on the data file — the schema of `system.json` is unchanged. Old watcher scripts already spawned (pre-upgrade) simply won't take the lock; the window shrinks to today's status quo for those, and fully closes after the next refresh. Document the ordering + sidecar contract in `base_agent/ANATOMY.md` entries for `messaging.py` and `lifecycle.py`, which currently describe the thread lock as the complete answer.

## Risks and alternatives

- **Deadlock risk:** two locks now exist. Mitigated by a strict documented order (thread lock → file lock) and by never calling back into another lock-taking helper while holding the file lock. `system_events_file_lock` must not be re-entered; the helper should be used only at the four writer call sites.
- **Watcher hang risk:** a crashed-while-holding-flock process releases the lock automatically (kernel-managed), so the main hazard is a *live but wedged* agent holding it. The 5s timeout + fail-open keeps the watcher's terminal-failure publication guaranteed, degrading to today's unlocked behavior with a logged marker.
- **Network filesystems:** `flock` semantics on NFS are historically shaky. Agent workdirs already rely on `flock` for `.agent.lock` (`workdir.py`), so this adds no new assumption.
- **Windows:** `msvcrt.locking` locks byte ranges and is non-blocking in `LK_NBLCK` mode; the retry loop handles that. Worst case on exotic platforms, the fail-open path preserves current behavior.
- **Alternative — append-only event journal** (`system.events.jsonl` + reader-side merge/compaction): eliminates read-modify-write entirely and is the most principled fix, but changes the channel file protocol consumed by `collect_notifications`, the sync fingerprint, dismissal, and the TUI — a much larger change for the same practical result.
- **Alternative — optimistic concurrency** (re-read after write, or compare inode/mtime and retry): no lock-file, but retry loops in the watcher's terminal path are harder to reason about than a bounded lock wait, and ABA windows remain.
- **Alternative — lock `system.json` itself:** broken by design here, because `publish` replaces the inode via `os.replace`; a lock on the old inode guards nothing. Hence the sidecar. The sidecar-flock plan wins on smallest diff, existing in-repo precedent, and zero schema change.

## Testing plan

- `tests/test_system_notifications.py::test_system_events_file_lock_mutual_exclusion` — two `multiprocessing` children each acquire `system_events_file_lock` around a critical section that records enter/exit timestamps; assert the sections never overlap.
- `tests/test_system_notifications.py::test_concurrent_cross_process_merges_lose_no_events` — regression test for the bug: N subprocesses each append one uniquely ref-id'd event through the locked merge path (with N ≤ 20); assert all N events are present in the final `system.json`. (Optionally add a companion test demonstrating loss with the lock helper monkeypatched to a no-op, marked as documentation of the failure mode.)
- `tests/test_system_notifications.py::test_lock_timeout_fails_open` — hold the sidecar lock in the test process, call the helper with a tiny timeout, assert it yields `False`, the merge still completes, and no exception escapes.
- `tests/test_deep_refresh.py::test_watcher_append_system_notification_respects_lock` — exercise the generated watcher script's `_append_system_notification` (via the same mechanism existing deep-refresh tests use to run watcher code) while the test holds the sidecar lock and injects a competing event mid-window; assert both the injected event and `refresh_failed_permanent` survive.
- `tests/test_notification_sync.py::test_lock_sidecar_invisible_to_channel_readers` — create `.notification/.system.merge.lock`, assert `collect_notifications` output and the directory fingerprint are unchanged by its presence/mtime churn.
- Extend `tests/test_deep_refresh.py`'s existing terminal-failure test to assert the new `refresh_failed_permanent_lock_timeout` marker is absent in the normal path.

huangzesen:
Fixes #742.

## Problem

The generated refresh-watcher program publishes its terminal `refresh_failed_permanent` alert by merging into `.notification/system.json` with an **unguarded read-modify-write** (read -> append -> 20-event cap -> atomic replace), while every in-agent writer (`_enqueue_system_notification`, notifications dismiss/clear, goal nudges, daemon terminal events) serializes through the Notification Store under an advisory flock on `.notification/.store.lock`. A `threading.Lock` never crossed the process boundary; `_write_json_atomic` prevents torn reads but not lost updates. A watcher publishing concurrently with an in-agent system-event burst can silently lose events in either direction, and the 20-event cap masks the loss.

## Fix

The generated watcher now takes the **same** `.store.lock` sidecar flock the in-agent Notification Store uses, around its entire `system.json` merge:
- non-blocking `fcntl.flock` with a bounded 5s retry deadline;
- **fail-open** on timeout / no-fcntl / open failure, with diagnosable markers (`refresh_failed_permanent_lock_timeout`, `refresh_failed_permanent_lock_unavailable`) so the terminal alert is never dropped by a wedged holder;
- lock release is guaranteed in `finally` (kernel also releases on process death), so the watcher can never hang.

Lock ordering stays single-level on the watcher side (no thread lock there), and the in-agent side is unchanged — this only closes the missing second participant.

## Tests

- `tests/test_perform_refresh_handshake.py::test_refresh_watcher_system_notification_append_serializes_with_store_lock` — parent holds `.store.lock` and performs a locked in-agent merge while the watcher child appends its alert; asserts the child blocks while the lock is held **and** both events survive (lost-update regression). Verified to fail on the pre-fix code.
- `tests/test_perform_refresh_handshake.py::test_refresh_watcher_system_notification_lock_timeout_fails_open` — lock held past the (shortened) deadline; asserts the alert still publishes and the timeout marker is logged. Verified to fail on the pre-fix code.

Local results (Python 3.13, macOS):
- `pytest tests/test_perform_refresh_handshake.py tests/test_refresh_watcher_process.py tests/test_refresh_watcher_windows.py tests/test_system_notifications.py tests/test_notification_store.py tests/test_process_match.py` → 136 passed, 2 skipped
- `pytest tests/test_deep_refresh.py` → 34 passed
- `ruff check` on changed files → clean

## Docs

`src/lingtai/kernel/notification_store/ANATOMY.md` now documents the watcher as a participant in the `.store.lock` contract.


huangzesen:

PR #1279, opened 2026-08-09
fix(daemon): preserve nested provider_defaults in detached child reconstruction (lingtai#750)

huangzesen:
Fixes Lingtai-AI/lingtai#750 (cross-repo reference; the issue was filed on `Lingtai-AI/lingtai` but the affected kernel code lives in this repo).

## Problem

Detached LingTai daemon execution drops provider-specific values stored under the public nested `provider_defaults.<provider>` configuration. A custom OpenAI-compatible provider configured with `wire_api: responses` is reconstructed as `wire_api: auto` inside the daemon child, which switches the selected session from the Responses API to Chat Completions. The daemon then calls `/chat/completions`; when the compatibility layer returns a scalar string rather than a `ChatCompletion`, LingTai dereferences `.choices` and fails at turn 0 with `AttributeError: 'str' object has no attribute 'choices'`.

Root cause: the parent serializes the effective provider bucket into the run manifest under the **public** `provider_defaults` key, while the private `_provider_defaults` alias never crosses the process boundary. The child's `_run_emanation` only read `_provider_defaults` (absent in the child) and fell back to `_llm_defaults_from_manifest`, which reads top-level manifest `llm` keys only — so the nested bucket (with `wire_api`, `api_compat`, `max_rpm`, …) was dropped.

Reproduced on current main with a focused test: the reconstructed `provider_defaults` contained only `base_url`; `wire_api` was lost (matches the issue's probe: `reconstructed provider config: base_url only; resolved wire_api: auto`).

## Fix

1. `src/lingtai/tools/daemon/__init__.py` — in `DaemonManager._run_emanation`, when the private `_provider_defaults` alias is absent, merge the nested public `provider_defaults[provider]` bucket over the manifest-derived defaults so provider-specific fields survive the detached-child reconstruction (nested bucket wins on key overlap).
2. `src/lingtai/llm/openai/adapter.py` — defense-in-depth: `_parse_response` now raises a typed `TypeError` for a non-`ChatCompletion` response (e.g. a scalar string) instead of failing through the bare `raw.choices` attribute dereference.

## Tests

- `tests/test_daemon.py::test_run_emanation_detached_child_merges_public_provider_defaults` (new) — simulates the detached child receiving the public `provider_defaults` map and asserts `wire_api: responses` (plus `api_compat`, `max_rpm`, `default_headers`) reaches `LLMService`. Failed on unpatched main (`wire_api` dropped), passes with the fix.
- `tests/test_wire_api.py::test_parse_chat_completion_rejects_non_object_response` (new) — a scalar-string response raises `TypeError` with a clear message.

Evidence:

```
$ pytest tests/test_wire_api.py tests/test_daemon.py::test_run_emanation_detached_child_merges_public_provider_defaults -q
48 passed

$ pytest tests/test_daemon.py tests/test_daemon_backend_options.py tests/test_daemon_detached_supervisor.py -q
226 passed

$ pytest tests/test_openai_responses_streaming.py tests/test_custom_responses_stateless.py tests/test_openai_overflow_recovery.py tests/test_openai_compact_threshold.py -q
107 passed, 2 failed  # the 2 failures are pre-existing on clean main (verified via git stash)
```

## Notes

Branch name deviates from the shared template (`fix/issue-750-dev4-20260809` was already taken by a sibling lane's PR for a different issue, kernel#750); this branch is `fix/lingtai-issue-750-provider-defaults-dev4-20260809`.

### DIFF
--- a/src/lingtai/llm/openai/adapter.py
+++ b/src/lingtai/llm/openai/adapter.py
@@ -1258,6 +1258,17 @@ def _parse_tool_calls(raw_tool_calls) -> list[ToolCall]:
 
 def _parse_response(raw) -> LLMResponse:
     """Parse a raw OpenAI ChatCompletion into a provider-agnostic LLMResponse."""
+    if not hasattr(raw, "choices"):
+        # Defense-in-depth: a non-conforming gateway (e.g. an OpenAI-compatible
+        # endpoint that returns a scalar string) must surface a typed protocol
+        # error instead of failing through a bare attribute dereference. This
+        # is deliberately a TypeError, not a ValueError: the response object
+        # has the wrong type, not merely an empty payload.
+        raise TypeError(
+            "OpenAI-compatible endpoint returned a non-ChatCompletion response "
+            "(expected an object with a `choices` list); check the provider's "
+            "wire_api/endpoint compatibility"
+        )
     if not raw.choices:
         return LLMResponse(raw=raw)
 
--- a/src/lingtai/tools/daemon/__init__.py
+++ b/src/lingtai/tools/daemon/__init__.py
@@ -2848,6 +2848,21 @@ class DaemonManager:
             base_defaults = effective_preset_llm["_provider_defaults"]
         else:
             base_defaults = self._llm_defaults_from_manifest(effective_preset_llm)
+            # A detached child receives the provider bucket under the public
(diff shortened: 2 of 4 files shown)


huangzesen:
Superseded by #1330. Closing as superseded.

huangzesen:
Superseded by #1330. Closing as superseded.

PR #1281, opened 2026-08-09
fix(mail): stop silently dropping unreadable inbox messages (#732)

### LINKED ISSUE #732: FilesystemMailService Phase 1 poller permanently and silently drops messages whose message.json fails to read or parse
## Summary

In `FilesystemMailService.listen()`'s Phase 1 inbox loop (`src/lingtai_kernel/services/mail.py:241-248`), any `json.JSONDecodeError` or `OSError` raised while reading `message.json` is swallowed with a bare `pass`, and the entry is then unconditionally added to `_seen`. The message is lost forever with zero logging: a transient read failure (permissions blip, network-filesystem hiccup) is treated the same as a permanently corrupt file, and neither leaves any trace. The MCP inbox poller in `src/lingtai/core/mcp/inbox.py` already models the correct behavior (`_dead_letter` + `log.warning`); the mail poller should distinguish transient from permanent failures, retry the former, dead-letter the latter, and log both.

## Narrative

`FilesystemMailService` is the kernel's inter-agent mail transport. Agents deliver mail by writing a `message.json` (plus optional `attachments/`) into a UUID-named directory under the recipient's `mailbox/inbox/`, and each agent runs a background poll thread (`listen()` → `_poll_loop`, polling every 0.5s) that dispatches newly appeared messages to an `on_message` callback. Deduplication is done with an in-memory `_seen` set of directory names.

The Phase 1 loop currently reads:

```python
# src/lingtai_kernel/services/mail.py:241-248
if msg_file.is_file():
    try:
        payload = json.loads(msg_file.read_text(encoding="utf-8"))
        on_message(payload)
    except (json.JSONDecodeError, OSError):
        pass
    self._seen.add(entry.name)
```

Consider what happens when an agent's inbox sits on a network filesystem (or the host is under memory/IO pressure) and `read_text()` raises a transient `OSError` on exactly one poll cycle. The `except` clause swallows it, and the very next line adds `entry.name` to `_seen`. On every subsequent cycle the entry is skipped at the `if entry.name in self._seen` check. The message — which is perfectly intact on disk — will never be delivered. The sending agent believes delivery succeeded (the file write did succeed), the receiving agent never wakes, and nothing is logged anywhere. From the operator's point of view, an agent simply "didn't respond to a message that is sitting right there in its inbox directory," which is one of the most confusing failure modes a mailbox system can have. The same applies to a genuinely malformed `message.json`: it silently disappears into `_seen` with no dead-letter artifact and no log line to grep for.

There is a second, subtler consequence of the current `try` block: `on_message(payload)` is *inside* the same `try`. If the callback itself raises an `OSError` (e.g., the handler touches the filesystem), the message is also silently dropped and marked seen — the error class of the transport is conflated with the error class of the consumer.

It's easy to see how the design got here: the poll loop must never die, so defensive exception swallowing was applied broadly, and marking the entry seen prevents a hot loop on a permanently bad file. Those are both legitimate goals. But the codebase already contains the mature version of this pattern. The MCP inbox poller distinguishes the failure classes explicitly and dead-letters with a logged error report:

```python
# src/lingtai/core/mcp/inbox.py:418-428
try:
    raw = entry.read_text(encoding="utf-8")
except OSError as e:
    _dead_letter(entry, f"read failed: {e}")
    continue

try:
    event = json.loads(raw)
except json.JSONDecodeError as e:
    _dead_letter(entry, f"invalid JSON: {e}")
    continue
```

with `_dead_letter` (`src/lingtai/core/mcp/inbox.py:348-364`) moving the bad file into a `.dead/` subdirectory alongside a `*.error.json` report and emitting `log.warning`. Notably, the Phase 2 pseudo-outbox path in `mail.py` itself is also more careful than Phase 1 — it at least uses `continue` (line 305-306) so an unreadable outbox message is retried next cycle rather than permanently skipped, and it logs claim/restore failures via `logger.warning`/`logger.exception`.

Better looks like: transient `OSError` → log a warning and leave the entry out of `_seen` so the next 0.5s cycle retries it, with a bounded retry count so a persistently unreadable entry eventually gets dead-lettered instead of retried forever; `JSONDecodeError` → dead-letter immediately (the sender writes `message.json` atomically via `tmp → os.replace` at `mail.py:199-206`, so a parse error means the file is genuinely malformed, not half-written); and `on_message` exceptions → logged separately so a consumer bug is never misdiagnosed as a transport failure.

## Current behavior

- **File/function:** `src/lingtai_kernel/services/mail.py`, `FilesystemMailService.listen()`, inner `_poll_loop()`, Phase 1 (lines 234-248).
- A `json.JSONDecodeError` or `OSError` from `msg_file.read_text()` / `json.loads()` — or an `OSError` escaping the user-supplied `on_message` callback — is caught and discarded with `pass` (line 246-247).
- `self._seen.add(entry.name)` runs unconditionally after the `try` block (line 248), so the entry is never revisited. Because `_seen` is in-memory, the message *would* be redelivered after a process restart, making the loss intermittent and even harder to diagnose.
- No log line is emitted at any level. `mail.py` already has a module logger (`logger = logging.getLogger(__name__)`, line 26) that is used by the Phase 2 pseudo-outbox path but not by Phase 1.
- No dead-letter artifact is produced, unlike the MCP inbox path (`src/lingtai/core/mcp/inbox.py:348-364`).
- Additionally, Phase 1 does not skip dot-directories, so any future `.dead` directory under `inbox/` would itself be iterated (it is currently only skipped implicitly by ending up in `_seen`).

All changes are in `src/lingtai_kernel/services/mail.py` unless noted.

1. **Add a module-level constant and per-instance retry tracking.**
   - Near the top of the file: `_MAX_READ_RETRIES = 10` (at 0.5s per cycle this gives ~5s of retries before giving up) and `_DEAD_DIRNAME = ".dead"`.
   - In `FilesystemMailService.__init__` (around line 116), add `self._read_failures: dict[str, int] = {}` alongside `self._seen`.

2. **Add a `_dead_letter_inbox_entry` method on `FilesystemMailService`.** Unlike the MCP poller (which dead-letters single files), mail inbox entries are *directories* (`<uuid>/message.json` + optional `attachments/`), so move the whole directory:

   ```python
   def _dead_letter_inbox_entry(self, entry: Path, error: str) -> None:
       """Move a bad inbox entry dir into inbox/.dead/ with an error report."""
       dead_dir = self._inbox_dir / _DEAD_DIRNAME
       try:
           dead_dir.mkdir(parents=True, exist_ok=True)
           target = dead_dir / entry.name
           shutil.move(str(entry), str(target))
           (target / "error.json").write_text(
               json.dumps(
                   {"error": error, "timestamp": datetime.now(timezone.utc).isoformat()},
                   ensure_ascii=False, indent=2,
               ),
               encoding="utf-8",
           )
           logger.warning("mail inbox: dead-lettered %s: %s", target, error)
       except OSError as e:
           logger.error(
               "mail inbox: failed to dead-letter %s: %s (original error: %s)",
               entry, e, error,
           )
       # Either way, stop re-processing this entry.
       self._seen.add(entry.name)
       self._read_failures.pop(entry.name, None)
   ```

   (`shutil` is already imported; add `datetime`/`timezone` imports if not present, or reuse an ISO-timestamp helper.)

3. **Rewrite the Phase 1 body** (lines 236-248) to separate the three failure classes:

   ```python
   for entry in self._inbox_dir.iterdir():
       if not entry.is_dir() or entry.name.startswith("."):
           continue  # skip files and .dead/
       if entry.name in self._seen:
           continue
       if not msg_file.is_file():
           continue  # not fully delivered yet; sender's os.replace pending

       try:
           raw = msg_file.read_text(encoding="utf-8")
       except OSError as e:
           n = self._read_failures.get(entry.name, 0) + 1
           self._read_failures[entry.name] = n
           if n >= _MAX_READ_RETRIES:
               self._dead_letter_inbox_entry(entry, f"read failed after {n} attempts: {e}")
           else:
               logger.warning(
                   "mail inbox: transient read failure on %s (attempt %d/%d): %s",
                   entry.name, n, _MAX_READ_RETRIES, e,
               )
           continue  # NOT added to _seen — retried next cycle

       try:
           payload = json.loads(raw)
       except json.JSONDecodeError as e:
           self._dead_letter_inbox_entry(entry, f"invalid JSON: {e}")
           continue

       self._read_failures.pop(entry.name, None)
       self._seen.add(entry.name)  # mark seen BEFORE dispatch: consumer errors must not redeliver
       try:
           on_message(payload)
       except Exception:
           logger.exception("mail inbox: on_message handler failed for %s", entry.name)
   ```

   Key semantics: transient read errors retry with a bound; malformed JSON dead-letters immediately (safe because `send()` writes atomically via `tmp → os.replace`, `mail.py:199-206`); handler exceptions are logged but do not cause redelivery (preserving today's at-most-once dispatch semantics) and no longer masquerade as transport errors.

4. **Also skip dotted directory names in the startup snapshot** in `listen()` (lines 224-227), so `.dead` is not pointlessly added to `_seen`:

   ```python
   if entry.is_dir() and not entry.name.startswith("."):
       self._seen.add(entry.name)
   ```

5. **Back-compat / migration:** none required. The `.dead/` directory is created lazily inside the existing `inbox/`; anything that lists inbox entries should already tolerate (or now explicitly skips) dotted names. No schema or config changes. The `error.json` filename inside the moved entry directory cannot collide with message content because entries only ever contain `message.json` and `attachments/`.

6. **Optional follow-up (can be a separate issue):** extract a shared dead-letter helper used by both `src/lingtai/core/mcp/inbox.py` and `mail.py`, since the two now implement sibling patterns. Not required for the fix; keeping the change local to `mail.py` minimizes risk.

## Risks and alternatives

- **Retry loop on a persistently unreadable entry.** Mitigated by `_MAX_READ_RETRIES`: after ~5 seconds of failures the entry is dead-lettered and marked seen. Without the bound, a permissions-denied file would log a warning twice per second forever.
- **Dead-lettering a transiently unreadable file too eagerly.** The alternative — retry forever, never dead-letter on `OSError` — risks unbounded log spam and an inbox that silently accumulates stuck entries. Ten attempts is a reasonable middle ground and the threshold is a named constant.
- **`.dead/` growth.** Dead-lettered entries accumulate indefinitely, same as the MCP inbox's `.dead/`. Acceptable: these should be rare, and operators need the artifacts for diagnosis. A retention sweep can be added later if needed.
- **Behavior change for `on_message` exceptions.** Today an `OSError` from the handler silently drops the message, and any *other* exception type propagates into the outer `except OSError: pass` at line 254-255 — where a non-OSError would actually kill the poll thread. The plan's explicit `try/except Exception` around dispatch fixes both, but consumers relying on handler exceptions crashing the thread (unlikely; nothing in-repo does) would see a change.
- **Alternative: persist `_seen` or use claim-by-rename like Phase 2.** Renaming inbox entries to a `processing/` dir on claim would give crash-safe exactly-once semantics, but is a much larger redesign of the delivery contract (senders and tests assume messages live under `inbox/<uuid>/`). This plan keeps the existing contract and fixes the concrete data-loss bug.
- **Alternative: log only, keep marking seen on all errors.** Simpler, but still permanently drops messages on transient `OSError`, which is the core of the bug; rejected.

## Testing plan

Add to `tests/test_services_mail.py` (which already exercises `FilesystemMailService`; `tests/test_filesystem_mail.py` also touches this class):

1. `test_listen_retries_transient_read_error_then_delivers` — deliver a valid message, monkeypatch `Path.read_text` (or the message file's permissions) to raise `OSError` for the first two poll cycles then succeed; assert `on_message` is eventually called with the payload, the entry is not in `inbox/.dead/`, and a warning was logged (`caplog`).
2. `test_listen_dead_letters_after_max_read_retries` — make `read_text` fail persistently; assert that after `_MAX_READ_RETRIES` cycles the entry directory has moved to `inbox/.dead/<uuid>/` with an `error.json` containing the error string, `on_message` was never called, and the poller no longer touches the entry (failure counter cleared, no further log spam).
3. `test_listen_dead_letters_malformed_json_immediately` — write an inbox entry whose `message.json` contains `{not json`; assert it is moved to `inbox/.dead/` on the first cycle with `"invalid JSON"` in `error.json`, `on_message` is not called, and `logger.warning` fired.
4. `test_listen_on_message_exception_is_logged_not_dropped_silently` — handler raises `RuntimeError`; assert the poll thread stays alive (a subsequent message is still delivered), the entry is marked seen (no redelivery), and `logger.exception` output appears in `caplog`.
5. `test_listen_skips_dead_dir` — pre-create `inbox/.dead/whatever/message.json` with valid JSON; assert `on_message` is never called for it and it is not added to `_seen`.
6. `test_valid_messages_still_delivered_once` — regression guard: normal delivery still dispatches exactly once and dedupes via `_seen` across cycles.

huangzesen:
Fixes #732

## Root cause

`PosixFilesystemMailAdapter._poll_own_inbox_slice()` (the Phase 1 own-inbox
poller, formerly `FilesystemMailService.listen()` in
`src/lingtai_kernel/services/mail.py`, moved to the adapter layer by the
Ports & Adapters refactor) collapsed three distinct failure classes into one
silent drop:

```python
if msg_file.is_file():
    try:
        payload = json.loads(msg_file.read_text(encoding="utf-8"))
        on_message(payload)
    except (json.JSONDecodeError, OSError):
        pass
    self._seen.add(entry.name)
```

- A transient `OSError` (permissions blip, network-filesystem hiccup on one
  0.5 s cycle) dropped a perfectly intact message **forever** — the entry was
  marked seen with no retry and no log line.
- A genuinely malformed `message.json` vanished into `_seen` with no
  dead-letter artifact and no log line.
- `on_message()` ran inside the same `try`: a consumer `OSError` was
  misdiagnosed as a transport failure, and any other handler exception
  escaped into the loop's `except OSError` and **killed the poll thread**.

This mirrors the bug report exactly; the defect is still reproducible on
current main (verified at HEAD `df2b523d`). The earlier attempt, PR #786
(`fix(mail): stop silently dropping unreadable inbox messages`), was closed
unmerged and predates the adapter-layer move.

## Fix (all in `src/lingtai/adapters/posix/mail.py`)

Follows the mature dead-letter pattern the MCP inbox poller already uses
(`src/lingtai/core/mcp/inbox.py`), applied to directory-based inbox entries:

- **Transient `OSError` on read** → log a warning and skip **without** marking
  `_seen`, so the next 0.5 s cycle retries the still-intact file. A bounded
  retry counter (`_MAX_READ_RETRIES = 10`, ~5 s) dead-letters a persistently
  unreadable entry instead of retrying forever.
- **`JSONDecodeError`** → dead-letter the whole entry directory into
  `inbox/.dead/<uuid>/` with an `error.json` report and a logged warning.
  Safe to treat as permanent because `send()` writes `message.json`
  atomically via `tmp → os.replace`.
- **`on_message` dispatch** moved out of the read `try` and wrapped
  separately: the entry is marked seen **before** dispatch (at-most-once
  preserved), and a handler exception is logged via `logger.exception`
  instead of dropping the message or killing the poll thread.
- **Dotted directories** (the new `.dead/` store) are skipped in both the
  startup snapshot and the poll loop, so dead-lettered entries are never
  re-dispatched.

## Validation

New regression tests in `tests/test_filesystem_mail.py`:
`test_listen_retries_transient_read_error_then_delivers`,
`test_listen_dead_letters_after_max_read_retries`,
`test_listen_dead_letters_malformed_json_immediately`,
`test_listen_on_message_exception_is_logged_not_silent`, and
`test_listen_skips_dead_dir`.

Failing-first run (new tests against the pre-fix `mail.py`):

```
$ pytest tests/test_filesystem_mail.py -q -k "transient or dead_letters or on_message_exception or skips_dead"
5 failed, 27 deselected in 21.17s
```

After the fix:

```
$ pytest tests/test_filesystem_mail.py -q -k "transient or dead_letters or on_message_exception or skips_dead"
5 passed, 27 deselected in 5.91s
```

Full touched-module suites:

```
$ pytest tests/test_filesystem_mail.py -q
32 passed in 17.31s
$ pytest tests/test_filesystem_mail.py tests/test_services_mail.py tests/test_mail_transport.py -q
56 passed in 18.48s
```

`tests/test_three_agent_email.py` shows 5 pre-existing collection errors in
this minimal venv (missing third-party deps in the agent boot path); they are
identical on pristine `main` and unrelated to this change.

`git diff --check` is clean.

### DIFF
--- a/src/lingtai/adapters/posix/mail.py
+++ b/src/lingtai/adapters/posix/mail.py
@@ -34,6 +34,14 @@ from lingtai.kernel.services.mail import _new_mailbox_id
 
 logger = logging.getLogger(__name__)
 
+# A ``message.json`` that fails to read (transient OSError: permissions blip,
+# network-filesystem hiccup) is retried for this many poll cycles before being
+# dead-lettered. At the 0.5 s poll cadence this is ~5 s of retries.
+_MAX_READ_RETRIES = 10
+# Dead-letter store for permanently bad inbox entries, mirroring the MCP inbox
+# poller's ``.dead/`` convention.
+_DEAD_DIRNAME = ".dead"
+
 
 def _presence_store_for(recipient_dir: Path):
     """Build a target-bound POSIX presence adapter for a recipient directory."""
@@ -78,6 +86,9 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         self._poll_thread: threading.Thread | None = None
         self._poll_stop = threading.Event()
         self._seen: set[str] = set()
+        # Consecutive read-failure counts per inbox entry name, for bounded
+        # retry of transiently unreadable ``message.json`` files.
+        self._read_failures: dict[str, int] = {}
         # Own-inbox scans can be expensive on large/slow external-volume
         # mailboxes. Keep iterator progress across poll ticks so Phase 2 can
         # be sliced instead of restarting from the first historical entry.
@@ -198,10 +209,12 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         re-delivered.  New directories that appear with a ``message.json``
         trigger *on_message*.
         """
-        # Snapshot existing inbox entries so we don't re-notify
+        # Snapshot existing inbox entries so we don't re-notify. Dotted
+        # directories (e.g. the ``.dead/`` dead-letter store) are not inbox
+        # entries and are excluded from the snapshot.
         if self._inbox_dir.is_dir():
             for entry in self._inbox_dir.iterdir():
-                if entry.is_dir():
+                if entry.is_dir() and not entry.name.startswith("."):
                     self._seen.add(entry.name)
 
         self._poll_stop.clear()
@@ -294,21 +307,59 @@ class PosixFilesystemMailAdapter(MailTransportPort):
 
                 visited += 1
                 # _seen only holds handled directory names or pseudo-claim
-                # UUIDs, so skip before the stat.
+                # UUIDs, so skip before the stat. Dotted directories (e.g. the
+                # ``.dead/`` dead-letter store) are never inbox entries.
                 if entry.name in self._seen:
                     skipped_seen += 1
                     continue
-                if not entry.is_dir():
+                if not entry.is_dir() or entry.name.startswith("."):
                     continue
                 msg_file = entry / "message.json"
                 if msg_file.is_file():
+                    # Transient read failures are retried (NOT marked seen),
+                    # so a one-cycle OSError on a network filesystem cannot
+                    # permanently drop an intact message. Permanent failures
+                    # are dead-lettered with a logged error report.
                     try:
-                        payload = json.loads(msg_file.read_text(encoding="utf-8"))
+                        raw = msg_file.read_text(encoding="utf-8")
+                    except OSError as e:
+                        n = self._read_failures.get(entry.name, 0) + 1
+                        self._read_failures[entry.name] = n
+                        if n >= _MAX_READ_RETRIES:
+                            self._dead_letter_inbox_entry(
+                                entry, f"read failed after {n} attempts: {e}"
+                            )
+                        else:
+                            logger.warning(
+                                "mail inbox: transient read failure on %s (attempt %d/%d): %s",
+                                entry.name,
+                                n,
+                                _MAX_READ_RETRIES,
+                                e,
+                            )
+                        continue
+
+                    try:
+                        payload = json.loads(raw)
+                    except json.JSONDecodeError as e:
+                        # `send()` writes message.json atomically (tmp -> rename),
+                        # so a parse error means genuinely malformed content;
+                        # dead-letter immediately.
+                        self._dead_letter_inbox_entry(entry, f"invalid JSON: {e}")
+                        continue
+
+                    self._read_failures.pop(entry.name, None)
+                    # Mark seen BEFORE dispatch: a consumer error must not
+                    # redeliver (at-most-once), and must not be misdiagnosed
+                    # as a transport failure.
+                    self._seen.add(entry.name)
+                    try:
                         on_message(payload)
                         dispatched += 1
-                    except (json.JSONDecodeError, OSError):
-                        pass
-                    self._seen.add(entry.name)
+                    except Exception:
+                        logger.exception(
+                            "mail inbox: on_message handler failed for %s", entry.name
+                        )
 
         except OSError:
             self._reset_own_inbox_iter()
@@ -324,6 +375,44 @@ class PosixFilesystemMailAdapter(MailTransportPort):
             )
 
         return visited > 0
+
+    def _dead_letter_inbox_entry(self, entry: Path, error: str) -> None:
+        """Move a bad inbox entry directory into ``inbox/.dead/`` with a report.
+
+        Mail inbox entries are directories (``<uuid>/message.json`` plus
+        optional ``attachments/``), so unlike the single-file MCP inbox
+        dead-letter path the whole entry directory is moved. An ``error.json``
+        report and a warning log line are left behind for operators.
+        """
+        from datetime import datetime, timezone
+
+        dead_dir = self._inbox_dir / _DEAD_DIRNAME
+        try:
+            dead_dir.mkdir(parents=True, exist_ok=True)
+            target = dead_dir / entry.name
+            shutil.move(str(entry), str(target))
+            (target / "error.json").write_text(
+                json.dumps(
+                    {
+                        "error": error,
+                        "timestamp": datetime.now(timezone.utc).isoformat(),
+                    },
+                    ensure_ascii=False,
+                    indent=2,
+                ),
+                encoding="utf-8",
+            )
+            logger.warning("mail inbox: dead-lettered %s: %s", target, error)
+        except OSError as e:
+            logger.error(
+                "mail inbox: failed to dead-letter %s: %s (original error: %s)",
+                entry,
+                e,
+                error,
+            )
+        # Either way, stop re-processing this entry.
+        self._seen.add(entry.name)
+        self._read_failures.pop(entry.name, None)
 
     def _reset_own_inbox_iter(self) -> None:
         iterator = self._own_inbox_iter
--- a/tests/test_filesystem_mail.py
+++ b/tests/test_filesystem_mail.py
@@ -967,3 +967,222 @@ def test_mail_poll_slow_phase_telemetry_is_body_free(tmp_path, caplog):
     assert "phase=own_inbox_slice" in caplog.text
     assert "agent=agent01" in caplog.text
     assert "do-not-log-body" not in caplog.text
+
+
+def _write_message(msg_dir, payload):
+    """Write an inbox message atomically (tmp -> os.replace), exactly like
+    ``PosixFilesystemMailAdapter.send()`` does, so tests never expose a
+    half-written ``message.json`` to the poller."""
+    import os
+
+    msg_dir.mkdir(parents=True, exist_ok=True)
+    tmp = msg_dir / "message.json.tmp"
+    tmp.write_text(json.dumps(payload), encoding="utf-8")
+    os.replace(str(tmp), str(msg_dir / "message.json"))
+
+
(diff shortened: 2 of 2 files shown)


huangzesen:

PR #1282, opened 2026-08-09
fix(init-schema): reject bool for any int-accepting field in _check_type (#737)

### LINKED ISSUE #737: _check_type accepts bool for tuple-typed int fields — `context_limit: true` validates and becomes a 1-token context window
## Summary

`_check_type()` in `src/lingtai/init_schema.py` only rejects booleans when the expected type is *exactly* `int` or *exactly* `(int, float)`. Fields typed with any other tuple containing `int` — notably `manifest.context_limit`, typed `(int, type(None))` — silently accept `True`/`False` because `bool` is a subclass of `int` in Python. Verified against the current tree: `validate_init` passes `context_limit: true` with zero warnings, `build_agent_config` hydrates it verbatim, and every downstream consumer then computes against a context window of `True == 1` token.

## Narrative

An operator hand-editing an agent's `init.json` (or a tool generating one) writes something like:

```json
{
  "manifest": {
    "llm": { "provider": "anthropic", "model": "..." },
    "context_limit": true
  }
}
```

Maybe they meant "yes, enforce a context limit", maybe a templating bug substituted a boolean, maybe a YAML-to-JSON conversion mangled a value. Whatever the cause, the natural expectation is that schema validation catches it — the whole point of `validate_init` is to reject wrong types on known fields with a clear message. It does not. Validation passes cleanly, the agent boots, and things get strange: `src/lingtai_kernel/session.py:421` does

```python
ctx_window = self._config.context_limit or self._chat.context_window()
```

`True` is truthy, so the model-default fallback never fires and `ctx_window` is `True`. All subsequent arithmetic treats it as the integer `1`. `src/lingtai_kernel/meta_block.py:2359` makes this explicit:

```python
config_limit = int(getattr(getattr(agent, "_config", None), "context_limit", 0) or 0)
```

— a 1-token context limit. Context-usage percentages become astronomically over 100%, and compaction/molt pressure logic behaves as if the agent is permanently, catastrophically out of context. Nothing in the failure points back at the one-character mistake in `init.json`. (The `false` case is quieter but equally wrong: it's falsy, so it silently behaves like `null` instead of being rejected.)

How did the code get here? The author of `_check_type` clearly knew about the bool/int subclass trap — there is a comment saying exactly that at `src/lingtai/init_schema.py:425-427`:

```python
# bool is a subclass of int in Python — reject bools for numeric fields
if isinstance(value, bool) and expected_type in (int, (int, float)):
    raise ValueError(f"{path}: expected number, got bool")
```

But the guard enumerates specific `expected_type` values instead of asking the general question "is `int` among the accepted types?". `(int, type(None))` — the type of `manifest.context_limit` (`src/lingtai/init_schema.py:85`) and `manifest.llm.compact_threshold` (line 326) — is not in the enumeration, so bools sail through the subsequent `isinstance(value, expected_type)` check.

The codebase has already paid for this gap twice, in the form of hand-rolled per-field bool guards inside `validate_init`. Interestingly, an audit shows the current guard set is exactly inverted from where the risk is:

- `manifest.summarize_notification_threshold` (guard at lines 284-289) and `manifest.cache_miss_budget` (guard at lines 295-302) are typed plain `int` in `MANIFEST_OPTIONAL`, so `_check_type` already rejects bools for them before those guards ever run — verified: passing `cache_miss_budget: true` raises `"expected number, got bool"` from `_check_type`, not the hand-rolled message. **These two guards are dead code.**
- `manifest.llm.compact_threshold` (guard at lines 329-331) is typed `(int, type(None))` and genuinely needs its guard today.
- `manifest.context_limit` — the same tuple type — has **no** guard and is the live hole.

Better looks like: fix the general check once in `_check_type` so that a bool is rejected whenever `int` is among the expected types (unless `bool` itself is explicitly listed), then delete all three hand-rolled bool guards. One rule, enforced in one place, covering every current and future int-bearing field.

## Current behavior

- `src/lingtai/init_schema.py:419-443`, `_check_type()`: the bool rejection at line 426 uses `expected_type in (int, (int, float))`, i.e. exact-match against two specific type specs. For any other tuple containing `int`, a bool value then passes `isinstance(value, expected_type)` (bool is an int subclass) and no error is raised.
- `src/lingtai/init_schema.py:85`: `"context_limit": (int, type(None))` in `MANIFEST_OPTIONAL` — so `validate_init({... "context_limit": true ...})` succeeds with no warnings (reproduced on current main).
- `src/lingtai/agent.py:48`, `build_agent_config()`: `context_limit=manifest.get("context_limit", defaults.context_limit)` hydrates the bool verbatim into `AgentConfig.context_limit` (`src/lingtai_kernel/config.py:91`, annotated `int | None` — the annotation is not enforced at runtime).
- Downstream consumers treat `True` as a 1-token window: `src/lingtai_kernel/session.py:421` and `:577`, `src/lingtai_kernel/meta_block.py:1507` and `:2359`, `src/lingtai_kernel/base_agent/identity.py:203`.
- Redundant/asymmetric guards in `validate_init`: lines 284-289 (`summarize_notification_threshold`, dead), 295-302 (`cache_miss_budget`, dead), 329-331 (`compact_threshold`, live and load-bearing).
- Related same-class gap: `src/lingtai_kernel/presets.py:306-309` validates preset `manifest.llm.context_limit` with `isinstance(ctx_limit, int)`, which a bool also passes.

1. **Fix the general guard in `_check_type`** (`src/lingtai/init_schema.py`). Normalize `expected_type` to a tuple and reject bools whenever `int` is accepted and `bool` is not explicitly listed:

   ```python
   def _check_type(value, expected_type, path):
       expected = expected_type if isinstance(expected_type, tuple) else (expected_type,)
       # bool is a subclass of int in Python — reject bools for numeric fields
       # unless bool is explicitly one of the accepted types.
       if isinstance(value, bool) and int in expected and bool not in expected:
           raise ValueError(f"{path}: expected number, got bool")
       ...
   ```

   Fields typed plain `bool` (`streaming`, `time_awareness`, `timezone_awareness`) are unaffected: their spec is `(bool,)`, which contains no `int`. Fields typed `int` or `(int, float)` behave exactly as before. Fields typed `(int, type(None))` — `context_limit`, `compact_threshold` — now reject bools.

2. **Delete the three hand-rolled bool guards in `validate_init`**, keeping the range checks that follow them:
   - `summarize_notification_threshold`: remove the `isinstance(..., bool)` branch (lines 286-289); keep the `< 0` check.
   - `cache_miss_budget`: remove the `isinstance(..., bool)` branch (lines 299-302) and its now-stale comment; keep the `<= 0` check.
   - `compact_threshold`: remove the `isinstance(..., bool)` branch (line 330-331); keep the `<= 0` check. Also update the schema comment at lines 93-96 (`cache_miss_budget`) which currently claims the bool rejection happens "explicitly in validate_init below".

3. **Update existing test expectations if needed.** `tests/test_init_schema.py` has `test_summarize_notification_threshold_rejects_bool` (line 96) and `test_cache_miss_budget_rejects_bool` (line 129). The raised message changes from the bespoke per-field text to `_check_type`'s `"{path}: expected number, got bool"`. Adjust `pytest.raises(..., match=...)` patterns if they pin the old wording; the field path remains in the message either way.

4. **(Optional hardening, same class of bug)** In `src/lingtai_kernel/presets.py` (~line 306), tighten the preset `context_limit` check to `isinstance(ctx_limit, int) and not isinstance(ctx_limit, bool)` so a preset file can't smuggle the same bool in through the preset path. This can be a follow-up if scope should stay minimal.

5. **Back-compat**: any existing `init.json` carrying `context_limit: true/false` was already misbehaving at runtime (1-token window or silent null); after this change it fails validation at load with an explicit `manifest.context_limit: expected number, got bool`. That is the desired outcome — it converts an invisible runtime pathology into a clear config error. No migration is needed; operators fix the field to an integer or `null`.

## Risks and alternatives

- **Risk: newly rejecting previously "valid" files.** An agent that has been running with `context_limit: true` will now refuse to validate. This is intentional and strictly better than the current behavior (permanent compaction/molt pressure against a 1-token window), but it is a behavior change worth a line in the changelog.
- **Risk: over-rejecting.** The `bool not in expected` escape hatch keeps the door open for any future field that legitimately accepts `bool | int`. No current schema entry does, so nothing changes today.
- **Alternative considered: add a fourth hand-rolled guard for `context_limit`.** Rejected — it perpetuates the whack-a-mole pattern that produced two dead guards and one missed field, and the next tuple-typed int field would hit the same hole. Fixing `_check_type` closes the class of bug.
- **Alternative considered: clamp/coerce at hydration in `build_agent_config` (`src/lingtai/agent.py`).** Rejected — hydration comments explicitly state values are "hydrated verbatim" because validation is the schema layer's job; coercing `True → None` there would hide a real config error instead of surfacing it.

## Testing plan

Add to `tests/test_init_schema.py`:

- `test_context_limit_rejects_bool` — a minimal valid init dict plus `manifest.context_limit: True` raises `ValueError` matching `manifest\.context_limit.*bool`; repeat with `False`.
- `test_context_limit_accepts_int_and_null` — `context_limit: 200000` and `context_limit: None` both validate with no error (regression guard for the tuple-type happy path).
- `test_compact_threshold_rejects_bool_via_check_type` — `manifest.llm.compact_threshold: True` still raises after the hand-rolled guard is removed.
- `test_check_type_bool_allowed_when_listed` — direct unit test of `_check_type(True, (int, bool), "x")` not raising, documenting the escape hatch.
- Update `test_summarize_notification_threshold_rejects_bool` and `test_cache_miss_budget_rejects_bool` match patterns for the unified `_check_type` message; keep them as regression tests that bools are still rejected for plain-`int` fields.

Add to `tests/test_agent_config_hydration.py` (or nearest hydration test module):

- `test_build_agent_config_context_limit_none_defaults` — asserts `build_agent_config({"llm": {...}}, max_rpm=...)` leaves `context_limit` at the `AgentConfig` default, confirming the schema layer is the sole gate and hydration semantics are unchanged.

huangzesen:
## Summary

- Fix `_check_type` in `src/lingtai/init_schema.py` to reject a `bool` whenever `int` is among the accepted types (unless `bool` is explicitly listed), instead of enumerating only the exact specs `int` and `(int, float)`.
- Delete the three now-redundant hand-rolled per-field bool guards in `validate_init`: `summarize_notification_threshold` and `cache_miss_budget` (both plain `int` — their guards were already dead code, `_check_type` rejected the bool first) and `compact_threshold` (`(int, None)` — its guard was live and is now covered by the general rule). The `< 0` / `<= 0` range checks are kept.
- Add regression tests for `manifest.context_limit`, `manifest.llm.compact_threshold`, and the `_check_type` general rule + escape hatch.
- Update the stale `cache_miss_budget` schema comment that claimed bool rejection happened "explicitly in validate_init below".

### Root cause

`bool` subclasses `int` in Python, so `isinstance(True, int)` is `True`. The guard rejected bools only for `expected_type in (int, (int, float))`. `manifest.context_limit` is typed `(int, type(None))` (`init_schema.py:89`), which was not in that enumeration, so `context_limit: true` passed `validate_init` with zero warnings, hydrated verbatim into `AgentConfig.context_limit`, and made `ctx_window = self._config.context_limit or ...` (`session.py:421`) evaluate to `True` — a 1-token context window. Context-usage percentages then ran astronomically over 100% with nothing pointing back at the one-character init.json mistake. Asking the general question "is `int` accepted?" closes the whole class of bug (the same gap had already produced two dead guards and one missed field).

### Non-goals

- `src/lingtai/kernel/presets.py` (`isinstance(ctx_limit, int)` at ~:356) has the same class of gap on the preset path. It is a separate module/consumer and the issue explicitly flagged it as optional follow-up; left out to keep this PR one concern.

### Back-compat

An existing init.json carrying `context_limit: true/false` was already misbehaving at runtime (1-token window, or silently null for `false`). It now fails validation loudly with `manifest.context_limit: expected number, got bool` — an invisible runtime pathology becomes a clear config error. No migration needed; operators set an integer or `null`.

## Validation

Failing-first run (fix reverted via `git stash push src/lingtai/init_schema.py`, new tests run against pre-fix code):

```
FAILED tests/test_init_schema.py::test_context_limit_rejects_bool[True]
FAILED tests/test_init_schema.py::test_context_limit_rejects_bool[False]
E   Failed: DID NOT RAISE ValueError
```

Post-fix, full module suite:

```
$ python -m pytest tests/test_init_schema.py -q
125 passed in 0.15s
```

Broader affected modules (schema, hydration, presets, agent config):

```
$ python -m pytest tests/ -q -k "hydrat or build_agent_config or preset or init_schema or agent_config"
496 passed, 4 failed (the 4 failures are pre-existing on pristine main in this env: test_daemon_preset_capabilities, test_feishu_automatic_task_cards, test_preset_runtime_model_docs x2)
```

Whitespace:

```
$ git diff --check
(no output)
```

Fixes #737

### DIFF
--- a/src/lingtai/init_schema.py
+++ b/src/lingtai/init_schema.py
@@ -106,8 +106,9 @@ MANIFEST_OPTIONAL: dict[str, type | tuple[type, ...]] = {
     "aed_timeout": (int, float),
     # Soft per-molt/session cache-miss token budget. Positive int; default
     # 1_000_000 lives in AgentConfig.cache_miss_budget. The range check
-    # (reject bool and <= 0) is enforced explicitly in validate_init below —
-    # the (int) type here only rejects non-int types like str/float/None.
+    # (<= 0) is enforced explicitly in validate_init below; the (int) type
+    # here rejects non-int types like str/float/None (bool is rejected by
+    # _check_type's general rule, since bool subclasses int).
     "cache_miss_budget": int,
     "admin": dict,
     "streaming": bool,
@@ -365,10 +366,6 @@ def validate_init(data: dict) -> list[str]:
 
     if "summarize_notification_threshold" in manifest:
         summarize_threshold = manifest["summarize_notification_threshold"]
-        if isinstance(summarize_threshold, bool):
-            raise ValueError(
-                "manifest.summarize_notification_threshold: expected non-negative int, got bool"
-            )
         if summarize_threshold < 0:
             raise ValueError(
                 "manifest.summarize_notification_threshold: expected non-negative int"
@@ -376,12 +373,6 @@ def validate_init(data: dict) -> list[str]:
 
     if "cache_miss_budget" in manifest:
         cache_miss_budget = manifest["cache_miss_budget"]
-        # bool is an int subclass — reject it explicitly, then require > 0.
-        # (_optional_keys already rejected non-int types like str/float/None.)
-        if isinstance(cache_miss_budget, bool):
-            raise ValueError(
-                "manifest.cache_miss_budget: expected positive int, got bool"
-            )
         if cache_miss_budget <= 0:
             raise ValueError(
                 "manifest.cache_miss_budget: expected positive int (> 0)"
@@ -399,8 +390,6 @@ def validate_init(data: dict) -> list[str]:
     _optional_keys(llm, LLM_OPTIONAL, prefix="manifest.llm")
     if "compact_threshold" in llm:
         compact_threshold = llm["compact_threshold"]
-        if isinstance(compact_threshold, bool):
-            raise ValueError("manifest.llm.compact_threshold: expected int | null, got bool")
         if isinstance(compact_threshold, int) and compact_threshold <= 0:
             raise ValueError(
                 "manifest.llm.compact_threshold: expected positive int or null"
@@ -541,8 +530,10 @@ def _check_type(
     path: str,
 ) -> None:
     """Validate a single value's type."""
+    expected = expected_type if isinstance(expected_type, tuple) else (expected_type,)
     # bool is a subclass of int in Python — reject bools for numeric fields
-    if isinstance(value, bool) and expected_type in (int, (int, float)):
+    # unless bool is explicitly one of the accepted types.
+    if isinstance(value, bool) and int in expected and bool not in expected:
         raise ValueError(f"{path}: expected number, got bool")
 
     if not isinstance(value, expected_type):
--- a/tests/test_init_schema.py
+++ b/tests/test_init_schema.py
@@ -1,6 +1,6 @@
 import json
 import pytest
-from lingtai.init_schema import validate_init
+from lingtai.init_schema import _check_type, validate_init
 
 
 def _valid_init() -> dict:
@@ -96,7 +96,7 @@ def test_summarize_notification_threshold_rejects_negative():
 def test_summarize_notification_threshold_rejects_bool():
     data = _valid_init()
     data["manifest"]["summarize_notification_threshold"] = True
-    with pytest.raises(ValueError, match="summarize_notification_threshold"):
+    with pytest.raises(ValueError, match="summarize_notification_threshold.*bool"):
         validate_init(data)
 
 
@@ -128,10 +128,39 @@ def test_cache_miss_budget_rejects_negative():
 
 def test_cache_miss_budget_rejects_bool():
     data = _valid_init()
-    # bool is an int subclass in Python — must be rejected explicitly.
+    # bool is an int subclass in Python — rejected by _check_type's general
+    # rule for any int-accepting field (issue #737).
     data["manifest"]["cache_miss_budget"] = True
-    with pytest.raises(ValueError, match="manifest.cache_miss_budget"):
+    with pytest.raises(ValueError, match="manifest.cache_miss_budget.*bool"):
         validate_init(data)
+
+
+@pytest.mark.parametrize("bad_value", [True, False])
+def test_context_limit_rejects_bool(bad_value):
+    data = _valid_init()
+    data["manifest"]["context_limit"] = bad_value
+    with pytest.raises(ValueError, match=r"manifest\.context_limit.*bool"):
+        validate_init(data)
+
+
+@pytest.mark.parametrize("good_value", [200_000, None])
+def test_context_limit_accepts_int_and_null(good_value):
+    data = _valid_init()
+    data["manifest"]["context_limit"] = good_value
+    assert validate_init(data) == []
+
+
+def test_compact_threshold_rejects_bool_via_check_type():
+    data = _valid_init()
+    data["manifest"]["llm"]["compact_threshold"] = True
+    with pytest.raises(ValueError, match=r"manifest\.llm\.compact_threshold.*bool"):
+        validate_init(data)
+
+
+def test_check_type_bool_allowed_when_listed():
+    # Escape hatch: a field that explicitly accepts bool | int is not rejected.
+    _check_type(True, (int, bool), "x")  # should not raise
+    _check_type(3, (int, bool), "x")  # should not raise
 
 
 @pytest.mark.parametrize(


huangzesen:

PR #1283, opened 2026-08-09
fix(mcp): suppress httpx request logging and redact Telegram bot credentials

huangzesen:
Fixes https://github.com/Lingtai-AI/lingtai/issues/812

## Problem

The curated MCP stdio entrypoint (`lingtai/mcp_servers/_entrypoint.py`) configures root logging at INFO. HTTP clients (`httpx`/`httpcore`) emit one INFO `HTTP Request: <METHOD> <url> ...` line per request. Providers such as Telegram embed the live bot token in the URL path (`https://api.telegram.org/bot<TOKEN>/<method>`), so at startup (e.g. `getMe`) the live token is written to the MCP process stderr and can be persisted into refresh-watcher event stderr tails — a credential exposure, not just noise.

## Fix (defense in depth at the shared stderr boundary)

1. **Suppress sensitive client request logging.** `httpx` and `httpcore` loggers are pinned to WARNING in `run_stdio_server_main()` so token-bearing INFO request lines never reach stderr. This protects every curated MCP server, not just Telegram.
2. **Redact before stderr.** A Telegram-credential redaction `logging.Filter` is attached to the root logger and every root handler (same pattern as the feishu addon's `_lark_credential_filter`), rewriting `https://api.telegram.org/bot<TOKEN>/...` URLs and bare Telegram-token-shaped strings to `[REDACTED]` before any handler renders them. Handler-level filters are required because ancestor logger filters are not consulted for records emitted by descendant loggers.

`configure_stdio_logging()` was extracted from `run_stdio_server_main()` so the setup is directly testable.

## Tests

- **New `tests/test_mcp_server_credential_logging.py` (6 tests, all pass):** filter redacts Bot API URLs and bare tokens; httpx INFO request lines are suppressed while WARNING lines are redacted; ordinary non-secret Telegram/MCP diagnostics stay visible; filter is attached exactly once (idempotent).
- **Real-httpx reproduction** (validated locally with a fake token): with pre-fix root INFO + httpx INFO, `httpx.get(.../bot<FAKE>/getMe)` writes the token-bearing URL to stderr; after `configure_stdio_logging()` the same call emits no token and no INFO request line.
- `tests/test_telegram_rate_limit.py`, `tests/test_telegram_slash_commands.py`: 24 passed.
- `tests/test_mcp_server_scaffold.py`, `tests/test_feishu_stdio_logging.py`, `tests/test_services_logging.py`: identical pass/fail set before and after this change (remaining failures are pre-existing environment issues — missing optional dep `msal` for the imap import and 2 fallback event-id cases that fail identically on pristine main).

## Notes / follow-ups

- Persistence-layer sanitization of captured stderr tails inside the refresh watchers (issue suggestion #3) is intentionally **not** included: it is a larger cross-cutting change with many capture sites. Fixing the child-process stderr at the source removes the credential from what the watcher can capture.
- Users of affected versions should still rotate any token that may have been logged and follow the incident-response guidance in the issue.

### DIFF
--- a/src/lingtai/mcp_servers/_entrypoint.py
+++ b/src/lingtai/mcp_servers/_entrypoint.py
@@ -4,23 +4,93 @@ Every bundled MCP server's ``__main__.py`` ran the same three steps: configure
 INFO logging to **stderr** (so logs never corrupt the JSON-RPC stdout channel),
 ``asyncio.run(serve())``, and swallow ``KeyboardInterrupt`` on Ctrl-C. This is
 the single copy.
+
+Since #812, the stderr logging setup also applies defense in depth against
+credential leakage:
+
+1. ``httpx``/``httpcore`` are pinned to WARNING so the INFO ``HTTP Request:``
+   lines (whose URL embeds provider credentials such as the Telegram bot token
+   in the request path) never reach stderr.
+2. A redaction filter is attached to the root logger so any record that still
+   contains a Telegram bot token or a ``api.telegram.org/bot<TOKEN>/...`` URL
+   is rewritten to a fixed placeholder before any handler renders it.
 """
 from __future__ import annotations
 
 import asyncio
 import logging
+import re
 import sys
 from collections.abc import Awaitable, Callable
 
+# HTTP clients log one INFO ``HTTP Request: <METHOD> <url> ...`` line per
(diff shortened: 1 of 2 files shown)


huangzesen:

PR #1284, opened 2026-08-09
fix(init): close schema/hydration drift for honored manifest fields (#736)

### LINKED ISSUE #736: Schema/hydration drift: build_agent_config honors manifest fields (aed_timeout, max_aed_attempts, snapshot_interval, activeness) that validate_init flags as unknown and never type-checks
## Summary

`build_agent_config` in `src/lingtai/agent.py` honors four manifest fields — `activeness`, `snapshot_interval`, `aed_timeout`, and `max_aed_attempts` — that the init.json schema in `src/lingtai/init_schema.py` does not know about. `validate_init` therefore emits only an "unknown field" *warning* for them and never type-checks their values, so a malformed value (e.g. `"aed_timeout": "oops"`) passes validation, is hydrated verbatim into `AgentConfig`, and detonates later at runtime (a `TypeError` in the STUCK-state timeout comparison). Conversely, `max_turns` is schema-known and type-checked but deliberately ignored by the hydrator, so the schema tables no longer describe what the hydrator actually honors.

## Narrative

The kernel has two independent views of what an init.json manifest may contain, and they have drifted apart.

The first view is the validation schema. `src/lingtai/init_schema.py` maintains module-level tables — `MANIFEST_REQUIRED` (line 75), `MANIFEST_OPTIONAL` (lines 79–111), `MANIFEST_LEGACY_IGNORED` (lines 120–123), and their union `MANIFEST_KNOWN` (lines 125–127). `validate_init` type-checks every present optional field via `_optional_keys(manifest, MANIFEST_OPTIONAL, ...)` (line 209) and, for anything not in `MANIFEST_KNOWN`, appends a soft warning (lines 280–282):

```python
for key in manifest:
    if key not in MANIFEST_KNOWN:
        warnings.append(f"unknown field: manifest.{key}")
```

The file even opens with a comment acknowledging the two-table maintenance burden: *"When adding a new manifest field, update BOTH MANIFEST_OPTIONAL and MANIFEST_KNOWN — test_init_schema.py enforces this"* (lines 12–13). And `tests/test_init_schema.py` (around lines 358–396) does cross-check the schema tables *against each other*. What nothing checks is the schema tables against the second view.

The second view is the hydrator. `build_agent_config` in `src/lingtai/agent.py` (lines 28–71) overlays manifest values onto `AgentConfig` defaults, and it reads several keys the schema has never heard of:

```python
activeness=manifest.get("activeness", defaults.activeness),          # line 47
...
snapshot_interval=manifest.get(
    "snapshot_interval", defaults.snapshot_interval                  # lines 59-61
),
...
aed_timeout=manifest.get("aed_timeout", defaults.aed_timeout),       # line 66
max_aed_attempts=manifest.get(
    "max_aed_attempts", defaults.max_aed_attempts                    # lines 67-69
),
```

None of `activeness`, `snapshot_interval`, `aed_timeout`, or `max_aed_attempts` appear in `MANIFEST_OPTIONAL` or `MANIFEST_KNOWN`. So an operator who sets `"aed_timeout": 600` in init.json gets a confusing "unknown field: manifest.aed_timeout" warning — even though the value *is* honored. Worse, an operator who typos `"aed_timeout": "600"` (a string) gets the same soft warning, validation succeeds, and the string is written into `AgentConfig.aed_timeout`, which is declared `float` (`src/lingtai_kernel/config.py:79`) but not enforced (dataclasses do not check types at runtime). The bad value lies dormant until the agent actually enters the STUCK state, at which point `src/lingtai_kernel/base_agent/lifecycle.py:560` executes:

```python
if now - agent._aed_start > agent._config.aed_timeout:
```

and raises `TypeError: '>' not supported between instances of 'float' and 'str'` — deep inside the lifecycle loop, far from the init.json that caused it, and only in the (already unhappy) stuck-recovery path. `snapshot_interval` has the same failure shape at `lifecycle.py:600-602`. A malformed `max_aed_attempts` fails slightly earlier but just as opaquely: `AgentConfig.__post_init__` runs `if self.max_aed_attempts < 1` (`src/lingtai_kernel/config.py:125-130`), so a string blows up with a `TypeError` during config hydration rather than a clear "manifest.max_aed_attempts: expected int" at validation time.

How did the design get here? The schema tables were curated during the legacy-field cleanup era — `molt_*`, `stamina`, and `max_turns` all carry deliberate comments about being retired or ignored — while the hydrator kept gaining pass-through reads for operational knobs (`aed_timeout`, `max_aed_attempts`, `snapshot_interval`) that were presumably set programmatically by hosts and never routed through init.json review. The mirror-image drift also exists: `max_turns` sits in `MANIFEST_OPTIONAL` (init_schema.py line 91) and is type-checked as an honored field, but `build_agent_config` deliberately ignores it (agent.py lines 41–45), so the schema advertises a knob that does nothing.

Better looks like: the schema tables are the single source of truth for what init.json may say, every field the hydrator honors is typed there, fields that are recognized-but-ignored live in `MANIFEST_LEGACY_IGNORED` (or an equivalent documented bucket), and a test cross-checks the hydrator's `manifest.get(...)` keys against `MANIFEST_KNOWN` so the two views can never silently drift again.

## Current behavior

- `build_agent_config` (`src/lingtai/agent.py:28-71`) reads `manifest["soul"]`, `manifest["llm"]`, and top-level manifest keys `language`, `activeness`, `context_limit`, `cache_miss_budget`, `snapshot_interval`, `time_awareness`, `timezone_awareness`, `aed_timeout`, `max_aed_attempts`.
- `MANIFEST_OPTIONAL` (`src/lingtai/init_schema.py:79-111`) types `language`, `context_limit`, `cache_miss_budget`, `time_awareness`, `timezone_awareness` (plus others) — but **not** `activeness`, `snapshot_interval`, `aed_timeout`, `max_aed_attempts`.
- Result: those four fields validate with only an "unknown field: manifest.X" warning (`init_schema.py:280-282`) regardless of type, then hydrate verbatim into `AgentConfig` (`src/lingtai_kernel/config.py`) both at boot (`Agent._setup_from_init`, which calls `validate_init` at `agent.py:1067`) and on config reload (`agent.py:1306`).
- Runtime consumers assume correct types with no guards: `lifecycle.py:560` (`aed_timeout` float comparison in STUCK handling), `lifecycle.py:600-602` (`snapshot_interval` comparison), `config.py:125-130` (`max_aed_attempts < 1` clamp in `__post_init__`).
- Inverse drift: `max_turns` is typed in `MANIFEST_OPTIONAL` (`init_schema.py:91`) but deliberately not honored by the hydrator (`agent.py:41-45`).
- `tests/test_init_schema.py` cross-checks the schema tables against each other (`test_*` around lines 358–396) but nothing cross-checks them against `build_agent_config`.

1. **Add the honored fields to `MANIFEST_OPTIONAL`** in `src/lingtai/init_schema.py`, with types matching the `AgentConfig` declarations in `src/lingtai_kernel/config.py`:

   ```python
   MANIFEST_OPTIONAL: dict[str, type | tuple[type, ...]] = {
       ...
       # Honored by build_agent_config (src/lingtai/agent.py) — keep in sync.
       "activeness": (str, type(None)),          # AgentConfig.activeness: str | None
       "snapshot_interval": (int, float, type(None)),  # seconds; None = off
       "aed_timeout": (int, float),              # AgentConfig.aed_timeout: float
       "max_aed_attempts": int,                  # clamped to >= 1 in __post_init__
       ...
   }
   ```

   `MANIFEST_KNOWN` picks these up automatically via the union at lines 125–127. `_check_type` already rejects bools for `int` and `(int, float)` schemas (`init_schema.py:426-427`), so `"aed_timeout": true` is rejected too.

2. **Add range checks in `validate_init`** next to the existing `cache_miss_budget` block (`init_schema.py:295-306`), mirroring its bool-rejection pattern:
   - `aed_timeout`: require `> 0`.
   - `snapshot_interval`: require `> 0` when not `None`.
   - `max_aed_attempts`: require `>= 1` (matches the `__post_init__` clamp in `config.py:125-130`, but fails loudly at validation instead of silently clamping or crashing on a str).

3. **Resolve the `max_turns` inverse drift.** Two options; pick one and document it:
   - Preferred: move `"max_turns"` from `MANIFEST_OPTIONAL` to `MANIFEST_LEGACY_IGNORED` (`init_schema.py:120-123`), since `build_agent_config` deliberately ignores it (`agent.py:41-45`) and tool-loop safety is kernel-owned in `lingtai_kernel.safety_limits`. It stays warning-free on old init.json but stops being advertised as a live typed knob. Note this drops the int type-check on stale values — acceptable because nothing reads it.
   - Alternative: keep it typed in `MANIFEST_OPTIONAL` with a comment that it is validated-but-ignored. Weaker, because the drift-guard test in step 4 then needs a permanent exception.

4. **Add a drift-guard test** in `tests/test_init_schema.py` (or `tests/test_agent_config_hydration.py`) that cross-checks the hydrator against the schema. Simplest robust approach: parse `build_agent_config`'s source for manifest key reads:

   ```python
   def test_hydrator_manifest_keys_are_schema_known():
       import ast, inspect
       from lingtai.agent import build_agent_config
       from lingtai.init_schema import MANIFEST_KNOWN
       tree = ast.parse(inspect.getsource(build_agent_config))
       keys = {
           node.args[0].value
           for node in ast.walk(tree)
           if isinstance(node, ast.Call)
           and isinstance(node.func, ast.Attribute) and node.func.attr == "get"
           and isinstance(node.func.value, ast.Name) and node.func.value.id == "manifest"
           and node.args and isinstance(node.args[0], ast.Constant)
       }
       assert keys <= MANIFEST_KNOWN, f"hydrator reads unknown manifest keys: {sorted(keys - MANIFEST_KNOWN)}"
   ```

   Optionally assert the reverse direction too: every honored key is in `MANIFEST_OPTIONAL` specifically (not merely legacy-ignored), so a future field cannot be honored while untyped.

5. **Back-compat / migration.** No agent-domain migration needed: existing init.json files that already set these fields with correct types go from "warns + honored" to "silent + honored" (strictly better). Files that set them with *wrong* types go from "validates then crashes at runtime" to "fails validation with a precise message" — that is the intended behavior change and should be called out in the changelog. Update the hydration note in `src/lingtai/ANATOMY.md` (the `_setup_from_init` / AgentConfig-hydration bullet) to list the newly schema-known fields.

## Risks and alternatives

- **Hard-fail on previously soft-warned files.** An init.json in the wild with e.g. `"aed_timeout": "360"` currently boots (and later crashes); after this change it fails validation at boot. This is the point of the fix, but it converts a latent bug into an immediate error — flag it in release notes. The blast radius is small: these fields were never documented as schema-supported, so setters are rare.
- **Alternative: remove the hydrator reads instead** (make `aed_timeout` etc. kernel-fixed like the molt thresholds). Rejected: unlike `molt_*`, these are legitimate per-agent operational knobs (a long-running research agent plausibly wants a longer STUCK window or a custom snapshot cadence), and removing them would regress hosts that set them programmatically.
- **Alternative: type-enforce inside `build_agent_config` or `AgentConfig.__post_init__`.** Rejected as the primary fix: validation errors should surface at `validate_init` time with field-path messages, not as hydration exceptions; and `AgentConfig` is kernel-side while init.json policy is wrapper-side. (A cheap defensive `float(...)` coercion in `__post_init__` could be added *in addition*, but it hides operator typos rather than reporting them.)
- **AST-based drift test brittleness.** If `build_agent_config` is refactored away from direct `manifest.get(...)` calls the test could silently match nothing; guard it with `assert keys`, so an empty extraction fails loudly.

## Testing plan

- `tests/test_init_schema.py`:
  - `test_honored_hydrator_fields_are_schema_known` — the AST cross-check from step 4; asserts hydrator-read keys ⊆ `MANIFEST_KNOWN` and asserts the extracted set is non-empty.
  - `test_aed_timeout_rejects_wrong_type` — `validate_init` raises `ValueError` mentioning `manifest.aed_timeout` for `"oops"`, `True`, and `-1`; accepts `360` and `360.5` with no unknown-field warning.
  - `test_snapshot_interval_type_and_range` — accepts `None`/`30`/`2.5`; rejects `"30"`, `True`, `0`, `-5`.
  - `test_max_aed_attempts_type_and_range` — accepts `1`, `5`; rejects `"3"`, `True`, `0`.
  - `test_activeness_type` — accepts `"balanced"`/`None`; rejects `3`.
  - `test_max_turns_is_legacy_ignored` — (if step 3 preferred option is taken) asserts `"max_turns"` in `MANIFEST_LEGACY_IGNORED`, not in `MANIFEST_OPTIONAL`, and that an init.json containing it produces no warning.
- `tests/test_agent_config_hydration.py`:
  - Extend `test_build_agent_config_overlays_explicit_values_and_ignores_stale_molt` (line 117) to set `aed_timeout`, `max_aed_attempts`, `snapshot_interval`, `activeness` in the manifest and assert they land on the returned `AgentConfig` with correct values/types.
- Regression: full `pytest tests/test_init_schema.py tests/test_agent_config_hydration.py` plus the existing schema-consistency tests (lines 358–396) must stay green.

huangzesen:
Fixes #736.

## Problem
`build_agent_config` (src/lingtai/agent.py) honors four manifest fields, but `init_schema` only knew two of them. PR #825 already added `aed_timeout` and `max_aed_attempts`; `activeness` and `snapshot_interval` were still missing from `MANIFEST_KNOWN`, so valid values produced a misleading "unknown field: manifest.X" warning and malformed values (e.g. `"snapshot_interval": "oops"`) passed validation untyped, hydrated verbatim into `AgentConfig`, and detonated later in the lifecycle snapshot loop (`src/lingtai/kernel/base_agent/lifecycle.py`). The inverse drift also remained: `max_turns` was advertised as a typed knob in `MANIFEST_OPTIONAL` while the hydrator deliberately ignores it.

## Fix
- `src/lingtai/init_schema.py`: type `activeness` as `(str, type(None))` and `snapshot_interval` as `(int, float, type(None))` in `MANIFEST_OPTIONAL`; move `max_turns` to `MANIFEST_LEGACY_IGNORED` (recognized-and-ignored, no longer advertised as live); add range checks in `validate_init`: `aed_timeout > 0`, `snapshot_interval > 0` (or null), `max_aed_attempts >= 1`, each rejecting bool explicitly.
- `tests/test_init_schema.py`: new type/range tests for the honored fields, a `max_turns` legacy-status test, and an AST-based drift guard asserting every manifest key `build_agent_config` reads is schema-known AND type-checked (non-empty guard so a refactor cannot silently disable it).
- `src/lingtai/ANATOMY.md`: hydration note now lists the schema-known honored fields.

## Tests
- `pytest tests/test_init_schema.py tests/test_agent_config_hydration.py` — **159 passed** (15 new).
- Regression sweep over every test touching validate_init / hydration / daemon max_turns manifests: `pytest tests/test_init_schema.py tests/test_init_schema_procedures.py tests/test_preset_materialization.py tests/test_deep_refresh.py tests/test_daemon_per_batch_limits.py tests/test_agent.py tests/test_agent_config_hydration.py` — **296 passed**.
- Full suite `pytest tests/ -q` run separately; the only failures observed so far are pre-existing LLM-transport contract failures (e.g. `test_send_str_reaches_transport_as_provider_text[claude_code]`, reproduced on clean main without this change).

## Behavior note
An init.json with a previously-unchecked malformed value (e.g. `"aed_timeout": "600"` or `"snapshot_interval": 0`) now fails validation at boot with a precise field-path message instead of crashing later at runtime — this is the point of the fix; blast radius is small because these fields were never documented as schema-supported.


huangzesen:

PR #1286, opened 2026-08-09
fix: stage-then-rename mail delivery to prevent orphaned inbox dirs (#731)

### LINKED ISSUE #731: FilesystemMailService.send() leaves orphaned partial message directories in the recipient inbox on attachment failure
## Summary

`FilesystemMailService.send()` (`src/lingtai_kernel/services/mail.py`) creates the message directory inside the *recipient's* inbox and starts copying attachments into it **before** validating that all attachment paths exist. When any attachment is missing, `send()` returns an error string with no cleanup, leaving a permanent orphan directory (an `attachments/` subdir with zero or more partially copied files, and no `message.json`) in another agent's mailbox. The recipient's poll loop only dispatches directories that contain `message.json`, so the orphan is never delivered and never removed — garbage silently accumulates in the recipient's inbox on every failed send.

## Narrative

The filesystem mail transport delivers a message by writing a directory `mailbox/inbox/<msg_id>/` into the recipient agent's working tree, containing `message.json` plus an optional `attachments/` folder. The design is deliberately simple and fire-and-forget: the sender does all the writing, and the recipient's `listen()` poll loop picks up any new inbox subdirectory that contains a `message.json`.

The care taken for the *final* write is visible in the code — `message.json` is written via tmp-file-then-`os.replace` so the poller can never observe a half-written JSON file (`src/lingtai_kernel/services/mail.py:199-206`). But the attachment handling that runs *before* that atomic write is not transactional at all:

```python
# src/lingtai_kernel/services/mail.py:181-195
attachment_paths = message.get("attachments")
if attachment_paths:
    att_dir = msg_dir / "attachments"
    att_dir.mkdir(parents=True, exist_ok=True)
    local_copies: list[str] = []
    for fpath in attachment_paths:
        src = Path(fpath)
        if not src.is_file():
            return f"Attachment not found: {fpath}"
        dst = att_dir / src.name
        shutil.copy2(src, dst)
        local_copies.append(str(dst))
```

Note the ordering: `att_dir.mkdir(parents=True, ...)` creates `mailbox/inbox/<msg_id>/attachments/` in the recipient's tree *first*, and only then does the loop check each source path. Suppose an agent sends a message with two attachments and the second path is stale (the file was moved, a temp file was cleaned up, or the LLM simply hallucinated a path — all common in agent workflows). The first file is copied into the recipient's inbox, then the `is_file()` check fails on the second and `send()` returns `"Attachment not found: ..."`. Nothing unwinds the writes already made. The sender gets a clean error string and typically retries with a corrected path — that retry gets a *fresh* `msg_id` from `_new_mailbox_id()`, so the broken directory from the first attempt is never reused or overwritten.

From the recipient's perspective, the orphan is invisible-but-present. The poll loop explicitly requires `message.json` before dispatching:

```python
# src/lingtai_kernel/services/mail.py:241-242
if msg_file.is_file():
```

so the orphan directory is skipped on every tick, forever. It is never delivered, never marked read, never archived, never deleted — no code path in the kernel ever touches it again. Over the lifetime of a long-running agent that receives mail from many peers, these orphans accumulate: each one slows the 0.5 s poll scan slightly (every `iterdir()` pass re-stats the dead entry), confuses anyone inspecting the mailbox on disk, and — most objectionably — represents one agent leaving garbage inside another agent's working directory with no owner responsible for cleaning it up.

The same non-transactional structure affects two sibling failure modes:

1. `shutil.copy2(src, dst)` at line 192 is not wrapped in try/except. If the copy fails (permission error, disk full, source deleted between the `is_file()` check and the copy — a classic TOCTOU window), `send()` doesn't even return an error string; it *raises* `OSError` out of the transport, and the partial directory is again left behind.
2. The final write block at lines 202-208 catches `OSError` and returns `"Failed to write message: ..."` — but likewise performs no cleanup of `msg_dir`, the already-copied attachments, or a possibly-partial `message.json.tmp`.

The fix that matches the codebase's own idioms is staging + atomic rename. The same file already does exactly this dance, carefully, in `_poll_pseudo_outbox()` (`src/lingtai_kernel/services/mail.py:261-382`): claim via `os.replace`, roll back with `_remove_tree_retry()` on failure, publish via `os.replace`. `send()` should assemble the entire message (attachments **and** `message.json`) in a hidden staging directory and publish it into the inbox with a single atomic rename. Then the recipient can never observe a partial message, and any failure path only has to delete the sender-owned staging directory.

## Current behavior

- `FilesystemMailService.send()` (`src/lingtai_kernel/services/mail.py:131-210`):
  - Computes `msg_dir = recipient_dir / mailbox_rel / "inbox" / msg_id` (lines 166-168).
  - If the message has attachments, creates `msg_dir/attachments/` and copies files one at a time (lines 181-195). A missing attachment triggers `return f"Attachment not found: {fpath}"` at line 190 **after** the directory was created and earlier attachments were copied. No cleanup runs; `message.json` is never written.
  - A `shutil.copy2` failure at line 192 raises `OSError` uncaught, with the same orphan left behind.
  - A failure writing `message.json` (lines 202-208) returns an error string, also without removing `msg_dir` or the copied attachments.
- The recipient's poll loop (`listen()`, lines 236-248) dispatches only inbox subdirectories containing `message.json`, so an orphaned partial directory is skipped on every poll tick and persists indefinitely. No other kernel code path garbage-collects inbox entries that lack `message.json`.
- The existing test `test_attachment_file_not_found` (`tests/test_services_mail.py:216-232`) asserts only that the error string is returned; it does not check the recipient's inbox state, so the leak is untested and unnoticed.

All changes are confined to `src/lingtai_kernel/services/mail.py` plus tests. No schema or on-disk format change — the published inbox entry looks byte-for-byte identical to today's successful sends, so there is no migration or back-compat concern for readers.

1. **Validate all attachment paths up front.** In `send()`, before touching the recipient's filesystem, resolve and check every path:

   ```python
   attachment_paths = message.get("attachments") or []
   srcs: list[Path] = []
   for fpath in attachment_paths:
       src = Path(fpath)
       if not src.is_file():
           return f"Attachment not found: {fpath}"
       srcs.append(src)
   ```

   This turns the most common failure (bad path) into a zero-side-effect early return.

2. **Stage the whole message in a hidden temp directory inside the recipient's inbox.** Using the inbox as the staging parent guarantees the final `os.replace` is a same-filesystem rename (atomic), and a dot-prefixed name keeps it self-describing:

   ```python
   inbox_dir.mkdir(parents=True, exist_ok=True)
   staging_dir = inbox_dir / f".{msg_id}.staging"
   ```

   (Or `tempfile.mkdtemp(dir=inbox_dir, prefix=f".{msg_id}.")` for collision safety; either is fine since `msg_id` is a fresh UUID.) The poll loop is unaffected: staging dirs contain no `message.json` until publish, and after publish the name no longer matches — but see step 5 for belt-and-braces.

3. **Copy attachments and write `message.json` into staging, wrapping all I/O.** Copy each `src` into `staging_dir / "attachments"`. Crucially, the recipient-local paths recorded in the payload must point at the **final** location, not the staging path:

   ```python
   local_copies = [str(msg_dir / "attachments" / src.name) for src in srcs]
   message = {**message, "attachments": local_copies}
   ```

   Then write `message.json` directly inside staging (the current tmp-file + `os.replace` dance at lines 199-206 becomes unnecessary because nothing under staging is visible yet, but keeping it is harmless). Wrap the whole staging phase in `try/except OSError` so a mid-copy failure returns an error string instead of raising — this fixes the uncaught `shutil.copy2` exception as a side benefit.

4. **Publish atomically; clean up on any failure.**

   ```python
   try:
       # ... copy attachments, write message.json into staging_dir ...
       os.replace(str(staging_dir), str(msg_dir))
   except OSError as e:
       _remove_tree_retry(staging_dir)   # helper already exists in this module
       return f"Failed to write message: {e}"
   ```

   Reuse the existing `_remove_tree_retry()` helper (`src/lingtai_kernel/services/mail.py:509-514`) rather than adding a new one. After this change there is no code path in which the recipient's visible inbox contains a directory without `message.json` that `send()` created.

5. **(Optional hardening, small)** Make the Phase-1 poll scan in `listen()` skip entries whose name starts with `.` (mirroring the `.{uuid}.claim-` convention already used for the sent/ directory in `_poll_pseudo_outbox()`), so even a crashed sender's leftover staging dir is cheap to skip and visually identifiable for manual or future automated sweeping. This is a one-line `if entry.name.startswith("."): continue`.

6. **Update docs.** Adjust the `send()` docstring (lines 138-150) and the corresponding line in `src/lingtai_kernel/services/ANATOMY.md` (which currently describes the copy-then-write sequence) to describe the stage-then-rename delivery.

## Risks and alternatives

- **`os.replace` on directories:** atomic and supported on POSIX for the rename we need; the destination `msg_dir` never exists beforehand because `msg_id` is a freshly generated UUID, so the rename cannot hit the "destination non-empty directory" edge case. On Windows, `os.replace` onto a *nonexistent* target path also works for directories.
- **Same-filesystem requirement:** renames are only atomic within one filesystem; staging under `recipient_dir/mailbox/inbox/` (not the system temp dir) guarantees this. This is the main correctness pitfall for implementers and worth a code comment.
- **Crash mid-send:** a sender process dying between staging and publish leaves a `.{msg_id}.staging` dir. This is strictly better than today (hidden, identifiable, skippable) but not zero. A stale-staging sweep could be added later; it is out of scope here.
- **Alternative A — cleanup-on-error only** (keep in-place copying, add `shutil.rmtree(msg_dir)` before each error return): simpler diff, but the recipient still observes a mutating partial directory during the copy window, a crash mid-copy still orphans, and every future edit to `send()` must remember to unwind. Rejected in favor of making partial states unobservable by construction.
- **Alternative B — validate-then-copy without staging:** fixes the common missing-file case but leaves the TOCTOU race (file deleted between check and copy) and copy-failure orphans. Rejected as incomplete.
- **Alternative C — teach the poller to garbage-collect message-less dirs:** puts cleanup responsibility on the victim, risks deleting a concurrent sender's in-progress delivery (today's `send()` legitimately has message-less dirs in flight), and doesn't fix the visibility race. Rejected; note that *after* this fix plus step 5's dot-prefix convention, such a sweeper would become safe to add.

## Testing plan

Add to `tests/test_services_mail.py` (fixtures `_setup_agent_dir` / `_keep_heartbeat_alive` already exist there):

1. `test_attachment_not_found_leaves_no_inbox_orphan` — send a message whose only attachment path does not exist; assert the error string is returned **and** `receiver_dir/mailbox/inbox` contains no entries (or only pre-existing ones).
2. `test_partial_attachment_failure_leaves_no_inbox_orphan` — two attachments where the first exists and the second does not; assert the error is returned, the inbox has no new entries, and no copy of the first attachment exists anywhere under the recipient's mailbox.
3. `test_attachment_copy_error_returns_error_not_raise` — monkeypatch `shutil.copy2` to raise `OSError`; assert `send()` returns an error string (does not raise) and the inbox is clean.
4. `test_delivered_attachment_paths_point_at_final_location` — successful send with one attachment; assert the delivered `message.json`'s `attachments` entries are paths under `mailbox/inbox/<msg_id>/attachments/` and that each file exists at exactly that path (guards against accidentally recording staging paths after the refactor).
5. `test_listener_never_dispatches_partial_message` — with a listener running, perform a failing send (missing attachment) followed by a successful send; assert exactly one message is dispatched and the failing send left nothing behind.
6. Extend the existing `test_attachment_file_not_found` with an inbox-emptiness assertion so the regression is pinned in the test that already covers the error string.

huangzesen:
Fixes #731.

## Summary

`PosixFilesystemMailAdapter.send()` (`src/lingtai/adapters/posix/mail.py`) created the message directory and copied attachments directly into the recipient's inbox **before** validating that every attachment path existed. Two failure modes left permanent orphans in another agent's mailbox:

- A missing attachment (`is_file()` false — a stale/moved file, a cleaned-up temp file, or a hallucinated path) returned `"Attachment not found: ..."` *after* `att_dir.mkdir(...)` had created `inbox/<id>/attachments/` and any earlier attachments were copied. No cleanup ran.
- `shutil.copy2` was uncaught, so a copy error (permission, disk full, TOCTOU delete) **raised** `OSError` out of the transport, again leaving the partial directory behind.

Also, a failure writing `message.json` returned an error string without removing `msg_dir` or the copied attachments. The recipient's poll loop only dispatches inbox dirs containing `message.json`, so these orphans are never delivered, never cleaned up, and accumulate on every failed send.

**Root cause:** the pre-`message.json` attachment phase was non-transactional while the final JSON write was already atomic. The fix stays at the owning layer (`send()`) and reuses the module's own claim/publish idiom from `_poll_pseudo_outbox()`:

1. Validate all attachment paths up front → the common bad-path case is a zero-side-effect early return.
2. Stage the whole message (attachments + `message.json`) in a hidden `.<id>.staging` sibling of the inbox (same filesystem → `os.replace` is atomic; dot prefix keeps it out of the listener scans).
3. Wrap all staging I/O so a `copy2` failure returns an error string instead of raising.
4. Publish with a single atomic `os.replace`; on any failure clean up the sender-owned staging dir via the existing `_remove_tree_retry`.
5. Belt-and-braces: the listener's Phase-1 snapshot and Phase-2 own-inbox scan skip dot-prefixed entries, so even a crashed sender's leftover staging dir is never dispatched and is identifiable for future sweeping.

Recipient-local attachment paths recorded in the payload point at the **final** location (`inbox/<id>/attachments/<name>`), not the staging path. Published entries are byte-identical to today's successful sends — no schema/format change, no migration, no new config flag. Docs updated: `send()` docstring, `src/lingtai/adapters/posix/ANATOMY.md`, and `src/lingtai/kernel/mail_transport/CONTRACT.md` (failed sends leave no partial entry in the recipient's inbox).

## Validation

Failing-first (against unmodified `send()`), the bug-exposing tests fail:

```
$ PYTHONPATH=$PWD/src python -m pytest tests/test_services_mail.py -q -k "orphan or copy_error or final_location or partial_message or attachment_file_not_found"
FAILED tests/test_services_mail.py::TestMailAttachments::test_attachment_file_not_found
FAILED tests/test_services_mail.py::TestMailAttachments::test_attachment_not_found_leaves_no_inbox_orphan
FAILED tests/test_services_mail.py::TestMailAttachments::test_partial_attachment_failure_leaves_no_inbox_orphan
FAILED tests/test_services_mail.py::TestMailAttachments::test_attachment_copy_error_returns_error_not_raise
FAILED tests/test_services_mail.py::TestMailAttachments::test_listener_never_dispatches_partial_message
5 failed, 1 passed, 10 deselected
```
(`test_attachment_copy_error_...` failed by *raising* the uncaught `OSError`; the rest failed on orphaned inbox dirs / stray copies.)

After the fix, the full module passes:

```
$ python -m pytest tests/test_services_mail.py -q
16 passed in 6.86s
```

Targeted existing mail/email suites for the touched transport (identical pass/fail to baseline on unmodified main — no regressions):

```
$ python -m pytest tests/test_filesystem_mail.py tests/test_mail_transport.py \
    tests/test_email_abs_reply_route.py tests/test_three_agent_email.py \
    tests/test_layers_email.py tests/test_email_identity.py \
    tests/test_time_awareness_mail.py tests/test_time_awareness_email.py -q
141 passed in 51.23s
```

```
$ git diff --check   # (no output — clean)
```

### DIFF
--- a/src/lingtai/adapters/posix/ANATOMY.md
+++ b/src/lingtai/adapters/posix/ANATOMY.md
@@ -88,11 +88,14 @@ co-located owning ANATOMY.md files.
 - `PosixFilesystemMailAdapter` implements `MailTransportPort` by delivering
   messages as files into a recipient's inbox and polling its own inbox plus
   subscribed pseudo-agent outboxes (`src/lingtai/adapters/posix/mail.py:34-69`).
-- `send()` handshakes, injects mailbox metadata, copies attachments, and writes
-  `message.json` atomically via `os.replace`
-  (`send()`, `src/lingtai/adapters/posix/mail.py:102`; atomic write at :184);
+- `send()` handshakes, injects mailbox metadata, validates every attachment
+  path up front, stages the full message (attachments + `message.json`) in a
+  hidden `.<id>.staging` dir under the inbox, and publishes it with a single
+  atomic `os.replace` — the recipient never observes a partial entry and any
+  failure removes the sender-owned staging dir
+  (`send()`, `src/lingtai/adapters/posix/mail.py:102`; publish at :199);
   `listen()`/`stop()` own the 0.5-second daemon poll loop with pseudo-outbox
-  priority and per-phase `OSError` isolation.
+  priority, dot-prefixed staging skips, and per-phase `OSError` isolation.
 - `PosixWorkdirLeaseAdapter` implements `WorkdirLeasePort` by holding an exclusive
   non-blocking `fcntl.flock` on `<workdir>/.agent.lock`
   (`src/lingtai/adapters/posix/workdir_lease.py:27-95`); `acquire()` polls at
--- a/src/lingtai/adapters/posix/mail.py
+++ b/src/lingtai/adapters/posix/mail.py
@@ -112,8 +112,12 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         1. ``{address}/.agent.json`` must exist.
         2. ``{address}/.agent.heartbeat`` must be fresh (< 2 s).
 
-        Then write ``message.json`` atomically into the recipient's inbox
-        and copy any attachment files.
+        Then stage the whole message (attachments + ``message.json``) in a
+        hidden ``.<id>.staging`` directory inside the recipient's inbox and
+        publish it with a single atomic ``os.replace``, so the recipient can
+        never observe a partial inbox entry. Every failure path removes the
+        sender-owned staging directory and leaves the recipient's inbox
+        untouched.
 
         Modes:
         - peer: resolve bare name against parent dir (default — sibling agents in same .lingtai/)
@@ -155,34 +159,47 @@ class PosixFilesystemMailAdapter(MailTransportPort):
             ),
         }
 
-        # Handle attachments
+        # Validate every attachment path up front: a missing file is a
+        # zero-side-effect early return instead of a partial inbox entry.
         attachment_paths = message.get("attachments")
+        srcs: list[Path] = []
         if attachment_paths:
-            att_dir = msg_dir / "attachments"
-            att_dir.mkdir(parents=True, exist_ok=True)
-            local_copies: list[str] = []
             for fpath in attachment_paths:
                 src = Path(fpath)
                 if not src.is_file():
                     return f"Attachment not found: {fpath}"
-                dst = att_dir / src.name
-                shutil.copy2(src, dst)
-                local_copies.append(str(dst))
-            # Replace original paths with recipient-local paths
-            message = {**message, "attachments": local_copies}
-        else:
-            msg_dir.mkdir(parents=True, exist_ok=True)
+                srcs.append(src)
 
-        # Atomic write: tmp → rename
-        tmp_path = msg_dir / "message.json.tmp"
-        final_path = msg_dir / "message.json"
+        # Assemble the whole message (attachments + ``message.json``) in a
+        # hidden staging dir inside the recipient's inbox, then publish with a
+        # single atomic ``os.replace``. Staging under the inbox (not the system
+        # temp dir) keeps the final rename on one filesystem, and the
+        # dot-prefixed name keeps the listener from ever dispatching a partial
+        # entry. Every failure path removes the sender-owned staging dir, so a
+        # failed send never leaves an orphan in another agent's mailbox.
+        inbox_dir.mkdir(parents=True, exist_ok=True)
+        staging_dir = inbox_dir / f".{msg_id}.staging"
         try:
-            tmp_path.write_text(
+            staging_dir.mkdir(parents=True, exist_ok=True)
+            if srcs:
+                staging_att_dir = staging_dir / "attachments"
+                staging_att_dir.mkdir(parents=True, exist_ok=True)
+                # Recipient-local paths recorded in the payload must point at
+                # the final location (``msg_dir``), not the staging path.
+                local_copies = [
+                    str(msg_dir / "attachments" / src.name) for src in srcs
+                ]
+                for src in srcs:
+                    shutil.copy2(src, staging_att_dir / src.name)
+                message = {**message, "attachments": local_copies}
+
+            (staging_dir / "message.json").write_text(
                 json.dumps(message, indent=2, ensure_ascii=False, default=str),
                 encoding="utf-8",
             )
-            os.replace(str(tmp_path), str(final_path))
+            os.replace(str(staging_dir), str(msg_dir))
         except OSError as e:
+            _remove_tree_retry(staging_dir)
             return f"Failed to write message: {e}"
 
         return None
@@ -201,6 +218,10 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         # Snapshot existing inbox entries so we don't re-notify
         if self._inbox_dir.is_dir():
             for entry in self._inbox_dir.iterdir():
+                if entry.name.startswith("."):
+                    # Sender-owned staging dirs (``.{id}.staging``) are
+                    # invisible to delivery; skip them here and in the scan.
+                    continue
                 if entry.is_dir():
                     self._seen.add(entry.name)
 
@@ -293,6 +314,12 @@ class PosixFilesystemMailAdapter(MailTransportPort):
                     return visited > 0
 
                 visited += 1
+                # Sender-owned staging dirs (``.{id}.staging``) contain a
+                # ``message.json`` while the message is being assembled; only
+                # the final atomic publish is deliverable, so never dispatch or
+                # record dot-prefixed entries.
+                if entry.name.startswith("."):
+                    continue
                 # _seen only holds handled directory names or pseudo-claim
                 # UUIDs, so skip before the stat.
                 if entry.name in self._seen:
--- a/src/lingtai/kernel/mail_transport/CONTRACT.md
+++ b/src/lingtai/kernel/mail_transport/CONTRACT.md
@@ -40,7 +40,8 @@ polling, or concurrency mechanisms.
 Agents and coding agents MUST preserve the current observable semantics:
 fire-and-forget `send` (no request/response coupling), background non-blocking
 `listen`, explicit idempotent `stop`, handshake-before-delivery, mailbox metadata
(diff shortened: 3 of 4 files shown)


huangzesen:

PR #1287, opened 2026-08-09
fix(kernel): serialize _sync_notifications against run-loop/heartbeat double-inject (#659)

### LINKED ISSUE #659: _sync_notifications has TOCTOU race: missing _system_notification_lock (run loop vs heartbeat)
## Summary

`_sync_notifications()` in `base_agent/__init__.py:1092` performs a check-then-act (TOCTOU) on `self._notification_fp` without holding `self._system_notification_lock`. Both the run-loop IDLE boundary (`turn.py:830`) and the heartbeat thread (`lifecycle.py:503`) call this method concurrently, creating a race that can double-inject notification pairs.

### No lock in `_sync_notifications`

The lock `self._system_notification_lock` is defined at `__init__.py:500` and used by:
- `notifications.py:368` (collect/dismiss operations)
- `notifications.py:988`
- `base_agent/messaging.py:186`
- `nudge/goal.py:126`

But `_sync_notifications()` at `__init__.py:1092` does **not** acquire this lock. The method performs:

```python
fp = notification_fingerprint(self._working_dir)      # read
if fp == self._notification_fp:                         # compare
    return
# ... collect, inject, skeletonize ...
self._notification_fp = fp                              # write
```

### Two concurrent callers

1. **Run loop** (`turn.py:830` via `_turn_boundary_housekeeping`): calls `_sync_notifications()` on every turn boundary.
2. **Heartbeat thread** (`lifecycle.py:503`): calls `_sync_notifications()` on every heartbeat tick.

Both callers can enter `_sync_notifications` simultaneously, both see the same stale fingerprint, and both proceed to inject notification pairs — resulting in duplicate notification blocks in conversation history.

### Impact

**MEDIUM** — Double-injected notification pairs are not data-loss-level (they are eventually deduplicated on the next successful sync), but they:
- Pollute conversation history with duplicate blocks
- Cause unnecessary LLM input tokens (duplicate content)
- Can confuse the agent with seemingly repeated notifications
- Are a race condition, so they are non-deterministic and hard to debug

## Suggested fix

Acquire `_system_notification_lock` at the top of `_sync_notifications`:

```python
def _sync_notifications(self) -> None:
    with self._system_notification_lock:
        fp = notification_fingerprint(self._working_dir)
        if fp == self._notification_fp:
            return
        # ... rest of method ...
```

The lock already exists for this purpose — it was defined at `__init__.py:500` but never wired into the sync method.

## Verification

Source verified:
- `_system_notification_lock` defined at `base_agent/__init__.py:500`
- Used in `notifications.py:368,988`, `messaging.py:186`, `nudge/goal.py:126`
- NOT used in `_sync_notifications()` at `base_agent/__init__.py:1092`
- Run-loop caller at `base_agent/turn.py:958-966` (via `_turn_boundary_housekeeping`)
- Heartbeat caller at `base_agent/lifecycle.py:503`

Review: em-14 (skeptical concurrency review).

huangzesen:
Fixes #659.

## Problem

`_sync_notifications()` performs a check-then-act on `self._notification_fp`:
both the run-loop IDLE boundary (`turn.py`) and the heartbeat thread
(`lifecycle.py`) call it concurrently, and nothing serializes the in-memory
fingerprint check + wire append. The notification store's flock guards
on-disk mutations only; it does not serialize the sync check-then-act.
Two concurrent callers can both observe the same stale fingerprint and
double-inject a synthesized notification pair: duplicate history blocks,
duplicate `MSG_TC_WAKE` messages, and wasted LLM tokens.

Reproduced on current main (HEAD `df2b523d`) with a two-thread harness:
both the "runloop" and "heartbeat" threads passed the fp check, producing
4 wire entries (double-injected pair) and 2 wake messages for one
fingerprint change.

## Fix

- Add `self._notification_sync_lock` (`threading.RLock`) per agent.
- `_sync_notifications()` now acquires it around the whole check-then-act
  (thin wrapper delegating to `_sync_notifications_locked`). The losing
  caller re-reads the fingerprint after the winner commits and no-ops.
- RLock (not plain Lock) so hook/synthesize paths that transitively
  re-enter sync cannot self-deadlock. Lazy `getattr` init keeps partial
  test doubles that bypass `__init__` working.

## Tests

- New regression test `test_sync_notifications_serialized_across_runloop_and_heartbeat`
  in `tests/test_notification_sync.py`:
  - pre-fix: **FAILS** with `double injection: both callers passed the fp
    check (['heartbeat', 'runloop'])`;
  - post-fix: **passes** (one injection, one pair on the wire, one wake).
- `pytest tests/test_notification_sync.py tests/test_turn_boundary_housekeeping.py tests/test_system_dismiss.py`
  → 109 passed; the only failures (3 in `test_system_dismiss.py`) are
  order-dependent flakes that pass individually and reproduce identically
  on pristine main with this change stashed (verified).
- Notification-family suites (`test_notification_store`, `test_tc_inbox*`,
  `test_system_notifications`, `test_molt_notification_persistence`,
  `test_post_molt_notification`, `test_goal_notification`,
  `test_messaging_notification_format`, `test_notification_tool`,
  `test_large_result_no_notification`) → **173 passed**.
- Run-loop/heartbeat suites (`test_base_agent`, `test_heartbeat`,
  `test_lifecycle_stop_wake`, `test_runtime_block_turn_integration`) → **38 passed**.
- `git diff --check` → clean.

### DIFF
--- a/src/lingtai/kernel/base_agent/__init__.py
+++ b/src/lingtai/kernel/base_agent/__init__.py
@@ -615,6 +615,14 @@ class BaseAgent:
         #   skeletonized in-place, not deleted).
         # See notifications.py and notification-filesystem-redesign.md.
         self._notification_fp: tuple = ()
+        # Serializes ``_sync_notifications()`` check-then-act between the
+        # run-loop IDLE boundary and the heartbeat thread (issue #659).
+        # The store's flock guards on-disk mutations only; it does NOT
+        # serialize this in-memory fingerprint check + wire append, so
+        # without this lock both callers can pass the fp check and
+        # double-inject notification pairs.  RLock so hook/synthesize
+        # paths that transitively re-enter sync cannot self-deadlock.
+        self._notification_sync_lock: threading.RLock = threading.RLock()
         # System-channel RMW serialization is owned by the injected
         # NotificationStorePort through compare_update_channel.
         # Last ACTIVE-state notification fingerprint that has already emitted
@@ -1273,6 +1281,22 @@ class BaseAgent:
     # ------------------------------------------------------------------
 
     def _sync_notifications(self) -> None:
+        """Sync `.notification/` state into the wire.
+
+        Serialized under ``_notification_sync_lock``: the run-loop IDLE
+        boundary (turn.py) and the heartbeat thread (lifecycle.py) call
+        this concurrently, and without the lock their check-then-act on
+        ``_notification_fp`` can both observe the same stale fingerprint
+        and inject duplicate notification pairs (issue #659).
+        """
+        lock = getattr(self, "_notification_sync_lock", None)
+        if lock is None:
+            # Partial test doubles bypass __init__; give them a real lock.
+            lock = self._notification_sync_lock = threading.RLock()
+        with lock:
+            self._sync_notifications_locked()
+
+    def _sync_notifications_locked(self) -> None:
         """Sync `.notification/` state into the wire.
 
         Computes the current fingerprint; if unchanged, no-op.  On change:
--- a/tests/test_notification_sync.py
+++ b/tests/test_notification_sync.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 import json
 import threading
+import time
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


huangzesen:

PR #1288, opened 2026-08-09
fix(kernel): migrate remaining fixed-temp-name atomic writers to _fsutil (resolved manifest, soul init.json, mail)

### LINKED ISSUE #756: Finish the _fsutil migration for remaining fixed-temp-name atomic writers (resolved manifest, soul init.json, mail message writes)
## Summary

`src/lingtai_kernel/_fsutil.py` was introduced (issue #510, plan in `docs/plans/2026-06-25-fsutil-migration.md`) specifically because an audit flagged fixed/pid-only temp names as a same-target collision risk, but three writers still hand-roll the old fixed-temp-name pattern: `write_resolved_manifest` in `src/lingtai_kernel/workdir.py`, `_atomic_write_init` in `src/lingtai_kernel/intrinsics/soul/config.py`, and three `message.json.tmp` sites in `src/lingtai_kernel/services/mail.py`. Migrating them to `_fsutil.atomic_write_text`/`atomic_write_json` removes a real concurrent-writer race on the resolved manifest, fixes a latent locale-dependent encoding bug in mail delivery, and deletes ~40 lines of duplicated atomic-write boilerplate.

## Narrative

The kernel persists a lot of small JSON state files, and for years each module re-solved "atomic write" independently: write a sibling temp file, then `os.replace` it over the target. An audit found that most of these hand-rolled writers use a *fixed* temp name (e.g. `manifest.resolved.json.tmp`) rather than a unique one. That is fine with exactly one writer, but the moment two writers target the same file concurrently — two threads, two tasks, or a boot racing a daemon refresh — they share one temp path. Writer A opens the temp, writer B truncates and rewrites it, writer A's `os.replace` publishes B's half-written bytes (or B's replace wins and A's replace then fails with `FileNotFoundError`). The fix was `_fsutil._unique_tmp` (`src/lingtai_kernel/_fsutil.py:55-73`), which embeds pid + uuid4 in the temp name and opens with `"x"` (exclusive create), making concurrent same-target writes collision-free. The module docstring and `docs/plans/2026-06-25-fsutil-migration.md` lay out a staged migration; `WorkingDir.write_manifest` was migrated as the proof of concept.

The highest-value unmigrated writer is `write_resolved_manifest` (`src/lingtai_kernel/workdir.py:191-233`). Per its own docstring the artifact is "regenerated on every boot/refresh" — it is called from `lingtai.Agent._read_init` on boot, live refresh, and molt-reload. It still writes through a *fixed* sibling temp exposed as a layout property:

```python
# src/lingtai_kernel/workdir.py:118-120
@property
def resolved_manifest_tmp(self) -> Path:
    return self.system_dir / (_RESOLVED_MANIFEST_FILE + ".tmp")
```

```python
# src/lingtai_kernel/workdir.py:224-230
target = layout.resolved_manifest
tmp = layout.resolved_manifest_tmp
tmp.write_text(
    json.dumps(artifact, indent=2, ensure_ascii=False) + "\n",
    encoding="utf-8",
)
os.replace(str(tmp), str(target))
```

A concurrent boot + refresh (or any two reconstruction paths overlapping) can race exactly as the audit described: both write `system/manifest.resolved.json.tmp`, one `os.replace` clobbers the other's in-flight temp. Because the function is deliberately best-effort (`except (OSError, TypeError, ValueError): return None`), the loser's failure is *silent* — the TUI/portal consumers of `manifest.resolved.json` (issue #259) can be left reading a stale or torn artifact with no log trail. The plan doc already lists this as an unchecked Stage 1 item: "`workdir.py::write_resolved_manifest` (`manifest.resolved.json`, trailing newline → pass explicit newline or keep inline)".

The second writer is `_atomic_write_init` (`src/lingtai_kernel/intrinsics/soul/config.py:325-351`), shared by `_persist_soul_config` and `_persist_soul_voice`, both of which rewrite the agent's `init.json` — the *user-owned input file*, the worst possible file to tear. It uses `init_path.with_suffix(init_path.suffix + ".tmp")`, i.e. the fixed name `init.json.tmp`. Two soul-tool invocations landing close together (e.g. a voice change and a delay change from different tasks) share the temp path. It also duplicates fsync/cleanup/error-string logic that `_fsutil.atomic_write_text(fsync=True)` already provides.

Third, `src/lingtai_kernel/services/mail.py` hand-rolls tmp→rename three times with the fixed name `message.json.tmp` (in `send` at lines 200-206, in the pseudo-agent claim path at lines 323/340-343, and in `_write_runtime_probe_reply` at lines 447-455). Here each message directory is keyed by a fresh uuid, so cross-writer collision is unlikely; the concrete defect is different: two of the three sites call `write_text` **without** `encoding=`:

```python
# src/lingtai_kernel/services/mail.py:203-206 (send)
tmp_path.write_text(
    json.dumps(message, indent=2, ensure_ascii=False, default=str)
)
os.replace(str(tmp_path), str(final_path))
```

`Path.write_text` without `encoding` uses the locale's preferred encoding, so on a non-UTF-8 locale a message containing non-ASCII (CJK subject lines are the norm in this project) raises `UnicodeEncodeError` or writes mojibake, while the third site (line 453) correctly passes `encoding="utf-8"` — three copies of the same pattern that have already drifted apart, which is precisely the disease `_fsutil` was built to cure.

Better looks like: all three modules route through `_fsutil.atomic_write_text`/`atomic_write_json`, byte-for-byte format-identical (the plan's migration discipline mandates golden tests), with the fixed-temp-name race and the encoding drift gone, and the plan checklist updated so coverage stays visible.

## Current behavior

- `src/lingtai_kernel/workdir.py:191-233` — `write_resolved_manifest` writes `system/manifest.resolved.json` via the fixed sibling temp `WorkdirLayout.resolved_manifest_tmp` (`workdir.py:118-120`) + `os.replace`. Format: `json.dumps(artifact, indent=2, ensure_ascii=False) + "\n"` (trailing newline), no fsync, best-effort (returns `None` on failure). Called on every boot/refresh/molt-reload from `src/lingtai/agent.py::_read_init` (~line 1083). Two concurrent callers for the same workdir race on one temp path.
- `src/lingtai_kernel/intrinsics/soul/config.py:325-351` — `_atomic_write_init` writes `init.json` via fixed `init.json.tmp`, `indent=2, ensure_ascii=False` + trailing newline, best-effort `os.fsync` (swallows `OSError`), returns `None` on success or a truncated error string. Callers: `_persist_soul_config` (line 277) and `_persist_soul_voice` (line 322).
- `src/lingtai_kernel/services/mail.py` — three hand-rolled `message.json.tmp` → `os.replace` sites: `send` (lines 200-206, no `encoding=`), pseudo-agent claim delivery in `_drain_pseudo_outbox` (lines 323, 340-343, no `encoding=`), `_write_runtime_probe_reply` (lines 446-455, has `encoding="utf-8"`). All write `indent=2, ensure_ascii=False, default=str`, no trailing newline.
- `docs/plans/2026-06-25-fsutil-migration.md` — `write_resolved_manifest` is an unchecked Stage 1 item; `intrinsics/soul/config.py` is an unchecked Stage 3 item; `services/mail.py` is not yet on the list at all (only `intrinsics/email/primitives.py` is).

Per the plan's migration discipline ("one concern per PR"), land as three small PRs — or one PR with three clearly separated commits.

1. **`write_resolved_manifest` → `atomic_write_text`** (`src/lingtai_kernel/workdir.py`)
   - Add `from ._fsutil import atomic_write_text` (or a lazy import inside the function, matching existing style).
   - Replace lines 224-230 with:
     ```python
     target = layout.resolved_manifest
     atomic_write_text(
         target,
         json.dumps(artifact, indent=2, ensure_ascii=False) + "\n",
     )
     return target
     ```
     Keep `json.dumps` at the call site (not `atomic_write_json`) because the format has a trailing newline, which `atomic_write_json` deliberately does not add — this is exactly the case plan rule 3 calls out. Serializing before calling also keeps `TypeError`/`ValueError` inside the existing `except (OSError, TypeError, ValueError)` best-effort envelope; `atomic_write_text` raises only `OSError` subclasses on I/O failure, so the envelope still holds.
   - The explicit `layout.system_dir.mkdir(parents=True, exist_ok=True)` (line 223) becomes redundant (`atomic_write_text` creates parents) but is harmless; drop it.
   - Delete the now-unused `WorkdirLayout.resolved_manifest_tmp` property (lines 118-120) and the assertion on it in `tests/test_workdir.py:53`. It is an internal layout detail; grep confirms no other consumers.
   - Tick the Stage 1 checkbox in `docs/plans/2026-06-25-fsutil-migration.md` and update the `ANATOMY.md` entries that describe the write.

2. **`_atomic_write_init` → `atomic_write_text`** (`src/lingtai_kernel/intrinsics/soul/config.py`)
   - Keep the function and its `str | None` error-string contract (both callers depend on it); replace its body:
     ```python
     def _atomic_write_init(init_path, data) -> str | None:
         import json
         from lingtai_kernel._fsutil import atomic_write_text
         try:
             atomic_write_text(
                 init_path,
                 json.dumps(data, ensure_ascii=False, indent=2) + "\n",
                 fsync=True,
             )
             return None
         except Exception as e:
             return f"failed to write init.json: {e!s}"[:200]
     ```
   - Behavior deltas to call out in the PR: (a) unique temp name instead of fixed `init.json.tmp` (the point of the change); (b) an `os.fsync` failure now surfaces as an error string instead of being silently swallowed — strictly safer for the user-owned `init.json`, and the callers already log the returned string. Temp-file cleanup on failure is handled inside `atomic_write_text`.

3. **`services/mail.py` message writers → `atomic_write_json`**
   - Replace all three tmp→rename blocks with `atomic_write_json(final_path, message, indent=2, ensure_ascii=False, default=str)` (defaults already match; `default=str` must be passed). `atomic_write_json` adds no trailing newline, matching all three sites today.
   - In `send`, keep the surrounding `try/except OSError` returning `f"Failed to write message: {e}"`; in the claim path and probe reply, keep the existing `except OSError` logging/rollback structure — only the write call changes, not the claim/rollback protocol around it (the `os.replace` calls that move *directories* at lines 328, 361, 521 are rename-based claims, not file writes, and stay as-is).
   - This also fixes the missing `encoding="utf-8"` at the two `send`/claim sites — flag it in the PR description as an intentional bug fix (byte-identical on UTF-8 locales, corrected behavior elsewhere).
   - Add `services/mail.py` to the Stage 1 list in `docs/plans/2026-06-25-fsutil-migration.md` and tick it.

No schema, data-structure, or on-disk format changes: unique temp names are transient (and `_unique_tmp`'s dot-prefixed names are *less* visible to directory scanners than the current bare `*.tmp` names — the mail poll loops at `services/mail.py:236-247` and `296-302` filter on directories / `message.json` presence, so they never observed the temps anyway). No migration or back-compat handling needed; stale fixed-name `.tmp` litter from old crashes is simply never referenced again.

## Risks and alternatives

- **Silent format drift** is the main risk the plan warns about. `atomic_write_json` omits the trailing newline that `write_resolved_manifest` and `_atomic_write_init` emit; using `atomic_write_text(json.dumps(...) + "\n")` for those two, guarded by byte-identity golden tests, eliminates it.
- **fsync semantics change in soul config** (swallowed → surfaced): could turn a previously "successful" write on an fsync-hostile filesystem into a reported error. Alternative: pass `fsync=False` and drop durability, or wrap fsync separately for exact parity. Surfacing the error wins: `init.json` is user-owned input, the callers already handle/log error strings, and silent non-durability is the worse failure.
- **Temp-name visibility**: anything that hard-codes the old temp names would break. Grep shows only `WorkdirLayout.resolved_manifest_tmp` and its single test assertion; both are removed in step 1.
- **Alternative: per-target file locking** instead of unique temp names — heavier, adds lock-file litter and deadlock surface, and `_fsutil` already implements the last-writer-wins semantics these regenerated artifacts want.
- **Alternative: leave mail.py for the later `intrinsics/email` stage** — defensible, but the encoding bug at `services/mail.py:203` and `:340` is worth fixing now, and the change is mechanical.

## Testing plan

Following the plan's golden-test discipline, add before swapping each implementation:

- `tests/test_resolved_manifest_artifact.py::test_write_resolved_manifest_byte_identical_to_legacy` — freeze `generated_at` (monkeypatch `datetime` or compare with the timestamp field normalized), write a known manifest, assert the produced bytes equal a captured pre-migration golden (including the trailing newline and `ensure_ascii=False` characters).
- `tests/test_resolved_manifest_artifact.py::test_write_resolved_manifest_concurrent_writers` — N threads calling `write_resolved_manifest` on the same workdir in a loop; assert every call returns the target path (no silent `None`), the final file parses as valid JSON with the expected schema, and no `*.tmp` files remain in `system/`.
- `tests/test_workdir.py` — drop the `resolved_manifest_tmp` assertion (line 53); keep the `resolved_manifest` one.
- `tests/test_soul.py::test_atomic_write_init_format_parity` — round-trip `_atomic_write_init` on a sample dict; assert bytes match the legacy format (indent=2, `ensure_ascii=False`, trailing newline) and no temp file is left beside `init.json`.
- `tests/test_soul.py::test_atomic_write_init_error_string_on_failure` — point it at an unwritable path (read-only dir); assert a `"failed to write init.json:"` string is returned, no exception escapes, and no temp litter remains.
- `tests/test_services_mail.py::test_send_writes_utf8_message_json` — send a message with non-ASCII subject/body under a forced non-UTF-8 locale encoding (monkeypatch `locale.getpreferredencoding` or set `PYTHONIOENCODING`-independent check by reading raw bytes); assert `message.json` is valid UTF-8 with unescaped non-ASCII and no `message.json.tmp` sibling exists.
- `tests/test_services_mail.py::test_probe_reply_format_parity` — golden-byte comparison for `_write_runtime_probe_reply` output with fixed uuid/timestamp injection.
- Existing suites (`tests/test_fsutil.py`, `tests/test_filesystem_mail.py`, `tests/test_three_agent_email.py`, `tests/test_soul.py`) must pass unchanged — they exercise the delivery/claim protocol end-to-end and act as the regression net.

huangzesen:
Fixes #756.

## Summary

Migrates the last three fixed-temp-name atomic writers to the shared `_fsutil` helpers (issue #510 plan, `docs/plans/2026-06-25-fsutil-migration.md`), removing the same-target concurrent-writer race on a shared temp path and deleting ~40 lines of duplicated atomic-write boilerplate:

1. **`workdir.py::write_resolved_manifest`** (`system/manifest.resolved.json`) — replaced the fixed sibling temp `WorkdirLayout.resolved_manifest_tmp` + `os.replace` with `_fsutil.atomic_write_text(target, json.dumps(..., indent=2, ensure_ascii=False) + "\n")`. The trailing-newline format is preserved exactly (plan rule 3); serialization stays at the call site so `TypeError`/`ValueError` remain inside the existing best-effort `except (OSError, TypeError, ValueError)` envelope. Removed the now-unused `resolved_manifest_tmp` layout property and its test assertion. On-disk format: byte-identical.
2. **`tools/soul/config.py::_atomic_write_init`** (user-owned `init.json`) — replaced the fixed `init.json.tmp` + `os.replace` + hand-rolled fsync with `_fsutil.atomic_write_text(init_path, json.dumps(data, ensure_ascii=False, indent=2) + "\n", fsync=True)`, keeping the `str | None` error-string contract both callers depend on. Intentional behavior delta (flagged in #756): an `os.fsync` failure now surfaces as the error string instead of being silently swallowed — strictly safer for user-owned input; callers already log the returned string. Format: byte-identical.
3. **`adapters/posix/mail.py`** — all three `message.json.tmp` → `os.replace` sites (`send`, pseudo-agent claim delivery, runtime-probe ack) now use `_fsutil.atomic_write_json(final, msg, indent=2, ensure_ascii=False, default=str)`, which produces byte-identical output (no trailing newline, `ensure_ascii=False`, `default=str`) and always writes UTF-8. The `os.replace` calls that move *directories* (claim/sent renames) are unchanged. The old code's missing `encoding=` was already fixed repo-wide by #1123; this PR consolidates the three drifted copies into one helper.

`docs/plans/2026-06-25-fsutil-migration.md` checklist updated: Stage 1 `write_resolved_manifest` ticked, `adapters/posix/mail.py` added to Stage 1 and ticked, Stage 3 `tools/soul/config.py` split out and ticked (psyche writers remain unchecked). Note: the mail service moved to `lingtai.adapters.posix.mail` during the Ports & Adapters refactor, so the issue's `services/mail.py` sites now live there. No ANATOMY entries referenced these writes.

No schema/data-structure/on-disk-format changes; unique dot-prefixed temp names (`_unique_tmp`, pid + uuid4, `"x"` exclusive create) make concurrent same-target writes collision-free and are invisible to the mail poll loops (they filter on directories / `message.json` presence).

## Validation

- `git diff --check` — clean.
- New golden/parity + concurrency coverage:
  - `tests/test_resolved_manifest_artifact.py::test_write_resolved_manifest_byte_identical_to_legacy_format` (frozen `generated_at`; byte-exact vs. legacy format incl. trailing newline and unescaped CJK)
  - `tests/test_resolved_manifest_artifact.py::test_write_resolved_manifest_concurrent_writers` (8 threads × 20 writes to one workdir: no silent `None`, valid artifact, no `*.tmp` litter)
  - `tests/test_soul.py::test_atomic_write_init_format_parity` (byte-exact legacy format, no temp sibling)
  - `tests/test_soul.py::test_atomic_write_init_error_string_on_failure` (read-only dir → error string, no exception escapes, no litter)
  - `tests/test_services_mail.py::test_send_writes_utf8_message_json` (CJK subject/body survive as raw UTF-8; no `message.json.tmp` sibling)
- Focused suites (`test_workdir.py`, `test_fsutil.py`, `test_services_mail.py`, `test_filesystem_mail.py`, `test_resolved_manifest_artifact.py`, `test_soul.py`, `test_tool_family_soul_migration.py`): **186 passed**, 2 failed — one is a pre-existing main failure (`test_artifact_redacts_api_key_like_secrets`, KeyError in the `_read_init` read path, identical on the base commit before this change, unrelated to the write mechanics) and one is a timing-sensitive listen test that passes in isolation and on repeated reruns.
- Broader mail/email/workdir regression net (`test_three_agent_email.py`, `test_mail_transport.py`, `test_layers_email.py`, `test_email_abs_reply_route.py`, `test_time_awareness_mail.py`, `test_tool_family_email_migration.py`, `test_workdir_lease.py`): **167 passed, 9 skipped**, 1 timing-sensitive flake that passes in isolation and 3/3 reruns.
- Environment: Python 3.14 venv with the package installed editable (`LINGTAI_SKIP_RUST_BUILD=1`); all changed files `py_compile` clean.

## Notes

- `tools/email/primitives.py` / `tools/email/manager.py` remain on the Stage 1 backlog (separate plan items, out of this issue's scope).
- `src/lingtai/agent.py::_activate_preset` still writes `init.json` via a fixed `init.json.tmp`; it is not one of the three writers listed in #756 and is left for a follow-up.

### DIFF
--- a/docs/plans/2026-06-25-fsutil-migration.md
+++ b/docs/plans/2026-06-25-fsutil-migration.md
@@ -60,7 +60,8 @@ These are migration *candidates*, not a commitment to change every one.
 ### Stage 1 — riskiest crash-sensitive / model-visible state (do first)
 - [x] `workdir.py::WorkingDir.write_manifest` (`.agent.json`) — **migrated here**
 - [x] `base_agent/__init__.py::_write_status_snapshot` (`.status.json`) — migrated with byte-parity coverage and opt-in existing-mode retention
-- [ ] `workdir.py::write_resolved_manifest` (`manifest.resolved.json`, trailing newline → pass explicit newline or keep inline)
+- [x] `workdir.py::write_resolved_manifest` (`manifest.resolved.json`, trailing newline kept via `atomic_write_text(json.dumps(...) + "\n")`; unique sibling temp removes the fixed-`.tmp` race) — migrated in the #756 PR
+- [x] `adapters/posix/mail.py` `message.json` writes (`PosixFilesystemMailAdapter` `send`, pseudo-agent claim delivery, runtime-probe ack) — migrated to `atomic_write_json` in the #756 PR
 - [x] notification files (`.notification/*.json`) — migrated to `atomic_write_json` (`adapters/posix/notification_store.py`)
 - [ ] `tools/email/primitives.py` message/`message.json` writes (currently direct `write_text`; some non-contact paths can escape non-ASCII)
 - [ ] `tools/email/manager.py` contact persistence (already atomic via `mkstemp`+`os.replace`; migrate for consistency)
@@ -73,7 +74,8 @@ These are migration *candidates*, not a commitment to change every one.
 
 ### Stage 3 — broader writes + retire local helpers
 - [ ] `migrate/*.py` write/replace sites
-- [ ] `tools/soul/config.py`, psyche snapshot/molt/archive writers
+- [x] `tools/soul/config.py` init.json persistence (`_atomic_write_init` → `atomic_write_text(fsync=True)`) — migrated in the #756 PR
+- [ ] psyche snapshot/molt/archive writers
 - [ ] remaining `read_json(default=...)`-style local readers → `_fsutil.read_json`
 
 Each checkbox above should land as (or within) its own PR with format-parity
--- a/src/lingtai/adapters/posix/mail.py
+++ b/src/lingtai/adapters/posix/mail.py
@@ -24,6 +24,7 @@ import uuid
 from pathlib import Path
 from typing import Callable, Iterator
 
+from lingtai.kernel._fsutil import atomic_write_json
 from lingtai.kernel.agent_presence import (
     is_agent as _presence_is_agent,
     observe_alive as _presence_observe_alive,
@@ -173,15 +174,12 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         else:
             msg_dir.mkdir(parents=True, exist_ok=True)
 
-        # Atomic write: tmp → rename
-        tmp_path = msg_dir / "message.json.tmp"
+        # Atomic write: unique sibling temp → os.replace (via _fsutil).
         final_path = msg_dir / "message.json"
         try:
-            tmp_path.write_text(
-                json.dumps(message, indent=2, ensure_ascii=False, default=str),
-                encoding="utf-8",
+            atomic_write_json(
+                final_path, message, indent=2, ensure_ascii=False, default=str
             )
-            os.replace(str(tmp_path), str(final_path))
         except OSError as e:
             return f"Failed to write message: {e}"
 
@@ -410,7 +408,6 @@ class PosixFilesystemMailAdapter(MailTransportPort):
             uuid_name = entry.name
             own_inbox_dir = self._inbox_dir / uuid_name
             own_msg_file = own_inbox_dir / "message.json"
-            own_tmp_file = own_inbox_dir / "message.json.tmp"
             claim_dir = sent_parent / f".{uuid_name}.claim-{uuid.uuid4()}"
             sent_dir = sent_parent / uuid_name
 
@@ -424,14 +421,12 @@ class PosixFilesystemMailAdapter(MailTransportPort):
             # Removed on rollback.
             self._seen.add(uuid_name)
 
-            # Step 1: write the payload to own inbox (tmp -> rename).
+            # Step 1: write the payload to own inbox (unique sibling temp -> rename).
             try:
                 own_inbox_dir.mkdir(parents=True, exist_ok=True)
-                own_tmp_file.write_text(
-                    json.dumps(payload, indent=2, ensure_ascii=False, default=str),
-                    encoding="utf-8",
+                atomic_write_json(
+                    own_msg_file, payload, indent=2, ensure_ascii=False, default=str
                 )
-                os.replace(str(own_tmp_file), str(own_msg_file))
             except OSError:
                 logger.warning(
                     "failed to write claimed pseudo-agent message %s to own inbox",
@@ -535,15 +530,12 @@ class PosixFilesystemMailAdapter(MailTransportPort):
         }
 
         inbox_dir = pseudo_dir / self._mailbox_rel / "inbox" / ack_id
-        tmp_file = inbox_dir / "message.json.tmp"
         final_file = inbox_dir / "message.json"
         try:
             inbox_dir.mkdir(parents=True, exist_ok=True)
-            tmp_file.write_text(
-                json.dumps(ack, indent=2, ensure_ascii=False, default=str),
-                encoding="utf-8",
+            atomic_write_json(
+                final_file, ack, indent=2, ensure_ascii=False, default=str
             )
-            os.replace(str(tmp_file), str(final_file))
         except OSError:
             logger.exception(
                 "failed to write runtime probe ack %s to %s",
--- a/src/lingtai/kernel/workdir.py
+++ b/src/lingtai/kernel/workdir.py
@@ -8,7 +8,6 @@ Core-owned ``lingtai.kernel.workdir_lease.WorkdirLeasePort`` and its production
 from __future__ import annotations
 
 import json
-import os
 import re
 from dataclasses import dataclass
 from datetime import datetime, timezone
@@ -105,10 +104,6 @@ class WorkdirLayout:
     def resolved_manifest(self) -> Path:
         return self.system_dir / _RESOLVED_MANIFEST_FILE
 
-    @property
-    def resolved_manifest_tmp(self) -> Path:
-        return self.system_dir / (_RESOLVED_MANIFEST_FILE + ".tmp")
-
     def notification_file(self, channel: str) -> Path:
         """Path to a ``.notification/<channel>.json`` file (no validation)."""
         return self.notification_dir / f"{channel}.json"
@@ -189,10 +184,12 @@ def write_resolved_manifest(working_dir: Path | str, data: dict) -> Path | None:
     (issue #259).
 
     Secrets (api_key/password/token-like fields) are removed recursively.
-    The write is atomic (``.tmp`` → ``os.replace``) and best-effort: returns
-    the artifact path on success, None when *data* has no manifest or the
-    write failed.
+    The write is atomic (unique sibling temp → ``os.replace`` via
+    ``_fsutil.atomic_write_text``) and best-effort: returns the artifact path
+    on success, None when *data* has no manifest or the write failed.
     """
+    from ._fsutil import atomic_write_text
+
     manifest = data.get("manifest") if isinstance(data, dict) else None
     if not isinstance(manifest, dict):
         return None
@@ -210,14 +207,11 @@ def write_resolved_manifest(working_dir: Path | str, data: dict) -> Path | None:
 
     try:
         layout = workdir_layout(working_dir)
-        layout.system_dir.mkdir(parents=True, exist_ok=True)
         target = layout.resolved_manifest
-        tmp = layout.resolved_manifest_tmp
-        tmp.write_text(
+        atomic_write_text(
+            target,
             json.dumps(artifact, indent=2, ensure_ascii=False) + "\n",
-            encoding="utf-8",
         )
-        os.replace(str(tmp), str(target))
         return target
     except (OSError, TypeError, ValueError):
         return None
--- a/src/lingtai/tools/soul/config.py
+++ b/src/lingtai/tools/soul/config.py
@@ -322,31 +322,24 @@ def _persist_soul_voice(agent, *, voice: str, voice_prompt: str) -> str | None:
 
 
 def _atomic_write_init(init_path, data) -> str | None:
-    """Write ``data`` to ``init_path`` via temp-file-then-rename.
+    """Write ``data`` to ``init_path`` atomically (unique sibling temp
+    via ``_fsutil.atomic_write_text``).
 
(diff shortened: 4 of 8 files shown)


huangzesen:

PR #1290, opened 2026-08-09
fix: make JSONC trailing-comma stripping string-aware

### LINKED ISSUE #725: load_jsonc trailing-comma stripping corrupts string values containing \", ]\" or \", }\"
## Summary

`load_jsonc()` in `src/lingtai_kernel/config_resolve.py` strips `//` comments in a string-aware way, but its trailing-comma pass (`re.sub(r',\s*([}\]])', r'\1', text)` at `config_resolve.py:38`) runs over the fully reassembled text — string literals included. Any config string value containing `, ]` or `, }` (or `,]` / `,}` with any whitespace between) is silently rewritten before parsing. Verified repro: `{"pattern": "a, ]b", "cmd": "x, }y"}` loads as `{'pattern': 'a]b', 'cmd': 'x}y'}` — no error, just corrupted data.

## Narrative

`load_jsonc` is the kernel's single entry point for reading preset manifests and other JSON/JSONC config. It is used by `src/lingtai_kernel/presets.py` (`presets.py:254,282`) to load every preset file, and by every migration module (`src/lingtai_kernel/migrate/m001_context_limit_relocation.py:51`, `src/lingtai_kernel/migrate/m002_description_object.py:79`). The function has to tolerate two JSONC-isms — `//` comments and trailing commas — on top of strict JSON.

The comment-stripping half of the job was already hardened. The docstring at `config_resolve.py:19-20` says it explicitly:

```
Comment stripping is string-aware: // inside a quoted JSON string value
is never treated as a comment (preserving URLs like "https://host/...").
```

The implementation tokenises the file into alternating string / non-string spans using `_JSONC_STRING_RE` (`config_resolve.py:13`, matching `"(?:[^"\\]|\\.)*"`), strips `//...EOL` only inside the non-string spans, and preserves string literals verbatim (`config_resolve.py:26-37`). This was done because a naive comment regex mangled URLs like `"https://host/..."` — the same class of bug, previously fixed for comments, and there is even a regression test for that fix (`tests/test_kernel_migrate.py:591`, `test_jsonc_migrations_preserve_double_slash_inside_strings`).

But then the very next line undoes the discipline:

```python
# config_resolve.py:37-39
text = ''.join(parts)
text = re.sub(r',\s*([}\]])', r'\1', text)
return json.loads(text)
```

The trailing-comma substitution runs on the *rejoined* text, so it happily matches inside the string literals the tokenizer just went out of its way to protect. A user hits this today with any string value containing a comma followed (possibly across whitespace, even newlines, since `\s` matches `\n`) by `]` or `}`. Realistic carriers in this codebase's configs: regex patterns and character classes (`"[a-z, ]"`), prompt or description snippets ("options: a, b, }"), shell/command arguments in MCP server specs, and JSON-ish examples embedded in system-prompt strings. The failure is the worst kind: `json.loads` succeeds, the config loads, and the value is quietly missing two characters. Nothing warns; the corruption surfaces far downstream as a regex that no longer matches or a prompt that reads wrong — and because migrations also read files through `load_jsonc`, a rewrite-on-migrate can bake the corruption back to disk permanently.

The design got here innocently: the tokenizer loop was introduced to fix the comment bug, and the pre-existing whole-text trailing-comma regex was left where it was, after the join. Better looks obvious in hindsight — both JSONC normalisations belong inside the tokenizer loop, applied only to non-string chunks, so string literals pass through the whole function byte-for-byte untouched.

## Current behavior

- `load_jsonc()` (`src/lingtai_kernel/config_resolve.py:16-39`) correctly protects string literals from comment stripping (lines 26-37) but then applies `re.sub(r',\s*([}\]])', r'\1', text)` to the reassembled text at line 38.
- Any string value matching `,\s*[}\]]` loses the comma and intervening whitespace. Confirmed with a temp file: `{"pattern": "a, ]b", "cmd": "x, }y"}` → `{'pattern': 'a]b', 'cmd': 'x}y'}`.
- The corruption is silent (no exception) and affects every `load_jsonc` caller: `presets.py` and the `migrate/m0xx` modules.
- One correct per-chunk observation for the fix: in valid JSON(C), a *genuine* trailing comma and its closing bracket can never be separated by a string literal (a string between them would make the comma a normal separator), so the comma and bracket of every real trailing comma always land in the same non-string chunk. Per-chunk substitution is therefore semantically equivalent to the current whole-text pass for all legal inputs — minus the string corruption.

1. **Refactor the non-string-chunk cleanup into one helper** in `src/lingtai_kernel/config_resolve.py`, e.g. a module-private function next to `_JSONC_STRING_RE`:

   ```python
   def _clean_jsonc_chunk(chunk: str) -> str:
       """Strip // comments and trailing commas from a non-string JSONC span."""
       chunk = re.sub(r'//[^\n]*', '', chunk)
       return re.sub(r',\s*([}\]])', r'\1', chunk)
   ```

2. **Use it inside the existing tokenizer loop** in `load_jsonc()`, replacing the two inline `re.sub(r'//[^\n]*', ...)` calls (`config_resolve.py:30` and `config_resolve.py:36`) with `_clean_jsonc_chunk(...)`:

   ```python
   parts: list[str] = []
   pos = 0
   for m in _JSONC_STRING_RE.finditer(text):
       parts.append(_clean_jsonc_chunk(text[pos:m.start()]))
       parts.append(m.group())          # string literal: verbatim
       pos = m.end()
   parts.append(_clean_jsonc_chunk(text[pos:]))
   return json.loads(''.join(parts))
   ```

3. **Delete the whole-text pass** at `config_resolve.py:38` (`text = re.sub(r',\s*([}\]])', r'\1', text)`).

4. **Update the docstring** (`config_resolve.py:17-21`) to state that *both* comment stripping and trailing-comma removal are string-aware, since the current docstring only claims it for comments.

5. **Ordering note (keep as in the sketch above):** comment stripping must run before the trailing-comma sub within each chunk, so `,  // note` followed by `]` on the next line still collapses correctly (`\s` in the comma regex spans the newline left behind by comment removal).

6. **No schema, data-structure, or migration changes.** The function signature and all call sites (`presets.py`, `migrate/m001_context_limit_relocation.py`, `migrate/m002_description_object.py`) are untouched. Behavior only changes for inputs that were previously being corrupted; genuine trailing commas are handled identically (see the per-chunk equivalence argument in Current behavior). Files that were parseable before remain parseable.

## Risks and alternatives

- **Risk — split comma/bracket across chunks:** if a comma and its closing bracket were separated by a string literal, per-chunk substitution would miss the "trailing" comma. As argued above, that shape is impossible for a genuine trailing comma in well-formed JSONC (`, "str" ]` means the comma separates elements). Inputs where it changes anything were malformed already and will now raise `json.JSONDecodeError` instead of being silently patched — arguably an improvement.
- **Risk — configs that accidentally relied on the corruption:** a file whose string value contains `, ]` would previously load with the characters removed; after the fix the value round-trips faithfully. This is the intended bug fix, but downstream consumers comparing against the corrupted value (unlikely; none known) would see a change.
- **Alternative A — full JSON5/JSONC parser dependency** (`json5`, `pyjson5`): correct by construction, but adds a runtime dependency to the kernel for two small normalisations that a 15-line tokenizer already handles; the repo has deliberately kept this helper dependency-free.
- **Alternative B — a hand-written character-by-character scanner** tracking in-string state: strictly more code for the same result; the existing `_JSONC_STRING_RE` tokenizer (which correctly handles escapes via `(?:[^"\\]|\\.)*`) is already the string-awareness mechanism this file trusts for comments.
- **Alternative C — leave line 38 but make the regex "string-aware" with lookarounds:** regexes cannot reliably determine in-string context; this is exactly the trap the comment-stripping fix escaped by tokenising. Reusing the tokenizer wins on correctness and consistency.

## Testing plan

Add a dedicated test module `tests/test_config_resolve_jsonc.py` (there is currently no direct unit coverage of `load_jsonc`; it is only exercised indirectly via `tests/test_presets.py:175` and `tests/test_kernel_migrate.py:569,591`):

1. `test_load_jsonc_preserves_comma_bracket_inside_strings` — write `{"pattern": "a, ]b", "cmd": "x, }y"}` to a tmp file; assert `load_jsonc` returns exactly `{"pattern": "a, ]b", "cmd": "x, }y"}`. This is the direct regression test for this issue and fails on current `main`.
2. `test_load_jsonc_preserves_comma_whitespace_bracket_inside_strings` — value containing `",\n  ]"` (comma, newline, spaces, bracket inside a string) survives verbatim, covering the `\s*` span.
3. `test_load_jsonc_still_strips_trailing_commas` — input `{"a": [1, 2, ], "b": {"c": 1, }, }` parses to `{"a": [1, 2], "b": {"c": 1}}`, proving the legitimate JSONC behavior is unchanged.
4. `test_load_jsonc_trailing_comma_before_comment` — input like `{"a": 1,  // note\n}` parses to `{"a": 1}`, locking in the comment-before-comma ordering (step 5 of the plan).
5. `test_load_jsonc_comment_and_comma_features_compose_with_strings` — a combined fixture with a URL value (`"https://host/x"`), a `, ]`-bearing value, a real `//` comment, and a trailing comma, asserting all four behaviors simultaneously; complements the existing URL-preservation test at `tests/test_kernel_migrate.py:591`.
6. Optional escape-hardening case: value `"end \\", ]"` (escaped quote before the pattern) survives, exercising `_JSONC_STRING_RE`'s escape handling under the new per-chunk path.

Run `pytest tests/test_config_resolve_jsonc.py tests/test_presets.py tests/test_kernel_migrate.py` to confirm the fix plus no regressions in preset loading and migrations.

huangzesen:
Fixes #725

## What

`parse_jsonc()` (used by `load_jsonc()` — the kernel's single entry point for reading preset manifests, `init.jsonc`, and migration inputs) applies the trailing-comma regex `re.sub(r',\s*([}\]])', ...)` to the **fully reassembled text**, string literals included. Any string value containing `, ]` or `, }` (or `,]` / `,}` with any whitespace between) is silently rewritten before parsing:

```python
>>> parse_jsonc('{"pattern": "a, ]b", "cmd": "x, }y"}')
{'pattern': 'a]b', 'cmd': 'x}y'}   # old — silent corruption, no error
```

## Fix

Move trailing-comma removal into a string-aware pass that only touches the non-string spans of the text, mirroring the existing string-aware comment stripping:

- `_strip_comments(text)` — the existing comment-stripping state machine, extracted as a helper.
- `_strip_trailing_commas(text)` — walks the text, applies `,\s*([}\]]) -> \1` only to non-string spans, and copies string literals byte-for-byte.
- `parse_jsonc()` = `_strip_comments` → `_strip_trailing_commas` → `json.loads`.

Genuine trailing commas always sit in the same non-string span as their closing bracket, so the pass is semantically identical to the old whole-text regex for every parseable input (including `,  // comment
}` — comment stripping still runs first). No schema, signature, or call-site changes.

## Tests

New `tests/test_config_resolve_jsonc.py` (direct unit coverage of `parse_jsonc`/`load_jsonc`, which previously only had indirect coverage):

- `test_load_jsonc_preserves_comma_bracket_inside_strings` — the #725 repro now round-trips exactly (fails on old code).
- `test_load_jsonc_preserves_multi_space_bracket_inside_strings` — `,   ]` / `,    }` inside strings survive.
- `test_parse_jsonc_string_aware_across_real_whitespace` — the comma pass never rewrites string content even when the span would cover newline/tab; malformed inputs now raise `JSONDecodeError` instead of being silently patched.
- `test_parse_jsonc_still_strips_trailing_commas` / `test_parse_jsonc_trailing_comma_before_comment` — legitimate JSONC behavior unchanged.
- `test_parse_jsonc_features_compose_with_strings` — URLs, `, ]`-bearing strings, comments (incl. quoted examples) and trailing commas compose.
- `test_parse_jsonc_preserves_escaped_quote_before_pattern` / `test_parse_jsonc_preserves_double_slash_inside_strings` — escape handling and URL preservation lock-ins.

### Evidence

```
$ PYTHONPATH=src python3 -m pytest tests/test_config_resolve_jsonc.py tests/test_presets.py tests/test_kernel_migrate.py tests/test_init_reader.py tests/test_inherit_fallback.py -q
160 passed in 0.80s

$ PYTHONPATH=src python3 -m pytest tests/test_agent_preset_manifest.py tests/test_preset_materialization.py tests/test_agent_config_hydration.py tests/test_preset_context_guard.py -q
86 passed in 4.88s
```

(Note: `tests/test_wechat_config_resolution.py` cannot be collected in this environment due to a pre-existing `mcp` SDK version mismatch — `ServerRequestContext` import error — unrelated to this change.)

### DIFF
--- a/src/lingtai/kernel/config_resolve.py
+++ b/src/lingtai/kernel/config_resolve.py
@@ -7,17 +7,17 @@ import re
 from pathlib import Path
 
 
-def parse_jsonc(text: str) -> dict:
-    """Parse JSON or JSONC text (strips // comments and trailing commas).
+_TRAILING_COMMA_RE = re.compile(r',\s*([}\]])')
 
-    Pure text→object transform with no I/O, so callers holding raw text (e.g.
-    migration transforms) use it directly. Comment stripping is string-aware: //
-    inside a quoted string is never a comment (URLs like "https://host/..." survive).
+
+def _strip_comments(text: str) -> str:
+    """Strip ``//`` comments while preserving string literals verbatim.
+
+    Implemented as a small state machine rather than splitting around string
+    literals: JSONC comments may themselves contain quoted examples (for
+    example ``// copy as "init.json"``); those quotes must not become JSON data
+    or hide the real delimiters on the following lines.
     """
-    # Strip comments with a small state machine rather than splitting around
-    # string literals. JSONC comments may themselves contain quoted examples
-    # (for example `// copy as "init.json"`); those quotes must not become JSON
-    # data or hide the real delimiters on the following lines.
     parts: list[str] = []
     in_string = False
     escaped = False
@@ -52,8 +52,59 @@ def parse_jsonc(text: str) -> dict:
             continue
         parts.append(ch)
         i += 1
-    text = "".join(parts)
-    text = re.sub(r',\s*([}\]])', r'\1', text)
+    return "".join(parts)
+
+
+def _strip_trailing_commas(text: str) -> str:
+    """Remove trailing commas without touching string literals.
+
+    Equivalent to applying ``re.sub(r',\\s*([}\\]]])', r'\\1', text)`` only to the
+    non-string spans of *text*, so a string value containing `, ]` or `, }`
+    (or `,]` / `,}`) survives byte-for-byte instead of being silently
+    rewritten. Genuine trailing commas always sit in the same non-string span
+    as their closing bracket, so this is semantically identical to the old
+    whole-text pass for every parseable input.
+    """
+    out: list[str] = []
+    pos = 0
+    in_string = False
+    escaped = False
+    i = 0
+    n = len(text)
+    while i < n:
+        ch = text[i]
+        if in_string:
+            out.append(ch)  # copy the string literal verbatim
+            if escaped:
+                escaped = False
+            elif ch == "\\":
+                escaped = True
+            elif ch == '"':
+                in_string = False
+                pos = i + 1
+            i += 1
+            continue
+        if ch == '"':
+            out.append(_TRAILING_COMMA_RE.sub(r"\1", text[pos:i]))
+            out.append(ch)  # opening quote starts the string literal
+            in_string = True
+            i += 1
+            continue
+        i += 1
+    out.append(_TRAILING_COMMA_RE.sub(r"\1", text[pos:]))
+    return "".join(out)
+
+
+def parse_jsonc(text: str) -> dict:
+    """Parse JSON or JSONC text (strips // comments and trailing commas).
+
+    Pure text→object transform with no I/O, so callers holding raw text (e.g.
+    migration transforms) use it directly. Both normalisations are string-aware:
+    // inside a quoted string is never a comment (URLs like "https://host/..."
+    survive), and a string value containing `, ]` or `, }` is left intact.
+    """
+    text = _strip_comments(text)
+    text = _strip_trailing_commas(text)
     return json.loads(text)
 
 
--- a/tests/test_config_resolve_jsonc.py
+++ b/tests/test_config_resolve_jsonc.py
@@ -1,0 +1,92 @@
+"""Direct unit coverage for the string-aware JSONC reader.
+
+``parse_jsonc`` / ``load_jsonc`` are the kernel's single entry point for
+reading preset manifests, init config, and migration inputs. These tests lock
+in the string-aware handling of both JSONC normalisations (``//`` comments and
+trailing commas): string values must survive byte-for-byte, even when they
+contain text that looks like a comment marker or a trailing-comma pattern.
+
+Regression: lingtai-kernel#725 — the whole-text trailing-comma regex used to
+corrupt string values containing `, ]` / `, }` (or `,]` / `,}`).
+"""
+
+import json
+
+import pytest
+
+from lingtai.kernel.config_resolve import (
+    _strip_trailing_commas,
+    load_jsonc,
+    parse_jsonc,
+)
+
+
+def test_load_jsonc_preserves_comma_bracket_inside_strings(tmp_path):
+    """String values containing `, ]` / `, }` are not rewritten (#725)."""
+    p = tmp_path / "carriers.jsonc"
+    p.write_text('{"pattern": "a, ]b", "cmd": "x, }y"}', encoding="utf-8")
+    assert load_jsonc(p) == {"pattern": "a, ]b", "cmd": "x, }y"}
+
+
+def test_load_jsonc_preserves_multi_space_bracket_inside_strings(tmp_path):
+    """Multiple spaces between comma and bracket inside a string survive."""
+    p = tmp_path / "ws.jsonc"
+    p.write_text('{"pattern": "a,   ]b", "cmd": "x,    }y"}', encoding="utf-8")
+    assert load_jsonc(p) == {"pattern": "a,   ]b", "cmd": "x,    }y"}
+
+
+def test_parse_jsonc_string_aware_across_real_whitespace():
+    """The comma pass never rewrites string content, even when the span
+    would cover real control whitespace (newline or tab).
+
+    Real control characters inside a string are invalid JSON, so those inputs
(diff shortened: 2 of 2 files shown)


huangzesen:

PR #1291, opened 2026-08-09
fix(nudge): run kernel_version remote probe off the heartbeat thread (#730)

### LINKED ISSUE #730: Synchronous PyPI fetch in kernel_version nudge blocks the 1s heartbeat loop past the 2s is_alive threshold
## Summary

The heartbeat thread — the only writer of `.agent.heartbeat` — also runs the nudge checks inline, and the `kernel_version` nudge performs a blocking `urllib.request.urlopen` call to PyPI with a 3.0s timeout once per UTC day. `handshake.is_alive` treats a heartbeat older than 2.0s as dead, so during the remote fetch every liveness consumer (mail delivery, karma actions, `cpr`) can observe a perfectly healthy agent as dead. The fetch should run on a background thread so `check()` never blocks the heartbeat tick.

## Narrative

Every LingTai agent runs a heartbeat daemon thread (`_heartbeat_loop` in `src/lingtai_kernel/base_agent/lifecycle.py:357`) whose contract is simple: write a fresh timestamp to `.agent.heartbeat` once per second. That file is the network-wide liveness signal. `handshake.is_alive` (`src/lingtai_kernel/handshake.py:43`) reads it with a default `threshold: float = 2.0` — if the newest timestamp is more than two seconds old, the agent is considered dead. That verdict gates real behavior: `services/mail.py:161` checks `is_alive(recipient_dir)` before delivering mail, and `intrinsics/system/karma.py` consults it in at least seven places for sleep/suspend/cpr decisions (e.g. lines 102, 117, 131, 153, 173, 191, 200).

The same heartbeat loop, after writing its tick, does a batch of housekeeping: signal-file detection, rules checks, and — the piece at issue here — nudge dispatch (`src/lingtai_kernel/base_agent/lifecycle.py:522-524`):

```python
try:
    from ..nudge import run_checks as _run_nudge_checks
    _run_nudge_checks(agent)
```

The comment above this block says the dispatcher wraps individual calls "so a misbehaving check cannot block the heartbeat loop", but `_run_one` in `src/lingtai_kernel/nudge/__init__.py:56` only wraps checks in `try/except`. That protects against a check that *raises*; it does nothing against a check that is *slow*. Latency inside any check is latency between heartbeat writes.

And one check is slow by design. The `kernel_version` nudge (`src/lingtai_kernel/nudge/kernel_version.py`) probes whether a newer `lingtai` kernel package has been published. When the daily remote check is due (`_remote_check_due`, line 285), `check()` calls `_fetch_latest_version()` synchronously (line 155), which does:

```python
def _fetch_latest_version() -> str:
    req = urllib.request.Request(
        _PYPI_JSON_URL,
        headers={"User-Agent": "lingtai-kernel-nudge/1"},
    )
    with urllib.request.urlopen(req, timeout=_REMOTE_TIMEOUT_SECONDS) as resp:
        raw = resp.read()
```

with `_REMOTE_TIMEOUT_SECONDS = 3.0` (line 33). A slow-but-successful fetch already stalls the loop past the 2.0s `is_alive` threshold. The failure mode is worse than 3 seconds, though: `urllib`'s timeout bounds socket operations, but DNS resolution (`getaddrinfo`) is not reliably covered by it on all platforms — a stalled resolver on a flaky network can hang the heartbeat thread for many seconds.

The user-visible symptom is a once-per-UTC-day flicker of apparent death: an agent that is idle and healthy briefly reads as dead. Mail sent in that window is refused or misrouted by the liveness gate in `services/mail.py`; a karma `cpr` or lifecycle probe that lands in the window draws the wrong conclusion about a process that never stopped running. Because the window is short, network-dependent, and occurs at most daily per agent (plus retries on fetch errors, since `_remote_check_due` re-arms when the date or installed version changes), this is exactly the kind of Heisenbug that produces unexplained one-off delivery failures.

The design got here honestly: the heartbeat loop was the natural existing 1s cadence to hang periodic checks off, and all the other nudges (`source_drift`, `goal`) are local-only and cheap — `kernel_version` is the only check that touches the network. Better looks like a hard invariant: **nothing on the heartbeat thread may block on the network.** The minimal fix is to make the PyPI fetch asynchronous — spawn a daemon thread when the daily check is due, return immediately, and consume the result on a later tick.

## Current behavior

- `_heartbeat_loop` (`src/lingtai_kernel/base_agent/lifecycle.py:357`) writes `.agent.heartbeat` via `_write_heartbeat_tick` (line 372, defined at line 646), then runs `run_checks(agent)` inline (lines 522-524), then sleeps 1.0s (line 611). Total inter-tick latency = 1s + everything in between.
- `run_checks` (`src/lingtai_kernel/nudge/__init__.py:44`) dispatches `kernel_version.check` synchronously; `_run_one` only catches exceptions.
- `kernel_version.check` (`src/lingtai_kernel/nudge/kernel_version.py:51`) self-throttles to one probe per 60s, and when `_remote_check_due` is true (first probe of a UTC day, or a new installed version) calls `_fetch_latest_version()` (line 155) — a blocking HTTP GET to `https://pypi.org/pypi/lingtai/json` with a 3.0s socket timeout, DNS not bounded.
- `is_alive` (`src/lingtai_kernel/handshake.py:43`) defaults to `threshold=2.0`. Consumers: `services/mail.py:161` (delivery gate) and `intrinsics/system/karma.py` (sleep/suspend/cpr logic).
- Net effect: once per UTC day per agent (and on each error-day retry), the heartbeat file goes stale for up to ~3s (unbounded if DNS stalls), and `is_alive` returns `False` for a live agent.

Keep the fix local to `src/lingtai_kernel/nudge/kernel_version.py` so all persistent-state reads/writes (`.notification/.nudge_state.json`) stay on the heartbeat thread; the worker thread only produces a value.

1. **Add an in-memory fetch slot on the agent** (no schema change). In `kernel_version.py`, add a small holder plus helpers:

   ```python
   @dataclass
   class _PendingFetch:
       thread: threading.Thread
       started_ts: float
       result: str | None = None   # written only by worker
       error: str | None = None    # written only by worker
       done: threading.Event = field(default_factory=threading.Event)

   def _fetch_slot(agent) -> _PendingFetch | None:
       return getattr(agent, "_nudge_kernel_fetch", None)
   ```

2. **Add `_start_fetch(agent) -> None`** that spawns a daemon thread:

   ```python
   def _start_fetch(agent) -> None:
       pending = _PendingFetch(thread=None, started_ts=time.time())
       def _worker() -> None:
           try:
               pending.result = _fetch_latest_version()
           except Exception as e:
               pending.error = str(e)[:200]
           finally:
               pending.done.set()
       t = threading.Thread(target=_worker, daemon=True,
                            name="nudge-kernel-version-fetch")
       pending.thread = t
       agent._nudge_kernel_fetch = pending
       t.start()
   ```

3. **Rework the remote-check tail of `check()`** (currently `kernel_version.py:150-210`). Replace the synchronous `latest = _fetch_latest_version()` block with a small state machine, factored into `_handle_remote_check(agent, info, kernel_state, persistent, today)`:
   - No fetch in flight and `_remote_check_due(...)` → `_start_fetch(agent)`; return (heartbeat tick completes immediately).
   - Fetch in flight, `done` not set → if `time.time() - started_ts` exceeds a hard deadline (e.g. `_FETCH_DEADLINE_SECONDS = 30.0`), abandon it: record `last_error="fetch timed out"` plus `last_remote_check_date=today` in `kernel_state`, clear `agent._nudge_kernel_fetch`, save state (the orphaned daemon thread dies on its own). Otherwise return and check again next probe.
   - Fetch done → clear the slot, then run the *existing* logic verbatim with `pending.result` / `pending.error` in place of the direct call: on error, update `last_remote_check_date` / `last_error` and `_log("kernel_version_update_check_error", ...)`; on success, update `latest_seen`, compare with `_is_newer`, and `upsert`/`remove` the nudge exactly as today.
   - **Single-flight guarantee**: never call `_start_fetch` while a slot exists; the 60s `_FAST_INTERVAL_SECONDS` throttle at the top of `check()` naturally paces the polling of the slot.

4. **Note on latency of emission**: with the 60s probe throttle, the nudge now appears one probe (~60s) after the fetch completes rather than in the same tick. For an at-most-daily advisory notification this is acceptable; document it in the module docstring's cadence notes.

5. **Fix the misleading comment** in `src/lingtai_kernel/base_agent/lifecycle.py:517-521`: the dispatcher wrapper protects against *exceptions*, not *latency*. Reword to state the invariant explicitly: "checks must never block on the network; long work must be handed to a background thread (see `nudge/kernel_version.py`)". Add the same rule to `nudge/ANATOMY.md` so future checks don't reintroduce the bug.

6. **Back-compat / migration**: none needed. `.nudge_state.json` keys are unchanged (`last_remote_check_date`, `checked_installed_version`, `latest_seen`, `last_error`, `emitted_for_latest`). The fetch slot is process-local; a refresh/restart mid-fetch simply re-arms the daily check, which is idempotent.

## Risks and alternatives

- **Thread-safety of state file**: the worker thread must not touch `.nudge_state.json` or the notification file — the plan enforces this by having it write only to the `_PendingFetch` fields, with `done` as the publication barrier. All `_save_persistent_state` / `upsert` calls stay on the heartbeat thread, preserving the current single-writer property.
- **Orphaned worker on shutdown**: the thread is a daemon and holds no locks or file handles beyond the socket, so interpreter exit is unaffected. The 30s abandon deadline prevents a wedged DNS lookup from pinning the slot forever.
- **Alternative A — raise the `is_alive` threshold to ~5s**: papers over this one check but weakens liveness detection network-wide, and still loses to DNS stalls that exceed the socket timeout. Rejected.
- **Alternative B — move all nudge dispatch (or all post-tick housekeeping) off the heartbeat thread to a dedicated maintenance thread**: the most robust design — the heartbeat thread would do nothing but write the file — but it is a larger structural change touching signal-file detection, notification sync, and the AED/STUCK logic that currently share the loop, with its own new races (two threads consuming signal files). Worth a follow-up issue; not required to fix this bug. The chosen plan fixes the only network-touching check (`source_drift` and `goal` are local-only) with a small, testable diff.
- **Alternative C — do the fetch in a subprocess**: heavier (spawn cost per day per agent) with no benefit over a daemon thread for a read-only GET.

## Testing plan

Extend `tests/test_kernel_version_nudge.py`:

- `test_check_does_not_block_on_slow_fetch`: monkeypatch `_fetch_latest_version` to block on a `threading.Event`; call `check(agent)` with the remote check due and assert it returns in well under `_REMOTE_TIMEOUT_SECONDS` (e.g. < 0.5s) and that no nudge/state update has happened yet.
- `test_fetch_result_consumed_on_later_probe`: continue from the above — set the event so the worker completes with a newer version, advance past `_FAST_INTERVAL_SECONDS` (monkeypatch `time.time`), call `check()` again, and assert the `kernel_version` nudge is upserted and `latest_seen` / `last_remote_check_date` are recorded.
- `test_fetch_error_recorded_asynchronously`: worker raises; assert the next probe writes `last_error` and `last_remote_check_date` (so the day is not retried in a loop) and emits `kernel_version_update_check_error`.
- `test_single_flight`: while a fetch is in flight, repeated due `check()` calls spawn no second thread (count via a monkeypatched `_start_fetch` or by asserting the slot object is identical).
- `test_fetch_deadline_abandons_stuck_worker`: worker never sets `done`; monkeypatch time past `_FETCH_DEADLINE_SECONDS` and assert the slot is cleared and `last_error` mentions the timeout.
- Regression guard in `tests/test_base_agent.py` (or a new `tests/test_heartbeat_latency.py`): run a real `_heartbeat_loop` iteration with `kernel_version._fetch_latest_version` patched to sleep 3s and assert consecutive `.agent.heartbeat` timestamps never diverge by more than the `is_alive` threshold.

huangzesen:
## Summary

**Root cause.** `_heartbeat_loop` (`base_agent/lifecycle.py`) is the sole writer of the liveness heartbeat, and `agent_presence.is_alive` treats a tick older than the default `threshold=2.0`s as *dead* — a verdict that gates mail delivery and karma sleep/suspend/cpr decisions. That same loop runs nudge checks inline, and `kernel_version.check()` called `_fetch_latest_version()` **synchronously**: a blocking `urllib` GET chain against the GitHub/Gitee release manifests (up to 4 bounded requests × 3.0s socket timeout, DNS not reliably bounded). The dispatcher's `try/except` (`nudge/__init__.py:_run_one`) only guards a check that *raises*, never one that is *slow* — so a slow-but-successful fetch stalled the heartbeat tick past the 2.0s threshold and a healthy idle agent briefly read as dead.

Note: since the issue was filed, the remote probe moved from the once-per-UTC-day PyPI JSON fetch to a GitHub/Gitee release-manifest comparison that runs on **every 60s probe** (`_remote_check_due` is now dead code that always returns `True`), so the blocking window is more frequent than the issue describes — the fix addresses the current code.

**Fix, at the owning layer.** A network probe has no business on the single-writer heartbeat thread. Rather than weaken the network-wide liveness threshold (which would still lose to DNS stalls), the fix makes the fetch asynchronous inside the nudge that owns the I/O (`src/lingtai/kernel/nudge/kernel_version.py`):

- The due `check()` spawns a short-lived daemon worker (`_start_fetch`) that writes only into a process-local `_PendingFetch` slot and returns immediately — the heartbeat tick completes at once (measured: 2.7ms instead of ≥3s).
- A later probe (past the existing 60s fast gate) consumes the result **on the heartbeat thread** and emits/clears the nudge exactly as the synchronous path did, including the mirror-mismatch and error branches.
- **Single-flight:** no second worker spawns while a slot exists; a wedged worker (e.g. hung `getaddrinfo`) is abandoned after `_FETCH_DEADLINE_SECONDS = 30.0` with a recorded `"fetch timed out"`.
- All `.nudge_state.json` writes stay on the heartbeat thread, preserving the single-writer property; the worker only produces a value.

**Trade-off / non-goal.** The advisory nudge now surfaces up to one probe (~60s) after the fetch completes rather than in the same tick — acceptable for an at-most-every-60s advisory, documented in the module docstring and `nudge/ANATOMY.md`. Moving *all* post-tick housekeeping off the heartbeat thread (Alternative B in #730) is a larger structural change with its own races and remains a deliberate non-goal.

The previously-incorrect lifecycle comment ("a misbehaving check cannot block the loop") and `nudge/ANATOMY.md` now state the real invariant: **checks must never block on the network; long work goes to a background thread.**

## Validation

Failing-first — the new regression tests fail on unmodified `kernel_version.py`:

```
$ python -m pytest tests/test_kernel_version_nudge.py::test_check_does_not_block_on_slow_fetch tests/test_heartbeat.py::TestHeartbeatNeverBlocksOnNetwork -q
2 failed in 9.80s   # check() blocked past the threshold; heartbeat went stale
```

After the fix:

```
$ python -m pytest tests/test_kernel_version_nudge.py tests/test_heartbeat.py -q
49 passed in 25.50s

$ python -m pytest tests/test_base_agent.py tests/test_nudge_inline_cap.py tests/test_nudge_policy.py \
    tests/test_nudge_prompts.py tests/test_agent_presence.py tests/test_anatomy_drift_checker.py -q
91 passed, 1 failed in 0.67s   # the 1 failure is pre-existing on main, unrelated

$ python -m pytest tests/test_agent_presence.py tests/test_anatomy_drift_checker.py tests/test_nudge_policy.py -q
55 passed in 0.27s
```

(The `test_nudge_inline_cap.py::test_dismissal_round_trip_for_capped_finding` failure reproduces identically with `src` restored to HEAD — pre-existing baseline drift, not introduced here.)

- Direct repro before fix: `check()` blocked the heartbeat thread for 3.00s with a 3s fetch (is_alive threshold 2.0s → dead-flicker). After fix: 2.7ms.
- End-to-end: real `_heartbeat_loop` with a 3s fetch — `.agent.heartbeat` stays fresh throughout (test run 3×, stable).
- `python -m compileall -q src/lingtai/kernel/nudge src/lingtai/kernel/base_agent` — ok.
- `git diff --check` — clean.
- Anatomy drift checker: same 3 pre-existing items as baseline, none introduced by this change.

Fixes #730

### DIFF
--- a/src/lingtai/kernel/base_agent/lifecycle.py
+++ b/src/lingtai/kernel/base_agent/lifecycle.py
@@ -532,9 +532,11 @@ def _heartbeat_loop(agent) -> None:
         # Per-agent periodic checks that publish to `.notification/nudge.json`
         # when something needs the agent's attention (e.g. a newer lingtai
         # wheel is installed on disk than the version this process imported).
-        # Each check throttles itself; the dispatcher wraps individual calls
-        # so a misbehaving check cannot block the heartbeat loop. See
-        # `nudge/ANATOMY.md`.
+        # Each check throttles itself; the dispatcher only guards against
+        # checks that *raise* -- latency inside a check is latency between
+        # heartbeat writes. Invariant: checks must never block on the network;
+        # long work is handed to a background thread (see
+        # `nudge/kernel_version.py`). See `nudge/ANATOMY.md`.
         try:
             from ..nudge import run_checks as _run_nudge_checks
             _run_nudge_checks(agent)
--- a/src/lingtai/kernel/nudge/ANATOMY.md
+++ b/src/lingtai/kernel/nudge/ANATOMY.md
@@ -67,7 +67,10 @@ the ordinary Notification Store channel; it does not create a second transport.
   `LINGTAI_NUDGE_FOLDER_SIZE_GB` (default `5` decimal GB) (`src/lingtai/kernel/nudge/folder_size.py:1-182`).
 - `kernel_version.py` — read-only installed/running observation plus bounded
   GitHub/Gitee release-manifest comparison; it does not own a product repeat
-  cadence (`src/lingtai/kernel/nudge/kernel_version.py:91-229`).
+  cadence (`src/lingtai/kernel/nudge/kernel_version.py:91-229`). Its remote
+  probe runs on a daemon worker thread and is consumed on a later probe, so
+  the 1s heartbeat tick never blocks on the network; the nudge surfaces up to
+  one probe (~60s) after the fetch completes.
 - `source_drift.py` — read-only runtime/source comparison in a separate
   integrity/diagnostic channel; it remains active for editable/source/dev
   runtimes (`src/lingtai/kernel/nudge/source_drift.py:21-111`).
@@ -81,7 +84,9 @@ the ordinary Notification Store channel; it does not create a second transport.
 
 ## Connections
 
-`base_agent/lifecycle.py:_heartbeat_loop` calls `run_checks` once per heartbeat;
+`base_agent/lifecycle.py:_heartbeat_loop` calls `run_checks` once per heartbeat
+(checks must never block on the network — long work goes to a background
+thread, see `kernel_version.py`);
 protected goal reminders are dispatched separately by
 `run_system_notifications`. Producer checks call `upsert`/`remove`; the shared
 `NotificationStorePort` persists `nudge.json`. `notification(action="dismiss_channel", input={"channel": "nudge", ...},
--- a/src/lingtai/kernel/nudge/kernel_version.py
+++ b/src/lingtai/kernel/nudge/kernel_version.py
@@ -4,15 +4,21 @@ This check is deliberately read-only. It surfaces the observed running and
 installed versions, plus the latest release-manifest facts published on the
 official GitHub/Gitee mirrors, through the shared ``nudge`` notification
 channel. It never mutates an installation.
+
+Cadence note: the remote probe runs on a daemon worker thread and is consumed
+on a later probe, so the nudge surfaces up to one probe (~60s) after the
+fetch completes. This is intentional: the check runs on the 1s heartbeat
+thread, which must never block on the network (#730).
 """
 
 from __future__ import annotations
 
 import hashlib
 import json
+import threading
 import time
 import urllib.request
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Mapping
@@ -27,6 +33,10 @@ from .prompts import (
 
 _FAST_INTERVAL_SECONDS = 60.0
 _REMOTE_TIMEOUT_SECONDS = 3.0
+# Hard ceiling for one background probe. Longer than the worst-case sequential
+# mirror fetch (4 bounded requests x 3.0s); a worker still running after this
+# (e.g. a stalled DNS lookup) is abandoned so the slot never pins forever.
+_FETCH_DEADLINE_SECONDS = 30.0
 _MAX_REMOTE_BYTES = 256 * 1024
 _MANIFEST_ASSET_NAME = "lingtai-kernel-release-manifest.json"
 _GITHUB_LATEST_RELEASE_URL = (
@@ -78,6 +88,63 @@ class _MirrorMismatchError(RuntimeError):
             for source, manifest in sorted(self.manifests.items())
         )
         super().__init__(f"release-manifest mirrors disagree: {versions}")
+
+
+@dataclass
+class _PendingFetch:
+    """Process-local slot for one in-flight remote probe.
+
+    ``result``/``error`` are written only by the daemon worker thread and read
+    by the heartbeat thread after ``done`` is set; ``done`` is the publication
+    barrier. The worker never touches ``.nudge_state.json`` or the
+    notification store, preserving the heartbeat thread's single-writer
+    property over persistent state.
+    """
+
+    thread: threading.Thread | None
+    started_ts: float
+    result: Any = None
+    error: BaseException | None = None
+    done: threading.Event = field(default_factory=threading.Event)
+
+
+def _fetch_slot(agent) -> _PendingFetch | None:
+    """Return the in-flight fetch slot for ``agent``, if any."""
+    return getattr(agent, "_nudge_kernel_fetch", None)
+
+
+def _start_fetch(agent) -> None:
+    """Start the remote probe on a daemon worker and return immediately.
+
+    The heartbeat tick must never block on the network (#730): a slow or
+    stalled fetch would hold the 1s tick past the 2s ``is_alive`` threshold
+    and make a live agent look dead. Single-flight is guaranteed because a
+    slot exists from spawn until consume/abandon, and ``check`` is gated by
+    the 60s fast interval.
+    """
+    pending = _PendingFetch(thread=None, started_ts=time.time())
+
+    def _worker() -> None:
+        try:
+            pending.result = _fetch_latest_version()
+        except Exception as exc:  # keep the slot consumable on any failure
+            pending.error = exc
+        finally:
+            pending.done.set()
+
+    try:
+        worker = threading.Thread(
+            target=_worker,
+            daemon=True,
+            name="nudge-kernel-version-fetch",
+        )
+    except Exception:
+        # Cannot even spawn the worker; stay inert and let the next probe
+        # re-arm the remote check.
+        return
+    pending.thread = worker
+    agent._nudge_kernel_fetch = pending
+    worker.start()
 
 
 def check(agent) -> None:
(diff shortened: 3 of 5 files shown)


huangzesen:
## Read-only triage: Windows contract failure NOT PR-owned

Run 31314944998 / job 93248573518 (Native Windows / kernel lifecycle, 13:06Z) failed in `Run Windows kernel contract tiers` on the **unchanged** test `tests/test_daemon_windows_supervisor.py::test_windows_adapter_spawns_real_detached_supervisor_reaching_capsule_read` — `AssertionError: condition not met within 60.0s`, `1 failed, 252 passed, 16 skipped`.

Head `13798ff` is CURRENT (1-commit clean rebase onto current main `c1f9789f`, ahead=1/behind=0). Whole PR diff (5 files: lifecycle.py, nudge/ANATOMY.md, nudge/kernel_version.py, test_heartbeat.py, test_kernel_version_nudge.py) touches only the heartbeat/nudge path — **no Windows supervisor/daemon code, and the failing test is not in the diff**.

This is the recurring Windows async-supervisor contract flake seen on multiple sibling PRs (same test, same 60s capsule-read condition, same window). The same PR branch at earlier head `680ab1b9` PASSED this workflow (run 31290283487, 02:24Z). Identity: author `Huang Zesen <hzsbazinga@outlook.com>` canonical; committer name lowercase `huangzesen` with canonical email (separate identity note).

Verdict: HOLD / NOT PR-OWNED. No rerun or mutation performed.

huangzesen:

PR #1293, opened 2026-08-09
fix(llm): notify agent when overflow recovery terminates after trimming entries

### LINKED ISSUE #653: Overflow recovery permanently deletes history entries on terminal failure — no notice, no rollback
## Problem

When context overflow recovery fails (all retry rounds exhausted or no more entries to trim), conversation history entries that were deleted during earlier successful trim rounds are **permanently lost** — no notice is injected and no rollback occurs.

## Root Cause

`_run_with_overflow_recovery` (`base.py:459-503`) trims entries from `_interface._entries` on each retry round via `_trim_context_one_round` (`del entries[first_conv:cut]`). On the terminal-failure path, the overflow error is re-raised without:

1. **Injecting a notice** — the agent does not learn that entries were trimmed
2. **Rolling back** — trimmed entries are not restored

The notice injection (`_inject_overflow_notice`) is only called on the **success path** (`adapter.py:1409`, `1593`), never on the failure path.

## Impact

- **Permanent history loss**: Up to ~65% of non-system entries can be deleted across 10 retry rounds (each round trims ~10%). The shortened history is persisted on the next `_save_chat_history` call.
- **Silent about the trim**: The agent sees a provider error but does not know that entries were also trimmed. It continues with a shortened context, potentially missing critical information.
- **Trigger**: Context overflow that cannot be resolved by trimming (e.g., system prompt itself exceeds limit, or trimming does not free enough space).

## Naive Fix Does Not Work

Injecting a user-role notice before re-raise is a **no-op**: every adapter's `except` block calls `drop_trailing(lambda e: e.role == "user")` which strips ALL trailing user-role entries — including the just-injected notice.

## Recommended Fix

**Option A (recommended, ~15-30 lines)**: Attach trim statistics to the exception object. In each adapter's `except` block, after `drop_trailing` revert, check for attached stats and inject the notice there.

**Option B**: Snapshot `_entries` before the first trim round and restore on failure (prevents loss entirely, but needs careful revert semantics).

## Missing Test Coverage

`test_recovery_gives_up_after_max_rounds` only asserts that the overflow error is re-raised. It does not check:
- Whether entries were trimmed
- Whether a notice was injected
- Whether the history length matches expectations

## Severity

**LOW-MEDIUM**: Real permanent data loss, but behind a rare trigger (untrimmable overflow). Oldest-first loss (often already molted/archived). Agent does see a provider error (not fully silent about the failure itself).

huangzesen:
Fixes #653

## Problem

When context-overflow recovery terminates (all `_OVERFLOW_MAX_ROUNDS` retry rounds exhausted, or `_trim_context_one_round` can no longer drop anything), entries deleted during earlier successful trim rounds stay deleted forever — and the agent never learns that history was trimmed. The `[kernel]` notice (`_inject_overflow_notice`) was only injected on the **success** path; the failure path just re-raised the provider error over a silently shortened history.

## Fix (Option A from the issue)

- `base.py` (`ChatSession._run_with_overflow_recovery`): on terminal failure, best-effort attach `_overflow_trim_stats = (total_dropped, rounds)` to the re-raised provider exception (only when `total_dropped > 0`, nothing to report otherwise).
- New `ChatSession._inject_overflow_notice_after_error(exc)`: reads the attached stats and injects the same `[kernel]` molt-recommendation notice the success path uses. No-op when the exception carries no stats.
- `openai` and `anthropic` adapters (`send` + `send_stream`): call `_inject_overflow_notice_after_error(exc)` in their `except` block **after** the `drop_trailing(lambda e: e.role == "user")` revert — the revert would strip a user-role notice injected before it (the naive fix the issue correctly rejects).

Not touched: `claude_code` / `kimi_code` already roll the canonical history back on failure via `_snapshot_interface()` + `restore()` (their `except` blocks restore the pre-trim entries), so no loss occurs there and no notice is needed.

Behavior after this PR on terminal overflow failure (e.g. system prompt itself exceeds the limit):

1. The provider error is still re-raised (unchanged).
2. The pending user message is still reverted (unchanged).
3. The agent now receives a `[kernel]` notice stating how many oldest entries were dropped and strongly recommending a molt — no more silent history loss.

## Tests

All in `tests/test_openai_overflow_recovery.py` (23 passed, was 18):

- `test_recovery_gives_up_after_max_rounds` — strengthened: asserts history is shorter after terminal failure and the exception carries `_overflow_trim_stats` with `rounds == _OVERFLOW_MAX_ROUNDS` and `total_dropped >= rounds`.
- `test_recovery_attaches_trim_stats_when_trim_stalls` — the "cannot trim further" terminal path also records earlier trims.
- `test_recovery_attaches_no_stats_when_nothing_trimmed` — system-only history: no stats, nothing to report.
- `test_send_terminal_overflow_injects_notice_and_persists_trim` — end-to-end `send()`: error re-raised, pending user message reverted, trims persist, exactly one `[kernel]` notice (with "dropped" and "molt").
- `test_send_stream_terminal_overflow_injects_notice_and_persists_trim` — same guarantee on the streaming path.
- `test_send_terminal_overflow_untrimmable_no_notice` — system-only history: no loss, no notice.

Full suite: `pytest tests/` → 7994 passed, 31 skipped. The 41 failed / 14 error cases are pre-existing environment failures (wheel builds, Windows shims, `claude_code` CLI transport, docs governance) — identical failure set on the pristine baseline (`git stash` comparison: same 5 failures in `test_openai_responses_streaming` / `llm_conversation_input` contracts before and after).

Also manually verified both Anthropic `send` and `send_stream` failure paths: error re-raised with `_overflow_trim_stats=(50, 10)`, exactly one `[kernel]` notice injected, pending message reverted.

## Notes / remaining scope

- The trimmed entries themselves are still not restored on the failure path (that is Option B, snapshot+restore, which the issue flags as needing careful revert semantics; `claude_code`/`kimi_code` already do it). The notice makes the loss non-silent, which is the issue's recommended Option A.

### DIFF
--- a/src/lingtai/kernel/llm/base.py
+++ b/src/lingtai/kernel/llm/base.py
@@ -517,6 +517,40 @@ class ChatSession(ABC):
         )
         self._interface._append("user", [TextBlock(text=notice)])
 
+    def _inject_overflow_notice_after_error(self, exc: Exception) -> None:
+        """Surface the trim on the terminal-failure path, mirroring the notice.
+
+        ``_run_with_overflow_recovery`` attaches ``_overflow_trim_stats`` to
+        the re-raised provider exception when retry rounds were exhausted (or
+        trimming stalled) after entries had already been deleted. Adapters
+        call this from their ``except`` block **after** the ``drop_trailing``
+        revert — the revert would otherwise strip the just-injected
+        user-role notice. No-op when the exception carries no stats, i.e.
+        nothing was trimmed to report.
+        """
+        stats = getattr(exc, "_overflow_trim_stats", None)
+        if not stats:
+            return
+        total_dropped, rounds = stats
+        if total_dropped > 0:
+            self._inject_overflow_notice(total_dropped=total_dropped, rounds=rounds)
+
+    @staticmethod
+    def _attach_overflow_trim_stats(
+        exc: Exception, total_dropped: int, rounds: int,
+    ) -> None:
+        """Best-effort: record trim stats on a re-raised overflow error.
+
+        Provider exception objects normally accept attribute assignment;
+        tolerate anything that does not so recovery behavior is unchanged.
+        """
+        if total_dropped <= 0:
+            return
+        try:
+            exc._overflow_trim_stats = (total_dropped, rounds)
+        except Exception:
+            pass
+
     def _run_with_overflow_recovery(self, do_call):
         """Run an API call with context-overflow auto-recovery.
 
@@ -545,6 +579,10 @@ class ChatSession(ABC):
                         "(dropped %d entries total) — re-raising provider error.",
                         rounds, total_dropped,
                     )
+                    # Entries trimmed in earlier rounds are already deleted;
+                    # record the loss on the exception so adapters can surface
+                    # it after their error-revert step (issue #653).
+                    self._attach_overflow_trim_stats(exc, total_dropped, rounds)
                     raise
                 dropped = self._trim_context_one_round()
                 if dropped == 0:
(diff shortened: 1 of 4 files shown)


huangzesen:

PR #1273, opened 2026-08-09
fix(telegram): surface taskcard loaded-runtime drift hint before refresh

huangzesen:
Fixes #987.

## Problem

PR #984's non-deletable Task Card self-heal only runs after the live Telegram manager has loaded the repair code. When the on-disk checkout is updated underneath a long-lived manager process, the installed source contains the repair while the loaded runtime predates it: the old card stays frozen, `/taskcard` still reports a healthy `enabled` setting, and no diagnostic says that one refresh is required.

This PR does **not** touch the bounded one-replacement semantics from #984 and does not auto-toggle Task Card or delete/re-send anything. It adds the bootstrap/observability half: a deterministic loaded-runtime drift diagnostic on the `/taskcard` surfaces.

## Change

- New `src/lingtai/mcp_servers/telegram/_runtime.py`: captures a non-secret SHA-256 prefix (12 hex) of the curated Telegram Task Card source files (`manager.py` + `task_card/*.py`) **at import time** (the loaded identity), and re-reads the same files from disk at query time (the installed identity). The kernel's issue-#178 fingerprint covers only `lingtai.kernel` sources, so it cannot see the exact #987 drift; this is a telegram-scoped identity.
- `account.py` `_cmd_taskcard` and `_cmd_taskcard_menu`: when loaded != installed, append one bounded, actionable line: `⚠️ Task Card runtime drift: this process predates the installed checkout — one /refresh is required to apply the installed Task Card self-heal.` Absence of the hint means the loaded Task Card runtime matches the installed checkout (the healthy state). Unprovable digests stay silent rather than misreport.
- `tests/test_telegram_task_card_runtime_drift.py`: 7 focused tests (helper determinism, on-disk change detection, hint only when drifting, both `/taskcard` surfaces).

## Tests

- `PYTHONPATH=src python3 -m pytest tests/test_telegram_task_card_runtime_drift.py -q` → **7 passed**
- `python3 -m pytest tests/test_telegram_task_card_*.py -q` → **234 passed, 1 failed**; the single failure (`test_public_telegram_action_reaches_manager`) is **pre-existing on main** (verified by stashing this change and rerunning: it fails identically — the test's `FakeService` lacks `list_accounts`, unrelated to this PR).
- `python3 -m py_compile src/lingtai/mcp_servers/telegram/_runtime.py src/lingtai/mcp_servers/telegram/account.py tests/test_telegram_task_card_runtime_drift.py` → OK
- `git diff --check` → OK

## Scope

[...]

PR #1296, opened 2026-08-09
fix(telegram,agent): accept null optional fields; drop _reasoning for flat third-party tools

### LINKED ISSUE #1237: bug: Telegram send/edit reject the model's own null values; _reasoning leaks into third-party MCP tools
## Summary

Two independent, 100%-reproducible bugs found during continuous production use of a 3-agent LingTai deployment (macOS, scheduled jobs). Both broke *every* call once triggered — not intermittent. Patch for both at the bottom.

### Bug A — Telegram `send`/`edit` reject explicitly-null optional fields

In `_telegram_input_schemas()` (`src/lingtai/mcp_servers/telegram/_family.py`), every optional field is wrapped in the `_nullable()` helper **except** `media` and `reply_markup`, which are bare object schemas. Once the model started explicitly passing `"media": null, "reply_markup": null` — normal behavior once a schema exposes many optional properties — validation rejected every send with a generic `invalid telegram input`.

This degraded **silently**: the kernel falls back to email on Telegram failure, so messages still went out over the wrong channel, which delayed detection by hours.

### Bug B — `_reasoning` forwarded into third-party MCP tools that never declared it

The executor renames `reasoning` → `_reasoning` on every tool call for audit logging, and `Agent._dispatch_tool()` renames it back — but only for tools matching the strict LTP-v2 envelope shape. Third-party MCP servers with flat, auto-generated schemas match neither branch: they never get the rename back, and never declared a `_reasoning` parameter to receive it. Every call then fails with an unexpected-keyword-argument error.

**This one is systemic** — it affects *any* third-party MCP integration not using the LTP-v2 envelope, not just the server I happened to hit it with (a FastMCP/pydantic-based Zotero MCP server).

## Reproduction steps

**Bug A:**
1. Configure an agent with the telegram MCP family.
2. Have it call `telegram(action="send", ...)` with optional fields explicitly null — i.e. `{"chat_id": "...", "text": "hi", "media": null, "reply_markup": null}`.
3. Schema validation rejects the call with `invalid telegram input`, even though all required fields are valid.

**Bug B:**
1. Register any third-party MCP server whose tools use flat, auto-generated parameter schemas rather than the LTP-v2 envelope (e.g. a FastMCP server exposing `add_item(source, source_type, tags, ...)`).
2. Have an agent call one of its tools.
3. The call fails every time — `_reasoning` is forwarded into a strict pydantic-validated function that has no such parameter.

## Expected behavior

- **A:** `media: null` / `reply_markup: null` should be accepted, same as every other optional field in the schema.
- **B:** `_reasoning` is an internal audit key. It should never reach a tool that didn't declare it.

## Actual behavior / logs

- **A:** every send fails schema validation with `invalid telegram input`; delivery silently falls back to email.
- **B:** every third-party MCP tool call fails with an unexpected keyword argument `_reasoning` error.

## LingTai / Python / OS versions

lingtai-kernel 0.19.5 (base commit `3643441c`), Python 3.11, macOS 15 (Darwin 25.5.0)

## Proposed fix

**A** — wrap both fields like every sibling optional field, in both the `send` and `edit` action schemas:

```python
"media": _nullable(media),
"reply_markup": _nullable({"type": "object"}),
```

**B** — add a second schema-shape check and drop the key when there's nowhere to put it:

```python
def _schema_declares_reasoning_param(schema: Any) -> bool:
    """True if a tool's own parameter schema has somewhere for reasoning to go."""
    if not isinstance(schema, dict):
        return False
    properties = schema.get("properties")
    if not isinstance(properties, dict):
        return False
    return "reasoning" in properties or "_reasoning" in properties
```

```python
if len(schemas) == 1 and "_reasoning" in tc.args:
    params = schemas[0].parameters
    if _is_strict_ltp_v2_family_schema(params):
        ...  # existing behavior: rename _reasoning -> reasoning
    elif not _schema_declares_reasoning_param(params):
        args = dict(tc.args)
        args.pop("_reasoning", None)   # nowhere to put it: drop, don't forward
        tc = ToolCall(name=tc.name, args=args, id=tc.id)
```

If a third-party tool *does* declare its own `reasoning`/`_reasoning`, it's left untouched — only tools with genuinely nowhere to put it get it stripped.

<details>
<summary>Full diff (applies against <code>3643441c</code>)</summary>

```diff
diff --git a/src/lingtai/agent.py b/src/lingtai/agent.py
index 3031a3d9..15be0fc8 100644
--- a/src/lingtai/agent.py
+++ b/src/lingtai/agent.py
@@ -145,6 +145,22 @@ def _is_strict_ltp_v2_family_schema(schema: Any) -> bool:
     )
 
 
+def _schema_declares_reasoning_param(schema: Any) -> bool:
+    """True if a tool's own parameter schema has somewhere for reasoning to go.
+
+    Covers both the LTP-v2 envelope's ``reasoning`` and any third-party tool
+    that happens to declare its own ``reasoning``/``_reasoning`` parameter
+    directly. Used by ``_dispatch_tool`` to decide whether the ``_reasoning``
+    audit key is safe to forward as-is or must be dropped before dispatch.
+    """
+    if not isinstance(schema, dict):
+        return False
+    properties = schema.get("properties")
+    if not isinstance(properties, dict):
+        return False
+    return "reasoning" in properties or "_reasoning" in properties
+
+
 class Agent(BaseAgent):
     """BaseAgent with composable capabilities.
 
@@ -1087,18 +1103,32 @@ class Agent(BaseAgent):
         MCP ToolFamily instead validates the original closed LTP-v2 envelope, so
         the wrapper composition root restores the public key immediately before
         the inherited handler dispatch.  The input ToolCall is never mutated.
+
+        Third-party MCP tools (flat, auto-generated parameter schemas — e.g. a
+        vendored server exposing ``add_item(source, tags, ...)`` directly, no
+        LTP-v2 envelope at all) have neither shape: they don't match the strict
+        envelope check above, so ``_reasoning`` is never renamed back, and they
+        also never declared a ``_reasoning``/``reasoning`` parameter of their
+        own to receive it. Forwarding the audit key straight into a strict,
+        schema-validated third-party call (e.g. FastMCP/pydantic) then fails
+        every single time with an "unexpected keyword argument" error — not a
+        model mistake, since ``_reasoning`` is attached to *every* tool call by
+        convention regardless of the target. For that case, drop the key
+        instead of forwarding it: it was never anything the tool could act on.
         """
         if tc.name in getattr(self, "_mcp_tool_names", set()):
             schemas = [schema for schema in self._tool_schemas if schema.name == tc.name]
-            if (
-                len(schemas) == 1
-                and _is_strict_ltp_v2_family_schema(schemas[0].parameters)
-                and "_reasoning" in tc.args
-            ):
-                args = dict(tc.args)
-                reasoning = args.pop("_reasoning")
-                args.setdefault("reasoning", reasoning)
-                tc = ToolCall(name=tc.name, args=args, id=tc.id)
+            if len(schemas) == 1 and "_reasoning" in tc.args:
+                params = schemas[0].parameters
+                if _is_strict_ltp_v2_family_schema(params):
+                    args = dict(tc.args)
+                    reasoning = args.pop("_reasoning")
+                    args.setdefault("reasoning", reasoning)
+                    tc = ToolCall(name=tc.name, args=args, id=tc.id)
+                elif not _schema_declares_reasoning_param(params):
+                    args = dict(tc.args)
+                    args.pop("_reasoning", None)
+                    tc = ToolCall(name=tc.name, args=args, id=tc.id)
         return super()._dispatch_tool(tc)
 
     def _expand_agent_placeholders(self, value):
diff --git a/src/lingtai/mcp_servers/telegram/_family.py b/src/lingtai/mcp_servers/telegram/_family.py
index c563e722..08a75521 100644
--- a/src/lingtai/mcp_servers/telegram/_family.py
+++ b/src/lingtai/mcp_servers/telegram/_family.py
@@ -76,8 +76,8 @@ def _telegram_input_schemas() -> dict[str, dict[str, Any]]:
             "account": _nullable({"type": "string"}),
             "chat_id": _chat_id(),
             "text": {"type": "string"},
-            "media": media,
-            "reply_markup": {"type": "object"},
+            "media": _nullable(media),
+            "reply_markup": _nullable({"type": "object"}),
             "placeholder": _nullable({"type": "boolean"}),
             "chat_action": _nullable({
                 "type": "string",
@@ -163,7 +163,7 @@ def _telegram_input_schemas() -> dict[str, dict[str, Any]]:
             {
                 "message_id": {"type": "string"},
                 "text": {"type": "string"},
-                "reply_markup": {"type": "object"},
+                "reply_markup": _nullable({"type": "object"}),
                 "rendering_mode": {
                     "type": "string", "enum": list(_RENDERING_MODES),
                     "description": (
```

</details>

## Verification

- **A:** reproduced the exact failing payload against the validation path; confirmed fail-before / pass-after. Regression-checked that the fix doesn't loosen validation generally — a call missing a required field, and a call passing the wrong *type* for `media`, are both still correctly rejected.
- **B:** confirmed via the kernel event log that previously-failing calls succeeded immediately after deploy, with no change to the calling code.
- Both have since run in production without recurrence.

## Suggested follow-ups

1. A lint rule or schema-shape test asserting that *every* optional property in a family input schema accepts null would have caught A before it shipped.
2. `_dispatch_tool` has no test coverage for non-LTP-v2 tool shapes. A unit test with a representative flat/third-party schema would catch B's whole class — worth doing given any third-party MCP integration is affected.

## Submission hygiene

- [x] I removed secrets, tokens, private paths that are not needed, and unrelated logs.

huangzesen:
Fixes #1237

Two independent, 100%-reproducible bugs, both reproduced on current main
(`df2b523d`) before this change and verified fixed after.

## Bug A — Telegram `send`/`edit` reject the model's own `null` values

Every optional field in `_telegram_input_schemas()` is wrapped in the
`_nullable()` helper **except** `media` and `reply_markup`, which were bare
object schemas. Once the model starts explicitly passing `"media": null` /
`"reply_markup": null` (normal once a schema exposes many optional properties),
validation rejected every `send`/`edit` with `invalid telegram input`, and the
kernel silently fell back to email delivery.

Fix: wrap both fields like every sibling optional field, in both the `send` and
`edit` action schemas (`media` keeps its description on the inner object branch,
matching the existing `entities` pattern):

```python
"media": _nullable(media),
"reply_markup": _nullable({"type": "object"}),
```

## Bug B — `_reasoning` leaks into third-party MCP tools that never declared it

`ToolExecutor` renames model-facing `reasoning` → `_reasoning` on every tool
call for audit logging, and `Agent._dispatch_tool()` renames it back — but only
for tools matching the strict LTP-v2 envelope shape. Third-party MCP servers
with flat, auto-generated schemas (e.g. FastMCP/pydantic) match neither branch:
the key is never renamed back and they never declared a `_reasoning` parameter,
so every call failed with an unexpected-keyword-argument error. This affects
*any* third-party MCP integration not using the LTP-v2 envelope.

Fix: add a second schema-shape check; when the target schema declares neither
`reasoning` nor `_reasoning`, drop the audit key instead of forwarding it:

```python
def _schema_declares_reasoning_param(schema):
    if not isinstance(schema, dict):
        return False
    properties = schema.get("properties")
    if not isinstance(properties, dict):
        return False
    return "reasoning" in properties or "_reasoning" in properties
```

```python
if len(schemas) == 1 and "_reasoning" in tc.args:
    params = schemas[0].parameters
    if _is_strict_ltp_v2_family_schema(params):
        ...  # existing behavior: rename _reasoning → reasoning
    elif not _schema_declares_reasoning_param(params):
        args = dict(tc.args)
        args.pop("_reasoning", None)   # nowhere to put it: drop, don't forward
        tc = ToolCall(name=tc.name, args=args, id=tc.id)
```

If a third-party tool *does* declare its own `reasoning`/`_reasoning`, it is
left untouched — only tools with genuinely nowhere to put it get it stripped.

## Tests

- `tests/test_telegram_toolfamily_ltpv2.py` + `tests/test_agent.py`: 76 passed,
  including 7 new regression tests:
  - `send` with `media: null` / `reply_markup: null` reaches the manager;
  - `edit` with `reply_markup: null` is accepted;
  - nullable `media`/`reply_markup` do not loosen required/type validation
    (wrong media type and missing required fields still rejected);
  - flat third-party tool gets `_reasoning` stripped before the handler;
  - strict LTP-v2 envelope tools still get `_reasoning` → `reasoning` restored;
  - a non-envelope tool that declares its own `reasoning` keeps the key;
  - `_schema_declares_reasoning_param` unit checks.
- `tests/test_telegram_send_media_contract.py`: updated for the nullable
  `anyOf` wrapper; the advertised media types remain exactly
  `["photo", "document"]` and the `null` branch is asserted.
- Broader telegram + tool-family run: 930 passed; the only 4 failures
  (`test_telegram_task_card_transport.py::test_public_telegram_action_reaches_manager`,
  3× `test_tool_family_vision_migration.py`) are pre-existing and fail
  identically on the unmodified baseline (`git stash` check).

Reproduction evidence (before → after):

```
# Bug A (main):
send  -> {'status': 'failed', 'error_code': 'INVALID_ARGUMENT', 'message': 'invalid telegram input'}
edit  -> {'status': 'failed', 'error_code': 'INVALID_ARGUMENT', 'message': 'invalid telegram input'}
# Bug A (fixed): validation passes; call reaches the manager.

# Bug B (main): handler args = {'source': 'x', '_reasoning': 'audit key'}  (leak)
# Bug B (fixed): handler args = {'source': 'x'}                            (clean)
```

### DIFF
--- a/src/lingtai/agent.py
+++ b/src/lingtai/agent.py
@@ -145,6 +145,22 @@ def _is_strict_ltp_v2_family_schema(schema: Any) -> bool:
     )
 
 
+def _schema_declares_reasoning_param(schema: Any) -> bool:
+    """True if a tool's own parameter schema has somewhere for reasoning to go.
+
+    Covers both the LTP-v2 envelope's ``reasoning`` and any third-party tool
+    that happens to declare its own ``reasoning``/``_reasoning`` parameter
+    directly. Used by ``_dispatch_tool`` to decide whether the ``_reasoning``
+    audit key is safe to forward as-is or must be dropped before dispatch.
+    """
+    if not isinstance(schema, dict):
+        return False
+    properties = schema.get("properties")
+    if not isinstance(properties, dict):
+        return False
+    return "reasoning" in properties or "_reasoning" in properties
+
+
 class Agent(BaseAgent):
     """BaseAgent with composable capabilities.
 
@@ -1174,19 +1190,28 @@ class Agent(BaseAgent):
         in-process and legacy MCP handlers consume that existing shape.  A native
         MCP ToolFamily instead validates the original closed LTP-v2 envelope, so
         the wrapper composition root restores the public key immediately before
-        the inherited handler dispatch.  The input ToolCall is never mutated.
+        the inherited handler dispatch.  Third-party MCP tools with flat,
+        auto-generated parameter schemas declare no ``reasoning``/``_reasoning``
+        at all; forwarding the audit key would fail their strict schema-validated
+        calls, so it is dropped instead.  The input ToolCall is never mutated.
         """
         if tc.name in getattr(self, "_mcp_tool_names", set()):
             schemas = [schema for schema in self._tool_schemas if schema.name == tc.name]
-            if (
-                len(schemas) == 1
-                and _is_strict_ltp_v2_family_schema(schemas[0].parameters)
-                and "_reasoning" in tc.args
-            ):
-                args = dict(tc.args)
-                reasoning = args.pop("_reasoning")
-                args.setdefault("reasoning", reasoning)
-                tc = ToolCall(name=tc.name, args=args, id=tc.id)
+            if len(schemas) == 1 and "_reasoning" in tc.args:
+                params = schemas[0].parameters
+                if _is_strict_ltp_v2_family_schema(params):
+                    args = dict(tc.args)
+                    reasoning = args.pop("_reasoning")
+                    args.setdefault("reasoning", reasoning)
+                    tc = ToolCall(name=tc.name, args=args, id=tc.id)
+                elif not _schema_declares_reasoning_param(params):
+                    # Third-party MCP tools expose flat, auto-generated parameter
+                    # schemas with no LTP-v2 envelope and no reasoning parameter
+                    # of their own; the audit key has nowhere to go, so drop it
+                    # instead of failing a strict schema-validated call.
+                    args = dict(tc.args)
+                    args.pop("_reasoning", None)
+                    tc = ToolCall(name=tc.name, args=args, id=tc.id)
         return super()._dispatch_tool(tc)
 
     def _expand_agent_placeholders(self, value):
--- a/src/lingtai/mcp_servers/telegram/_family.py
+++ b/src/lingtai/mcp_servers/telegram/_family.py
@@ -76,8 +76,8 @@ def _telegram_input_schemas() -> dict[str, dict[str, Any]]:
             "account": _nullable({"type": "string"}),
             "chat_id": _chat_id(),
             "text": {"type": "string"},
-            "media": media,
-            "reply_markup": {"type": "object"},
+            "media": _nullable(media),
+            "reply_markup": _nullable({"type": "object"}),
             "placeholder": _nullable({"type": "boolean"}),
             "chat_action": _nullable({
                 "type": "string",
@@ -98,7 +98,7 @@ def _telegram_input_schemas() -> dict[str, dict[str, Any]]:
             {"required": ["chat_action"]},
         ],
     )
-    send["properties"]["media"]["description"] = (
+    send["properties"]["media"]["anyOf"][0]["description"] = (
         "Media attachment: {type: 'photo'|'document', path: '/path/to/file'}. "
         "For charts, HTML/SVG/PNG reports, CSVs, PDFs, and other generated artifacts "
         "that should arrive as an intact file, use type='document'. Use type='photo' "
@@ -163,7 +163,7 @@ def _telegram_input_schemas() -> dict[str, dict[str, Any]]:
             {
                 "message_id": {"type": "string"},
                 "text": {"type": "string"},
-                "reply_markup": {"type": "object"},
+                "reply_markup": _nullable({"type": "object"}),
                 "rendering_mode": {
                     "type": "string", "enum": list(_RENDERING_MODES),
                     "description": (
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -7,7 +7,7 @@ from unittest.mock import MagicMock
 import pytest
 
 from lingtai.kernel.base_agent import BaseAgent
-from lingtai.agent import Agent
+from lingtai.agent import Agent, _schema_declares_reasoning_param
 from lingtai.kernel.message import Message, _make_message, MSG_REQUEST, MSG_USER_INPUT, MSG_TC_WAKE
 from lingtai.kernel.state import AgentState
 from lingtai.kernel.types import UnknownToolError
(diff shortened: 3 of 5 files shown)


huangzesen:
Superseded by #1322 (squash of #1321 + #1298 + #1296 + #1273). Closing as superseded.

huangzesen:
Merged via squash PR #1322 (86ac0732). Thanks!

Telegram: @lingtaidev4bot | Nickname: dev4bot

PR #1297, opened 2026-08-09
feat(taskcard): move row timestamp under API info line

huangzesen:
Tool rows drop their trailing per-row timestamp; each API-call info line now carries one timestamp on the line below it, derived from the group's first event ts. Per maintainer request. 93 taskcard tests pass.


huangzesen:
Superseded by the main timestamp series (#1303/#1304/#1306/#1307/#1308/#1310) which moved the per-API-call timestamp above the API metadata line. This PR proposed an earlier variant of the same change. Closing.

PR #1299, opened 2026-08-09
fix(kernel): 429 rate-limit retries honor Retry-After/backoff; 4xx client errors fail fast

### LINKED ISSUE #593: LLM 429 rate limit retried without backoff or Retry-After; 400 bad request retried identically
## Problem

When the LLM provider returns a 429 (rate limit) or 400 (bad request), the kernel treats both as generic non-transient errors and retries them through the same AED recovery path — with no backoff, no Retry-After header respect, and no differentiation between "wait and retry" (429) vs "do not retry" (400).

### Error classification (`turn.py:123-136`)
```python
def _is_transient_provider_error(exc: Exception) -> bool:
    """...
    4xx errors (including quota/rate limit) are not treated as transient here.
    """
    status_code = _exception_status_code(exc)
    if status_code is not None:
        return 500 <= status_code < 600
```

Only **5xx** is classified as transient. **429 and 400 are both non-transient** and go through the same general AED path.

### Retry behavior
- **Transient (5xx)**: exponential backoff (2^n, max 8s), up to `_TRANSIENT_AED_RETRY_LIMIT=3` (turn.py:77)
- **Non-transient (4xx including 429)**: general AED with `max_aed_attempts=3` (config.py:61), **NO backoff**, **NO Retry-After**
- Both paths inject the same `system.stuck_revive` message (i18n en.json:7, zh.json:7)

### Retry-After header
```
$ grep -rn "Retry-After\|retry.after\|retry_after\|RETRY_AFTER" src/lingtai_kernel/
(empty — zero results)
```

The standard HTTP rate limit protocol (`Retry-After` header) is completely ignored.

## Impact

**429 rate limit**:
1. Retried immediately with no delay — likely to hit rate limit again on next attempt
2. All 3 AED attempts wasted within seconds, then the agent goes STUCK
3. When multiple daemons share the same API key (as I experienced today: 3 daemons dispatched → 2 hit 429 → both failed), the tight retry loop worsens the rate limit problem instead of backing off

**400 bad request**:
1. Retried up to 3 times — but a malformed request will keep failing the same way
2. Only "recovers" if AED compaction happens to change the wire format (accidental, not targeted)
3. Wastes API quota on guaranteed-to-fail retries

## Suggested Fix

Split the non-transient path by error class:

```
429 (rate limit):
  - Read Retry-After header if present
  - If absent, use exponential backoff (start 5s, 2^n, max 60s)
  - Retry up to max_aed_attempts
  - Log rate_limit_retry events

400 (bad request):
  - Do NOT retry blindly
  - Log the error details and fail fast
  - Only retry after compaction or wire changes
```

## Reproduction

1. Dispatch 3+ daemons simultaneously using the same API key
2. Observe 429 errors from the provider
3. Daemons fail all 3 AED attempts within seconds (no backoff)
4. Daemons report failure despite the rate limit being transient if they had waited

I experienced this exact scenario today: dispatched em-5, em-6, em-7 in parallel → em-5 and em-7 both hit 429 → both exhausted 3 retries immediately → both failed.

huangzesen:
Fixes #593.

## Problem

429 (rate limit) and 400 (bad request) provider errors were both treated as
non-transient and retried through the same generic AED path: no backoff, no
`Retry-After` header respect, and no differentiation between "wait and retry"
(429) vs "do not retry" (400). 429s burned all 3 AED attempts within seconds
(aggravating the shared-key rate limit), and 400s were retried identically up
to 3 times on an unchanged wire.

## Change (`src/lingtai/kernel/base_agent/turn.py`)

1. **New rate-limit branch in the AED loop** (`_is_rate_limit_error`):
   - 429/quota/rate-limit errors retry with backoff, honoring the
     `Retry-After` header (delta-seconds or HTTP-date) when present, falling
     back to exponential backoff 5s \* 2^n capped at 60s otherwise, up to a
     dedicated `_RATE_LIMIT_RETRY_LIMIT` (3).
   - Rate-limit retries **never compact history** (quota is the problem, not
     the wire — compaction permanently discards context for zero benefit).
   - New `aed_rate_limit_retry` / `aed_rate_limit_exhausted` events; when the
     budget is exhausted the agent goes ASLEEP instead of burning AED
     attempts.
2. **Client-error fail-fast in the AED branch** (`_is_client_error`):
   - Deterministic 4xx errors (400/401/403/404..., excluding 429) only retry
     when retroactive compaction actually changed the wire
     (`compacted_blocks > 0` — this preserves the existing
     `messages_parameter_illegal` + compaction recovery, see #613).
   - Otherwise the remaining AED budget is not spent on guaranteed-to-fail
     retries: new `aed_client_error_noop` event + ASLEEP.
3. **`Retry-After` extraction** (`_rate_limit_retry_after_seconds`) covers the
   SDK shapes in the adapter zoo: `exc.retry_after*` attributes, response
   headers, JSON body (`resets_in_seconds`, zhipu-style), and the error
   string as a last resort.

## Tests

8 new regression tests in `tests/test_aed_recovery.py`:

- classifier unit tests (`_is_rate_limit_error`, `_is_client_error`)
- Retry-After extraction across header/body/string/HTTP-date shapes
- run-loop: 429 honors Retry-After (sleeps exactly the header value), no
  compaction, no AED budget spent
- run-loop: 429 without header uses exponential backoff [5, 10, 20] then
  ASLEEP on exhaustion
- run-loop: 400 with no wire change fails fast after 1 attempt (vs 3)
- run-loop: 400 with compaction that changed the wire still retries

Validation:

- `tests/test_aed_recovery.py`: 21 passed (13 pre-existing + 8 new)
- turn/AED-related kernel suites (aed_tool_pairing, runtime_block_turn,
  openai_overflow, tool_result_recovery, turn_boundary_housekeeping,
  tc_inbox_mid_turn_drain, retroactive_history_compaction): **112 passed**
- full suite: 7995 passed / 43 failed / 14 errors — all failures verified
  pre-existing on pristine main (wheel/windows env, docs line-number checks,
  claude_code adapter `tools=None`, flaky email/nudge timing); no failure
  touches kernel/AED paths
- `git diff --check` clean

### DIFF
--- a/src/lingtai/kernel/base_agent/turn.py
+++ b/src/lingtai/kernel/base_agent/turn.py
@@ -115,6 +115,41 @@ _TRANSIENT_MSG_FRAGMENTS = (
     "gateway timeout",
 )
 
+# Issue #593: 429 / quota / rate-limit errors are transient along the *time*
+# axis — the same wire succeeds once the limit resets — so they must be
+# retried with backoff (ideally honoring the ``Retry-After`` header) instead of
+# being funneled through the generic AED path, which retries with no delay and
+# compacts history that has nothing to do with the rate limit.
+_RATE_LIMIT_RETRY_LIMIT = 3
+_RATE_LIMIT_BACKOFF_START_S = 5.0
+_RATE_LIMIT_BACKOFF_MAX_S = 60.0
+_RATE_LIMIT_MSG_FRAGMENTS = (
+    "rate limit",
+    "rate_limit",
+    "rate-limit",
+    "too many requests",
+    "quota",
+    "usage limit",
+    "usage_limit",
+    "usage_limit_reached",
+    "resets_in_seconds",
+    "throttl",
+    "429",
+)
+
+# Deterministic 4xx client errors (excluding 429, which is handled above).
+# A 400/401/403/404-style error means the request itself is wrong on our side:
+# retrying the identical wire is guaranteed to fail again, so the AED branch
+# must only retry when retroactive compaction actually changes the wire.
+_CLIENT_ERROR_MSG_FRAGMENTS = (
+    "bad request",
+    "400 bad request",
+    "messages_parameter_illegal",
+    "invalid request",
+    "invalid parameter",
+    "malformed request",
+)
+
 
 def _exception_status_code(exc: Exception) -> int | None:
     """Best-effort HTTP-ish status extraction across SDK exception shapes."""
@@ -166,7 +201,10 @@ def _is_transient_provider_error(exc: Exception) -> bool:
     The adapter zoo wraps HTTP failures through different SDK exception
     classes.  Prefer explicit status-code handling when present; otherwise
     fall back to stable class names and conservative message fragments.
-    4xx errors (including quota/rate limit) are not treated as transient here.
+    4xx errors (including quota/rate limit) are not treated as transient here:
+    429/rate-limit errors get their own backoff branch (``_is_rate_limit_error``)
+    and deterministic 4xx client errors fail fast in the AED branch
+    (``_is_client_error``) instead of burning retries on an unchanged wire.
     """
     if isinstance(exc, EmptyLLMResponseError):
         return True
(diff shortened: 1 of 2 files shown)


huangzesen:

PR #1316, opened 2026-08-09
fix(vision): dedupe tilde/absolute allowed preset refs in list

huangzesen:
## Summary

Follow-up to #1315: `vision(action='list')` enumerated the same preset twice when `manifest.preset.allowed` lists it both as `~/.lingtai-tui/...` and as an absolute path (the mechanical list merged `resolve_allowed_presets`, which already expands `~`, with the raw allowed entries). Every allowed ref is now normalized with `Path.expanduser()` before the union, so each preset is reported once.

## Validation

- New regression test: tilde + absolute refs for the same file → `count == 1`.
- vision + manual-schemas suites: **66 passed / 0 failed**.
- `py_compile` and `git diff --check` clean.

Telegram: @lingtaidev4bot | Nickname: 守真

### DIFF
--- a/src/lingtai/tools/vision/__init__.py
+++ b/src/lingtai/tools/vision/__init__.py
@@ -575,8 +575,8 @@ class VisionManager:
                 init_data = _json.loads(init_path.read_text(encoding="utf-8"))
                 manifest = init_data.get("manifest") or {}
                 allowed = sorted(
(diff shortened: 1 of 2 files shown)


huangzesen:
Jason 决定：重复根源是 allowed 配置混用 ~ 与绝对路径写法，属配置问题，无需代码去重。关闭，不 merge。分支保留作证据。

PR #1298, opened 2026-08-09
fix(telegram): redact Bot API token from HTTP request URL logs

huangzesen:
Fixes Lingtai-AI/lingtai#680.

httpx/httpcore emit INFO records carrying the full request URL, and the Telegram Bot API authenticates via a token embedded in the URL path (/bot<id>:<secret>/...). The bundled Telegram addon therefore wrote unredacted bot tokens to the agent log on every ordinary request (getMe, setMyCommands, getUpdates), and the trajectory redactor did not recognize the URL form either, so tool/event logs persisted a second copy.

Changes:
- src/lingtai/mcp_servers/_entrypoint.py: attach a redacting logging filter to the httpx and httpcore loggers in the shared curated-MCP entrypoint, so token-bearing URLs are redacted before any handler persists the record (benefits every bundled addon, not just telegram).
- src/lingtai/kernel/trace_redaction.py: teach the trajectory redactor the Telegram Bot API URL form (/bot<id>:<secret>/... and /file/bot<id>:<secret>/...) so durable tool/event logs never persist the token.
- Tests with synthetic tokens: agent-log records and trajectory redaction both keep the /bot prefix visible while removing the secret; ordinary URLs and short secrets are untouched.

Validation:
- python -m pytest tests/test_trace_redaction.py tests/test_mcp_server_scaffold.py -> 51 passed
- Live httpx reproduction (httpx 0.28.1): INFO line before fix carried the full token URL; after fix it logs https://api.telegram.org/bot<REDACTED:telegram_bot_token>/getUpdates
- Full suite: no new failures vs clean main (2 flaky timing-sensitive tests fail identically on clean main; claude_code adapter env failure pre-existing)

PR #1321, opened 2026-08-09
fix(aed,telegram): sanitized AED-exhaust notice + typing TTL lease (lingtai#672)

huangzesen:
Closes https://github.com/Lingtai-AI/lingtai/issues/672

## Problem

When AED (API-error detector) exhausts its retry budget, the agent goes ASLEEP without any user-visible reply to the originating IM conversation. On Telegram the user is left staring at a typing indicator that never stops (the indicator only stops after a successful `_send`), in the worst case for 8+ minutes.

## Narrow fix

1. **Kernel `turn.py` — AED-exhaust sanitized notice.** On AED exhaustion, send exactly ONE sanitized, user-safe message to the originating Telegram conversation through the existing `telegram` tool handler (the same seam an LLM tool call uses) before going ASLEEP. The text is fixed and never embeds `err_desc` (which may carry secrets/paths/provider internals). Fail-open: any failure is logged and swallowed; the AED/ASLEEP decision is never affected.
   - **Immutable origin binding (Fable P1-1).** The `(account, chat_id)` route is captured ONCE at turn dequeue time (`_run_loop` → `turn_origin`), before any LLM/tool work can read or replace the notification snapshot. The AED-exhaust notice consumes that captured route — never a live re-read of the coalesced notification store — so a later chat or a dismissed notification cannot lose or misroute the notice. Routes are parsed strictly (`_parse_telegram_route`: exactly `<account>:<chat_id>:<message_id>`, non-empty account, integer ids, synthetic `updates` bucket rejected), so ambiguous/malformed refs are never sent to a fabricated chat.
   - **Observable ASLEEP (Fable P1-2).** The AED-exhaust branch now calls `_set_state(AgentState.ASLEEP, ...)` explicitly (like the worker-hang exit) so supervisors/humans observe ASLEEP — not a terminal STUCK — in addition to setting the `_asleep` event.
   - **Truthful failure audit (Fable P2-1).** The handler result is interpreted: returned `{status: error}` mappings are recorded as `failed` (not `sent`), and raised errors log only the exception class (never raw handler text that may carry secrets/paths).

2. **Telegram `manager.py` — typing indicator lifecycle guarantees.** `TypingIndicatorManager` now guarantees typing can never run forever:
   - **Hard TTL bound (Fable P1-4).** The deadline is checked *before* every Bot API send and recomputed after it, so no action can be issued at/after expiry. Custom TTL is validated finite and positive. An injectable clock makes the bound testable deterministically.
   - **Generation-safe workers (Fable P1-3).** Each lease has object identity; a worker removes its mapping only when `self._active_chats[key] is lease`, so a stale worker cannot delete a newer lease. Workers are tracked and `stop_all` boundedly joins them. `start_typing` after shutdown began is a no-op; repeated starts renew/replace the active lease.
   - **Shutdown ordering (Fable P1-3).** `TelegramManager.stop()` now quiesces producers first (`service.stop()` signals+joins account poll threads) then calls `stop_all()` — no in-flight callback can start a new lease after the clear.

3. **Tests (Fable P2-2/P2-3).** Added deterministic, failing-first lifecycle tests with fake-clock/event synchronization covering: no send after deadline, stale-worker/new-lease identity, repeated-start renewal, stop_all join + shutdown no-op, invalid TTL rejection, immutable dequeue-bound origin (later-chat replacement + notification dismissal), malformed-route rejection, returned-error `failed` status, and observable ASLEEP state. `test_telegram_typing_ttl.py` is linked into the Anatomy graph.

## Validation

- `tests/test_telegram_typing_ttl.py`: 8 passed (incl. fake-clock hard-deadline + identity/join/shutdown).
- `tests/test_aed_recovery.py`: 18 passed (incl. origin-capture, malformed rejection, returned-error status, ASLEEP observable).
- Broader Telegram adjacent matrix: 204 passed; Feishu native-progress file has the pre-existing `lark_channel` env collection error (reproduces on exact base).
- `compileall` clean; `git diff --check` clean.
- Architecture docs: candidate orphan set now matches exact base (5 pre-existing orphans; the new typing TTL test is linked).

## Boundaries

- No new outbound channel: the notice rides the existing telegram tool handler.
- No changes to AED retry/fallback logic, Task Card behavior, or config.
- Avoids kernel PR #1194 (ZigongXu token-expired recovery) and open lingtai#791.

Telegram: @lingtaidev3bot | Nickname: 澄一

### DIFF
--- a/src/lingtai/kernel/base_agent/turn.py
+++ b/src/lingtai/kernel/base_agent/turn.py
@@ -149,6 +149,91 @@ def _report_api_error_to_task_card(
         pass
 
 
+# lingtai#672: user-visible, sanitized notice sent exactly once when AED
+# exhausts. Fixed text — never embeds ``err_desc`` (which may carry
+# secrets, absolute paths, or provider internals).
+_AED_EXHAUSTED_USER_MESSAGE = (
+    "\u26a0\ufe0f The model service is temporarily unavailable and I couldn't "
+    "finish processing your request. I've paused to recover \u2014 please try "
+    "again in a little while."
+)
+
+
+def _aed_origin_route(agent):
+    """Derive ``(account, chat_id)`` of the current turn's originating
+    Telegram chat from the notification store, or ``None`` when unavailable.
+
+    Mirrors ``BaseAgent._setup_telegram_task_card``: the first Telegram
+    preview's ``message_ref`` is ``<account>:<chat_id>:<message_id>``.
+    Fail-open — any anomaly returns ``None`` and never raises.
+    """
+    try:
+        from ..notifications import is_channel_allowed
+        store = getattr(agent, "_notification_store", None)
+        if store is None:
+            return None
+        notifications = store.snapshot(is_channel_allowed)
+        telegram_data = notifications.get("mcp.telegram")
+        if not telegram_data or not isinstance(telegram_data, dict):
+            return None
+        data = telegram_data.get("data", {})
+        previews = data.get("previews", []) if isinstance(data, dict) else []
+        if not previews:
+            return None
+        first = previews[0] if isinstance(previews, list) and previews else {}
+        message_ref = first.get("message_ref", "") if isinstance(first, dict) else ""
+        if not message_ref or not isinstance(message_ref, str):
+            return None
+        parts = message_ref.split(":", 2)
+        if len(parts) < 2:
+            return None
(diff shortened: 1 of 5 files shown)


huangzesen:
Superseded by #1322 (squash of #1321 + #1298 + #1296 + #1273). Closing as superseded.

PR #1336, opened 2026-08-10
feat(behaviors): LABT v1 files for 6 module families

huangzesen:
## Summary
LABT v1 → v2 per fable review (comment 5235186729). All fable blocking issues addressed; validation test lands with the PR.

**Root spec (LABT v2)**
- Family-prefix IDs allowed (D/L/T/C/K/S/F###/W###/FE### etc.), anchor #behavior-<id-lower>; default B###
- contract: may be a list; guards must name the target contract frontmatter name exactly (or path)
- runner = agent shape (NOT pytest command); self-containment normative

**Module BEHAVIORS.md (7 families, ~44 LABTs)**
- tools/daemon 7 (D001-D007), tools/context 8 (L001-L008), tools/tool_family 9 (T001-T009), tools/telegram 2 (C001-C002), tools/web_search (W###), tools/file (F###), tools/feishu (FE###), kernel 7 (K001-K007), tools/bash (S###)

**Fable fixes applied**
- telegram frontmatter bug fixed (7 keys now parse); guards/runner normalized; C003 Pass/Fail added; C003-C006 re-homed to web_search/file/feishu
- shell → tools/bash re-home; old tools/shell/ removed; S004/S005/S002 contract names fixed; venv path removed; S001 POSIX-only
- kernel: K001/K002/K005/K006/K007 rewritten as observable-outcome LABTs (real runtime staging); guards → context-contract; contract list
- tool_family T006/T007 guards names fixed
- Contract → behaviors clause links added (13 Guarded by: lines across 8 CONTRACT.md)

**Validation test (new)**
- tests/test_labt_validation.py: frontmatter/labt_version/related_files, LABT fields + sections, guards→real contract clauses, contract→behaviors link completeness. 5/5 pass

**Verification**: test_labt_validation 5/5 + test_karma 19/19 pass; arch-doc 3 pre-existing Windows failures (baseline unchanged).

### DIFF
--- a/BEHAVIORS.md
+++ b/BEHAVIORS.md
@@ -1,7 +1,7 @@
 ---
 name: component-behavior-test-convention
 behavior_version: 2
-labt_version: 1
+labt_version: 2
 related_files:
   - CONTRACT.md
   - ANATOMY.md
@@ -11,9 +11,12 @@ related_files:
   - src/lingtai/tools/system/BEHAVIORS.md
   - src/lingtai/tools/context/BEHAVIORS.md
   - src/lingtai/tools/daemon/BEHAVIORS.md
-  - src/lingtai/tools/shell/BEHAVIORS.md
+  - src/lingtai/tools/bash/BEHAVIORS.md
   - src/lingtai/tools/telegram/BEHAVIORS.md
   - src/lingtai/tools/tool_family/BEHAVIORS.md
+  - src/lingtai/tools/file/BEHAVIORS.md
+  - src/lingtai/tools/feishu/BEHAVIORS.md
+  - src/lingtai/tools/web_search/BEHAVIORS.md
 maintenance: |
   This file is the normative root of the distributed behavior-test definition
   system: it owns the LingTai Agent Behavior Task (LABT) specification, its
@@ -50,7 +53,7 @@ It is the third leg of the **tridirectional linkage**:
 | `BEHAVIORS.md` | How do we verify the promise with an agent? | Self-contained agent tasks (LABT) |
 | `ANATOMY.md` | Where is the code? | Code navigation |
 
-## LingTai Agent Behavior Task (LABT) — version 1
+## LingTai Agent Behavior Task (LABT) — version 2
 
 A **LABT** is the unit of behavior verification in this repository: a single
 markdown section, fully self-contained, that an agent or daemon can execute
@@ -69,11 +72,11 @@ pytest are for traceability and review — never required for execution.
 
 | Field | Required | Meaning |
 |---|---|---|
-| `id` | yes | Stable anchor, `B###`; section heading becomes `#behavior-b###` |
+| `id` | yes | Stable anchor. Default scheme is `B###`; section heading becomes `#behavior-b###`. Per-family files MAY use a family prefix (e.g. `D###` daemon, `L###` lifecycle, `T###` tool-family, `C###` comms, `K###` kernel, `S###` shell) — prefix IDs are globally unique, collision-free across files, and readably self-describing in cross-refs; anchor becomes `#behavior-<prefix>###` (e.g. `#behavior-d003`). A validation test MUST NOT hardcode one scheme; it must accept the per-family scheme a child file declares. |
 | `title` | yes | One-sentence behavior promise |
-| `guards` | yes | Contract clause guarded (bidirectional ref) |
+| `guards` | yes | Contract clause guarded (bidirectional ref). Format: `<contract-frontmatter-name> § <clause heading>`; the `name:` MUST match the target contract's frontmatter exactly |
 | `supersedes` | no | pytest file(s) this LABT replaces/complements (traceability) |
-| `runner` | yes | Which agent/daemon shape can execute (e.g. any LingTai agent with `system`) |
+| `runner` | yes | Which agent/daemon shape can execute (e.g. any LingTai agent with `system`) — an agent shape, NOT a pytest command |
 | `prerequisites` | yes | Exact setup required (agents, dirs, files, permissions) |
 | `steps` | yes | Ordered concrete actions the executor performs |
 | `expected_evidence` | yes | Observable outcomes checklist (state/file/receipt/notification) |
@@ -147,8 +150,8 @@ in the same change. This is a review gate, not optional polish.
 ---
 name: <component>-behavior-tests      # e.g. system-behavior-tests
 behavior_version: 1                   # child-file format revision
-labt_version: 1                       # MUST match this root's labt_version
-contract: CONTRACT.md                 # the contract this guards (relative)
+labt_version: 2                       # MUST match this root's labt_version
+contract: CONTRACT.md                 # the contract(s) this guards (relative); may be a list
 anatomy: ANATOMY.md                   # the anatomy this pairs with (relative)
 related_files:                        # real repo-relative paths
   - src/lingtai/tools/system/karma.py
--- a/BEHAVIORS.md
+++ b/BEHAVIORS.md
@@ -7,6 +7,13 @@ related_files:
   - ANATOMY.md
   - dev-guide-skill/SKILL.md
   - tests/CONTRACT.md
+  - src/lingtai/kernel/BEHAVIORS.md
+  - src/lingtai/tools/system/BEHAVIORS.md
+  - src/lingtai/tools/context/BEHAVIORS.md
+  - src/lingtai/tools/daemon/BEHAVIORS.md
+  - src/lingtai/tools/shell/BEHAVIORS.md
+  - src/lingtai/tools/telegram/BEHAVIORS.md
+  - src/lingtai/tools/tool_family/BEHAVIORS.md
 maintenance: |
   This file is the normative root of the distributed behavior-test definition
   system: it owns the LingTai Agent Behavior Task (LABT) specification, its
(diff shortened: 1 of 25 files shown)


huangzesen:
## Review — LABT v1 files for 6 module families

**Conclusion: changes needed.** `daemon/`, `context/`, and `tool_family/` are excellent and approve-ready as-is — they are the best worked examples of LABT v1 in the repo. `kernel/` is landable with a trim. `telegram/` has a hard frontmatter bug plus a systematic misuse of `guards`/`runner` and should not land in its current form. Separately, the **contract → behaviors edge of the tridirectional loop is missing for all six families**, which the root spec calls "a review gate, not optional polish."

---

### Strengths

- **The good files are genuinely agent-executable, not pytest in disguise.** `D003` stages two synthetic run dirs sharing a legacy handle, pins the exact `daemon.json` payload, asserts `ambiguous: true / match_count: 2 / latest_run_id`, *and* cleans up in step 8 with the leftover dirs as an explicit fail condition. `L005` includes both a positive control (`no_limit.json` must **not** be refused) and a revert control. `T009` crafts the inbound `_return_route` message by hand and requires the ambiguity guard to refuse *loudly* rather than self-deliver. These are real behavioral tests.
- **Self-containment is honored where it counts.** Exact JSON bodies, exact error strings (`"old_string found 2 times — use replace_all=true or provide more context"`), exact anchors, exact caps (`<= 230` lines, `1000`, `3600.0`). An executor genuinely does not need to open another file.
- **`Pass / Fail` sections do real work.** Most name forbidden side effects rather than restating the evidence — "the molt must not delete `.notification/`", "an ambiguous reply must create no sent record", "never become a maintained flag catalog."
- **Traceability is accurate.** Every `supersedes` path I sampled resolves to a real file. The `src/lingtai/CONTRACT.md` rule 6 text quoted in `K001`/`K002`/`K007`/`L008` matches the contract **verbatim**, including the goal-reminder carve-out.
- **Anchors mostly resolve.** I checked every `guards` link against actual headings: `file#per-action-behavior`, `avatar#tool-surface`, `mcp#tool-surface`, `daemon#tool-surface`, `daemon#backend-support-matrix`, `bash#cross-platform-invariants`, `system#state--storage`, `context#molt-safety-invariants`, `notification#behavior`, `web_search#provider-ownership-and-routing` all exist.
- **Honest about limits** — `S001`'s Windows caveat and `L002`'s "missing `suspend_received` does NOT fail the LABT" are the right instinct.

---

### Blocking

**1. `telegram/BEHAVIORS.md` frontmatter is broken.** The block is closed by a `---` on **line 5**, so `related_files` and `maintenance` fall into the body and the `---` on line 30 is a stray. Verified:

```
yaml.safe_load(...) -> keys: ['behavior_version', 'labt_version', 'name']
contract: None   anatomy: None   n_related: 0
```

All five other files parse 7 keys. Required `contract:`/`anatomy:` are absent. Delete the line-5 `---`.

**2. `telegram` C003 has no `Pass / Fail` section** (6 LABTs, 5 `### Pass / Fail`). Required field.

**3. `telegram` `guards` does not reference contract clauses — in all 6 LABTs.** It carries test-isolation notes: *"test only runs with the telegram task-card fake transport; no network"*, *"provider factories are stubbed"*, *"operates on fixture files under tests/fixtures"*. Per spec `guards` is the behaviors→contract backlink. The real contract refs were put in `prerequisites` instead. As filed, this file has **zero** contract linkage.

**4. `telegram` `runner` is a pytest command, not an agent shape** — `Runner: pytest tests/test_telegram_task_card_last_message.py`. Spec: "Which agent/daemon shape can execute." This makes the file a pytest wrapper rather than a LABT; the self-containment rule is moot when the instruction is "run this pytest file." Compare `D001` or `T009`, which an agent can actually perform.

**5. `telegram` uses different field spellings** — `**LABT id**`, `**Title**`, `**Guards**`, `**Runner**` (capitalized) vs the template's lowercase `**id**`/`**title**`/`**guards**` used by the other five. Any validation test will need one spelling.

**6. The contract → behaviors edge is missing for all six families.** The spec requires `[B###](BEHAVIORS.md#behavior-b###)` on every important behavior clause. `system/CONTRACT.md` (the pre-existing reference impl) has 5 such links. Across the whole repo that is still the only file with any — `grep "BEHAVIORS.md#behavior" --include=CONTRACT.md` returns 5 hits, all in `system/`. This PR only added `BEHAVIORS.md` to `related_files`, which is the frontmatter edge, not the clause-level edge. `bash/`, `file/`, `avatar/`, `mcp/`, `psyche/`, `email/`, `notification/`, `web_search/CONTRACT.md` were not touched at all, and `kernel/CONTRACT.md` doesn't exist. The loop is 2/3 closed everywhere.

**7. The ID scheme deviates from the spec.** Root pins `id` as `B###` with anchor `#behavior-b###`; this PR uses `D/L/T/C/K/S`. Per-family prefixes are arguably *better* (globally unique, collision-free across files, readable in cross-refs like `S003` citing `D003`) — but the root spec is normative and unchanged, so today the six new files are non-conforming and the planned validation test would encode an undocumented rule. Please either amend the root LABT spec and bump `labt_version`, or use `B###`. I'd amend the spec.

**8. `src/lingtai/tools/shell/` and `src/lingtai/tools/telegram/` are new doc-only directories.** Neither exists on `main`; each now contains *only* `BEHAVIORS.md` — no code, no `CONTRACT.md`, no `ANATOMY.md`. The spec says a `BEHAVIORS.md` "sits beside a `CONTRACT.md` in the same directory." The shell implementation lives in `tools/bash/` (the file honestly points `contract: ../bash/CONTRACT.md`), and there is no telegram *tool* at all — telegram code is under `src/lingtai/mcp_servers/telegram/`. Suggest `tools/bash/BEHAVIORS.md` for the shell content and filing the telegram content beside the contract that owns it.

**9. `telegram` `related_files` point at files that do not exist**: `src/lingtai/tools/telegram/manager.py`, `src/lingtai/tools/telegram/SKILL.md`, `src/lingtai/tools/feishu/_family.py`. Worse, `C002`'s prerequisites name `telegram/SKILL.md` section `#reply-vs-send` as "the guard ... until a telegram root contract exists" — so `C002`'s only contract-ish linkage dangles.

**10. `telegram/BEHAVIORS.md` files three unrelated families.** `C003`/`C004` are `web_search`, `C005` is `file`, `C006` is `feishu`. The header calls this "comms/channels," but file-read pagination is not comms. `web_search/` and `file/` both already have real contracts with the exact anchors these LABTs cite — filed here, those two modules can never close their own loops, and a future `file/BEHAVIORS.md` will collide with `C005`.

---

### Non-blocking

- **Four `guards` contract names are wrong** (the `name:` in the target's frontmatter):
  - `shell` S004 → `psyche-contract`, actual is **`psyche-tool-contract`**. Also the anchor is `#behavior`, but the cited clause "Root reuse is not action compatibility" is a `###` under `## Purpose` → `#root-reuse-is-not-action-compatibility`.
  - `shell` S005 → `base-agent-contract`, actual is **`agent-runtime`**.
  - `kernel` K003/K004 → `context-tool`, actual is **`context-contract`**.
  - `tool_family` T006 → `tool-family-contract`, actual is **`tool-family`**.
  - (`notification-tool` in K005/K006 is correct.)
- **`kernel` K001/K002/K005/K006/K007 are pytest transliterated into markdown.** They monkeypatch privates (`kv._runtime_info`, `kv._fetch_latest_version`, `lc._capture_runtime_fingerprint`), build `MagicMock` agents, and assert internal call shapes — the opposite of the spec's "observable outcomes rather than internal call shapes." As markdown they're strictly *worse* than the pytest they supersede: no CI enforcement, no fixtures, and they silently rot when a private signature changes. **K003 and K004 are the right shape** (real schema/CLI introspection, no mocks). Suggest keeping the pytest for the other five and either dropping them or rewriting as observable checks — e.g. boot an agent against a stale installed distribution and read `.notification/nudge.json`.
- **`shell/BEHAVIORS.md` hardcodes an operator's personal path**, `C:\Users\zhuang\.lingtai-tui\runtime\venv\Scripts\python.exe`. It does say "substitute your own," but this shouldn't ship in-repo.
- **`shell` S001 step 6 asks the executor to run a pytest that fails by design on Windows.** A pass criterion containing known failures is ambiguous — mark the step POSIX-only or drop it.
- **`kernel` declares `contract: ../tools/notification/CONTRACT.md`** but its LABTs guard three different contracts (`src/lingtai/`, `notification`, `context`), and `kernel/` has no `CONTRACT.md`. The single-valued `contract:` key can't express this; consider allowing a list in the schema.
- **`behavior_version: 1` on all six** while `system/BEHAVIORS.md` is at `2`. Legal (it's a per-file revision) — just confirm it's intentional.
- **Land the validation test with this PR, not after.** Issues 1 and 2 are exactly what it would have caught mechanically.

---

### Recommendation

**Changes needed.** Concretely, I'd split this:

1. **Land now:** `daemon/`, `context/`, `tool_family/` — these are ready and set the standard for everyone else.
2. **Land after a trim:** `kernel/` — drop or rewrite the five mock-based LABTs, fix the `context-tool` name.
3. **Rework before landing:** `telegram/` — the frontmatter bug, the missing `Pass / Fail`, and the `guards`/`runner` inversion are individually small edits, but the file also needs re-homing (issues 8–10).
4. **Re-home:** `shell/` → `tools/bash/BEHAVIORS.md`.
5. **Decide the ID scheme** (issue 7) before the validation test hardcodes it, then add the `[B###]` clause links (issue 6) — that's the half of the tridirectional loop this PR is really about.

Great work on the daemon and context files in particular — `D003` and `L005` are the ones I'd point new contributors at.

PR #1342, opened 2026-08-10
fix(shell): win32_job drain_pipes deadlock — never close() live reader streams

huangzesen:
## Summary

Fixes the Windows sync shell-tool timeout hang: **`win32_job.drain_pipes` deadlocks by calling `stream.close()` on streams whose CPython reader threads are still blocked in `read()` holding the buffered-IO lock.**

### Root cause
On Windows, CPython drains subprocess pipes with **reader threads** (`subprocess.py:1514-1550`). A thread blocked in `fh.read()` holds the buffered-IO lock for the whole duration of the read; `stream.close()` on that same object must acquire the same lock — **without a timeout**. So the post-kill "bounded drain" (0.5s) deadlocks in exactly the situation it was written to survive:

- command times out → tree killed → `drain_pipes` runs → EOF not reached within 0.5s (grandchild still holding write ends) → `TimeoutExpired` → the close-loop runs → **blocks forever** → `_timeout_error` never returns → no `tool_result` ever → `finally: close_handle(job)` never runs (leaks job handle + loses `KILL_ON_JOB_CLOSE` backstop).

Observed live on the PowerShell agent (2026-08-10): sync `git rebase --continue` with `timeout=120` produced no result for 7+ minutes. `git rebase --continue` spawning a non-interactive editor (notepad/vim under `CREATE_NO_WINDOW`) is a realistic >120s command hang; the defect is that the timeout *enforcement* worked but timeout *reporting* deadlocked.

### Fix
1. **`drain_pipes` never `close()`s a live-reader stream.** Instead it detaches `process.stdout`/`process.stderr` (`= None`) and returns the partial output. The reader threads are already `daemon=True` (cannot block interpreter exit) and finish on their own once the killed tree releases the pipe write ends.
2. **Wait for the job to actually drain before the pipe drain** (`wait_job_empty(job, IO_DRAIN_TIMEOUT_SECONDS)` after `terminate_owned_tree`): in the common case real EOF arrives and the drain returns full output rather than the timeout partial.
3. **Fix `no_output` detection on the Windows contained path**: Windows `communicate` raises the bare `TimeoutExpired` (no `output=`/`stderr=`), so `exc.stdout`/`exc.stderr` are always `None` and `no_output` was unconditionally `True`. Now it is derived from the partial the bounded drain actually collected.
4. **`taskkill_tree_best_effort` runs with `DEVNULL` pipes** instead of `capture_output` (its output is discarded anyway), eliminating the same `close()` deadlock class in the fallback path.

### Tests
- **New regression test** `test_drain_pipes_never_blocks_on_live_reader_stream`: real `os.pipe()` + live daemon reader threads, runs `drain_pipes` on a worker thread and asserts it returns within 2s (the old close-loop fails this test instead of hanging the suite). Mutation-verified: reverting to the close-loop makes the test fail.
- Updated detach-semantics assertions (`process.stdout is None`) in the existing drain tests.
- Wiring test now asserts `wait_job_empty` fires between terminate and drain.
- **96 passed** in `test_win32_job.py` + `test_layers_bash.py` + `test_bash_shell_dialect.py`. 1 pre-existing base failure (`test_sync_contained_delivers_stdin_bootstrap_script`, "stdin_script requires the argv form") is byte-identical on clean `main` and unrelated.

### Follow-ups (not in this PR)
- Finding 3: the `OSError` fallback in `_run_sync_contained` still uses plain `subprocess.run`, whose Windows timeout path has the same documented hang class — deserves the same bounded treatment.
- Finding 4: `JOB_OBJECT_LIMIT_BREAKAWAY_OK` lets a descendant escape the kill while holding pipe write ends; worth reconsidering as a containment tightening.
- Operational mitigation already applied: set `GIT_EDITOR=true` + `GIT_TERMINAL_PROMPT=0` in the Windows agent env so `git rebase --continue` cannot block on an editor.

Full evidence: `tmp/windows-timeout-hang-investigation.md` on the reporting agent.

### DIFF
--- a/src/lingtai/adapters/windows/win32_job.py
+++ b/src/lingtai/adapters/windows/win32_job.py
@@ -15,7 +15,11 @@ The bounded output drain (``drain_pipes``) mirrors Codex's ``io_drain_timeout``
 (``codex-rs/core/src/exec.rs``): after a kill, stdout/stderr pipes are drained
 for only a bounded window, because a grandchild that inherited the pipe write
 ends and survived the kill keeps them open and would block an unbounded reader
-forever (Goose PR #7689).
+forever (Goose PR #7689).  The drain never ``close()``s a stream whose reader
+thread is still alive: on CPython the reader thread holds the buffered-IO lock
+for the duration of a blocking ``read()``, so ``close()`` would wait on that
+same lock forever.  The pipe reader threads are daemon threads and finish on
+their own once the killed tree releases the write ends.
 
 Importing this module is safe on every platform; every Job Object helper raises
 ``OSError`` on non-Windows at call time.  ``taskkill_tree_best_effort`` is the
@@ -344,7 +348,7 @@ def taskkill_tree_best_effort(pid: int) -> None:
     try:
         subprocess.run(
             ["taskkill", "/PID", str(pid), "/T", "/F"],
-            capture_output=True, text=True, encoding="utf-8", errors="replace",
+            stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
             timeout=_TASKKILL_TIMEOUT_SECONDS, stdin=subprocess.DEVNULL,
             creationflags=CREATE_NO_WINDOW,
         )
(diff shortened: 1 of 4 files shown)


huangzesen:
### CI triage (read-only, 2026-08-10)

PR #1342 current: head `22564abd` (2 commits: `ee5bad4` win32_job drain fix + `22564ab` docs), base `main`, 4 files +124/-22.

**Failed run** 31356407044 @ `ee5bad4` (04:43:51Z) / job 93356820454 Native Windows PowerShell 7: `tests/test_shell_windows_native.py::test_native_job_object_cancel_terminates_descendant_tree` — `PermissionError: [WinError 5]` at `_async_supervisor.py:54` `os.replace` (`_write_state_atomic`), `1 failed, 73 passed`.

**Control**: head `22564abd` (identical win32/bash code; docs-only delta) PASSED the same Shell PowerShell PR contract at 04:57:49Z (run 31357117505, job Native Windows/PowerShell 7 success). The failing test file SHA is byte-identical base (`28e43544`) vs head (`28e43544`). The PR diff does not touch `_async_supervisor.py` nor `tests/test_shell_windows_native.py`.

**Verdict**: recurring Windows async-supervisor contract flake; HOLD / not attributable to this PR. Next gate: author clean rerun or a clean-base control.

PR #1372, opened 2026-08-12
fix(wechat): harden outbound media uploads

runyuan-wang:
## Summary

- implement Tencent/OpenClaw-compatible encrypted WeChat media upload metadata and type-specific size fields;
- prefer the official static CDN upload construction when `upload_param` is present, while retaining the dynamic signed URL only as an absent-parameter compatibility fallback;
- retry only the encrypted CDN POST, at most three immediate attempts, for transport/TLS, HTTP 429/5xx, or a success response missing usable encrypted metadata;
- redact secret upload URLs, response bodies, local paths, recipient identifiers, unsent text suffixes, and raw provider errors from tool-visible failures;
- preserve truthful partial-delivery state for combined text+media and later text-chunk failures;
- treat final media transport/429/5xx outcomes as remote acceptance unknown and disable automatic retry, including media-only sends;
- make timeout cancellation task-owned: after a media task starts, wait for its real exit before returning; if scheduling is accepted but the task never starts, fail within a separate one-second handshake and prevent late launch;
- document the corrected retry, timeout, partial-delivery and manual-reconciliation behavior and add focused regressions.

## Why

The previous outbound path could report generic failures while exposing internal URLs or paths, could retry at the wrong layer, could lose the delivered prefix after a later text-chunk failure, could invite duplicate media sends when final acceptance was unknowable, and could return a terminal timeout while media work continued. It also did not follow the official static CDN upload route when `upload_param` was available.

This patch makes the protocol choice explicit, constrains retries to the encrypted CDN byte upload, preserves only redacted partial state, disables unsafe whole-message retries, and closes both post-start and pre-start task-detachment races.

## Evidence boundary

This is a source/regression fix, not proof of production media recovery. No real WeChat media upload was run. The actual dynamic signed upload host, exact failing substage and TLS root cause remain unknown. A cancellation-resistant child may extend wall-clock time beyond the nominal timeout because the caller now waits for its real exit rather than returning while it continues.

## Validation

- focused media-upload + outbound-containment tests: **59 passed**
- exact eight WeChat modules + outbound containment: **202 passed**
- architecture-document tests: **4 passed**
- no-bytecode compile of the changed runtime modules and affected test files: **passed**
- `git diff --check`: **passed**
- exact six-file scope, staged/untracked/debris and public-patch privacy gates: **passed**

## Out of scope

- merge, release, runtime/config/refresh or TLS-policy changes;
- real-media end-to-end acceptance;
- proving which dynamic signed hostname failed;
- Windows-style legacy basename display and inherited inbound/download exception redaction;
- the separate OpenAI/Anthropic read-timeout proposal.

Telegram: @wangrunyuansecbot | Nickname: 一心

### DIFF
--- a/src/lingtai/mcp_servers/wechat/SKILL.md
+++ b/src/lingtai/mcp_servers/wechat/SKILL.md
@@ -125,14 +125,22 @@ ATTACHMENTS below), not a mutually exclusive choice.
 - `send` and `reply` deliver to real users — external side effects. Confirm
   recipient and content before sending unsolicited messages.
 - Actions return a result dict on success or `{'error': <message>}` on failure
-  (e.g. missing `user_id`, unreadable `media_path`). If a combined text+media
-  send delivers text and the later media step fails, the result instead has
-  `status: 'partial'`, `partial_delivery: true`, `text_status: 'sent'`,
-  `media_status: 'failed'`, a redacted `failure` stage, and
-  `automatic_retry_allowed: false`. Do not repeat the whole call automatically;
-  reconcile the delivered text before retrying media alone. The CDN upload step
-  itself makes at most three immediate attempts for transport/TLS failures,
-  HTTP 429/5xx, or a success response missing its encrypted media reference;
-  other 4xx responses are not retried. A local media deadline requests coroutine
-  cancellation, but an already accepted remote request cannot be revoked. Check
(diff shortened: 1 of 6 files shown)


huangzesen:
Superseded by #1387 (merged 51663ed5) which implements the official static CDN URL priority + retry in a minimal 2-file change. Closing per maintainer direction; credit to @runyuan-wang for the original upload improvement.

PR #1391, opened 2026-08-14
feat(daemon): add supervisor return observer

runyuan-wang:
## Summary

Add an opt-in **Supervisor Return Observer** at the daemon terminal return seam.

When enabled for a newly dispatched daemon, the supervisor invokes a short isolated helper after terminal state/notification ownership is claimed and before terminal publication. The helper validates the exact dispatch/return generation, freezes source-bound receipt files, and returns a bounded advisory block that may be attached to the terminal notification.

The raw daemon result and ordinary terminal notification remain authoritative.

## What changed

- Snapshot `return_observer_enabled` into the immutable supervisor manifest at dispatch time.
- Add a thin fail-open supervisor hook and isolated no-network helper subprocess.
- Write bounded `host`, `portable`, and `status` receipts under the daemon supervisor directory with restrictive modes.
- Validate generation identity, result/artifact bytes, predecessor linkage, read completeness, supersession, path containment, and concurrent same-key publication.
- Keep bounded generations available when declared advisory artifacts exceed the per-file or total receipt-byte budget by recording explicit metadata-only omission instead of reading or hashing omitted content.
- Add only a privacy-safe additive `return_observation` notification block when a valid receipt is available.
- Document the seam and invariants in the daemon anatomy.
- Add focused fault-injection, integrity, concurrency, recovery, privacy, notification, terminal-state, and oversized-artifact regressions.

## Safety and compatibility

- **Default off.** Only `1`, `true`, `yes`, or `on` enable the observer.
- **Fail open to the raw result.** Import, spawn, timeout, crash, malformed output, receipt, or validation failures never suppress normal daemon delivery.
- **Raw first / additive only.** No daemon result bytes, completion state, `daemon.check` behavior, or public tool schema are replaced.
- **Advisory, not semantic judgment.** A valid receipt proves binding/integrity properties; it does not prove that the daemon's conclusion is correct.
- **Bounded omission is explicit.** An oversized or over-total-budget declared artifact is represented by relative reference, reason, observed size and cap (plus any declared-versus-observed size difference); it is absent from content-observed files and receives no content-digest claim.
- **No scheduler/resource claim.** This does not throttle or accelerate high-concurrency daemon execution.
- **Per-dispatch snapshot.** Existing daemon manifests are unchanged; the flag applies to new dispatches.

Operational rollback for an ordinary refresh should set `LINGTAI_DAEMON_RETURN_OBSERVER_ENABLED=0` explicitly. Deleting the key is not a reliable rollback when a parent process already inherited a truthy value. A cold launcher must likewise avoid exporting the flag as true when the observer should be off.

## Sanitized real-use reliability evidence

Two genuine same-batch concurrent daemon returns invoked the helper successfully (exit 0, empty stderr), but their declared transcript and event artifacts were each above the unchanged 262,144-byte per-file observation cap. Raw terminal delivery remained correct while the additive `g0000` generation was absent.

The correction preserves the cap. For oversized or over-total-budget advisory artifacts it records metadata-only omission, preserves declared-versus-observed size differences, does not read or claim a content digest for omitted files, and leaves raw results, terminal state, and exactly-one logical terminal notification unchanged.

Final local gates passed: **71 focused observer tests + 1 parent metadata-mismatch probe + 12 adjacent terminal tests**. Three copied real fixtures produced nine portable receipts in diagnostic/direct/wrapper paths, and `g0000` was restored for all three. This is bounded real-use reliability evidence for the observer seam, not an optimization, token-benefit, general quality, semantic-truth, security, authorization, or production-fitness claim.

No private transcript/event content, local path, daemon identifier, internal message identifier, credential, receipt body, or hidden-file content is included here.

## Validation

- Focused observer regressions: **71 passed**.
- Parent oversized-metadata mismatch probe: **1 passed**.
- Adjacent terminal suites: **12 passed** (1 detached exactly-once publication + 5 daemon notification/receipt + 6 run-directory terminal tests).
- Copied real-fixture reproduction: **3 fixtures**, direct and wrapper observation available for each, **9 portable receipts**, and `g0000` restored for all three.
- Disposable copied canary: OFF / ON / ON-helper-failure, repeated twice: **6/6 passed** with one ordinary terminal notification per run and unchanged raw results.
- Changed Python files: `py_compile` passed.
- `git diff --check` passed.
- Architecture-document test: candidate and frozen base have the exact same pre-existing result (**3 passed, 1 failed** in the same graph-coverage test); this patch adds no architecture-test regression.
- Wider adjacent daemon run from the accepted migration: **286 passed, 14 failed**; the exact same 14 failures reproduced on the frozen base from existing optional-MCP/default-max environment conditions.

## Scope

Exactly seven files are changed in the PR: the supervisor manifest/runtime seam, daemon anatomy/export surface, two observer modules, and one focused test module. The oversized-artifact correction changes only the existing helper and focused test files. No runtime configuration, deployment, live-channel behavior, merge, or release is included in this PR.

Telegram: @wangrunyuansecbot | Nickname: 一心

### DIFF
--- a/src/lingtai/kernel/daemon_supervisor/manifest.py
+++ b/src/lingtai/kernel/daemon_supervisor/manifest.py
@@ -265,9 +265,10 @@ def build_manifest(
     backend_argv: list[str] | None = None, language: str = "en",
     preset_name: str | None = None, preset_llm: dict | None = None,
     preset_capabilities: dict | None = None,
+    return_observer_enabled: bool = False,
 ) -> dict:
     """Build the supervisor input record without resolved credentials."""
-    return {
+    manifest = {
         "schema": MANIFEST_SCHEMA,
         "run_id": run_id,
         "backend": backend,
@@ -288,6 +289,9 @@ def build_manifest(
         "preset_llm": _safe_llm(preset_llm),
         "preset_capabilities": _safe_json(preset_capabilities) if isinstance(preset_capabilities, dict) else None,
     }
+    if return_observer_enabled:
+        manifest["return_observer_enabled"] = True
+    return manifest
 
 
 def write_manifest(run_dir: Path, manifest: dict) -> Path:
@@ -327,6 +331,8 @@ def read_manifest(path: Path) -> dict:
         raise ValueError("manifest max_turns must be a positive integer")
     if isinstance(data["timeout_s"], bool) or not isinstance(data["timeout_s"], (int, float)) or data["timeout_s"] <= 0:
         raise ValueError("manifest timeout_s must be positive")
+    if "return_observer_enabled" in data and not isinstance(data["return_observer_enabled"], bool):
+        raise ValueError("manifest return_observer_enabled must be a boolean")
     run_dir = Path(data["run_dir"]).resolve()
     if run_dir != path.parent or run_dir.name == "":
         raise ValueError("manifest run_dir does not match its canonical parent directory")
--- a/src/lingtai/tools/daemon/ANATOMY.md
+++ b/src/lingtai/tools/daemon/ANATOMY.md
@@ -50,6 +50,8 @@ related_files:
   - src/lingtai/adapters/posix/process_identity.py
   - src/lingtai/adapters/windows/daemon_supervisor.py
   - src/lingtai/tools/daemon/runtime.py
+  - src/lingtai/tools/daemon/return_observer_hook.py
+  - src/lingtai/tools/daemon/return_observer_helper.py
   - src/lingtai/tools/daemon/glossary-en.md
   - src/lingtai/tools/daemon/glossary-zh.md
   - src/lingtai/tools/daemon/glossary-wen.md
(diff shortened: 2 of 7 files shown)


huangzesen:
Review note (triage, 2026-08-14): the current diff is large for its observable benefit — about 870 lines of new logic (return_observer_helper + hook) plus ~1,000 lines of tests, touching the sensitive terminal-notification seam. The feature is default-off and fail-open, so runtime risk is low, but the PR does not state what real problem it solves: no linked issue, no referenced incident, no motivation section.

Before this can be considered for merge, please describe the concrete motivation — e.g. a real case where a terminal notification was misleading (cross-generation, superseded, or tampered result), or the specific consumer that needs this audit trail. If the goal is only to tell consumers which generation a notification belongs to, a much smaller change (including the generation id in the existing notification) would likely cover it.

Not merging at this time pending that clarification.

Telegram: @lingtaidev1bot | Nickname: 知微

runyuan-wang:
A product-positioning note after broader real use: LingTai's current daemon return path already binds a task instance to a dedicated run, terminal event, artifact/result paths, and an additive return-observation receipt. In our current 15-way concurrent run, we did not find a cross-task mix-up.

That means LingTai itself may not need RouteCertificate in the kernel for this problem. The more valuable target may be other harnesses whose task → terminal → artifact → result chain is weaker.

Please treat this PR as an experiment and design reference, not as an assertion that LingTai needs the feature. I am leaving it open for maintainer judgment; I am not requesting merge on the strength of the current evidence.

huangzesen:
Closing as not planned. This observer experiment is not planned for the LingTai kernel roadmap at this time.

PR #1409, opened 2026-08-18
fix(nudge): keep folder-size probes off heartbeat

rawpaper123:
## Summary

- move the recursive working-directory size walk off the 1-second heartbeat
- keep one daemon probe in flight per canonical working directory
- let heartbeat code alone consume results and mutate persistent nudge state
- preserve the previous observation while a new UTC-day measurement is pending
- discard results that complete after their UTC start date and retry for the current day

## Why

A roughly 45 GB working directory made the synchronous recursive walk exceed the
120-second watchdog window. The agent was killed, restarted, and immediately
entered the same scan again. This change makes heartbeat latency independent of
directory traversal time while retaining the existing daily gate and nudge policy.

## Behavioral boundaries

- no new dependency, scheduler, environment variable, or synchronous fallback
- dynamic threshold, dismissal/repeat/enable, notification retry, and symlink semantics remain unchanged
- a probe worker publishes only an in-memory result/error; persistent writes stay on heartbeat
- strict singleflight is retained if a filesystem call wedges; fresh observation then waits for process restart rather than leaking concurrent scans, while the prior persisted observation remains active

## Tests

- TDD regression first reproduced a heartbeat blocked for about 3 seconds
- `17 passed, 1 deselected` for `tests/test_folder_size_nudge.py` on Windows; the deselected existing symlink test requires a Windows privilege unavailable in this environment
- async concurrency/cross-day/error cases repeated 5 rounds successfully after final review
- direct Nudge Anatomy governance check passed
- real large-workdir smoke: `check()` returned in `0.000480s` while the background traversal started
- independent reviewer found no blocking issues after the cross-day pending-observation fix

## Full-suite limitation

The complete Windows suite cannot collect in this environment because existing
POSIX-only tests import `fcntl`. The repository-wide docs gates also have existing
unrelated baseline failures on current `main`; the changed Nudge Anatomy document
passes its focused governance validation.

## Architecture documentation

`src/lingtai/kernel/nudge/ANATOMY.md` now records the background singleflight
probe, heartbeat-only persistence boundary, and stale cross-day result behavior.
No public Contract or BEHAVIORS change is required because this corrects the
existing non-blocking heartbeat implementation invariant without adding a new
agent-facing capability.

### DIFF
--- a/src/lingtai/kernel/nudge/ANATOMY.md
+++ b/src/lingtai/kernel/nudge/ANATOMY.md
@@ -18,6 +18,7 @@ related_files:
   - src/lingtai/CONTRACT.md
   - tests/test_nudge_prompts.py
   - tests/test_kernel_version_nudge.py
+  - tests/test_folder_size_nudge.py
   - ENVIRONMENT_VARIABLES.md
 maintenance: |
   Keep related_files as repo-relative paths to real files. Include neighboring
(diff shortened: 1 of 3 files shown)


huangzesen:
Superseded by #1486 (`e831fc66`), whose current-main `NudgeObservationOwner` / nonblocking observation path keeps recursive folder-size walks off heartbeat under the shared owner. The possible UTC-rollover freshness edge is separate and unproven; this stale conflicting branch should not be revived. Preserving @rawpaper123's root-cause analysis and implementation credit. Closing under Jason's instruction to close all verified superseded PRs.

PR #1465, opened 2026-08-22
Add karma-gated target refresh action

BatalloLu:
## Summary

Add a karma-gated `system(action="target_refresh")` operation so manager/project-lead agents with karma can request a running target agent to refresh itself without using the heavier suspend→CPR path.

Behavior:
- Adds `target_refresh` to the system tool action/schema/registry/docs.
- Reuses the existing target-agent `.refresh` signal mechanism.
- Requires karma and the usual target-agent address checks.
- Requires the target to be currently running/alive before writing `.refresh`.
- Does not alter the target preset/config; it only asks the target runtime to reload from its current disk state.

## Validation

Focused validation rerun before PR:

- `python -m compileall -q src/lingtai/tools/system`
- `python -m pytest tests/test_karma.py -k target_refresh -q` → 3 passed
- `python -m pytest tests/test_tool_family_system_migration.py -k 'legacy or self_actions or schema or child_names' -q` → 2 passed
- `git diff --check`

Broader notes from local checkout:
- Full `tests/test_karma.py` passed earlier: 26 passed.
- Full `tests/test_tool_family_system_migration.py` has an existing unrelated sleep notification guard failure after 109 passed.
- `tests/test_architecture_documents.py` exposes pre-existing ANATOMY graph issues outside this patch; not modified here.

## Safety boundary

This PR only adds the action implementation/docs/tests. It does not change live runtime, model/preset/key/MCP/config, or release behavior.


huangzesen:
Closing as superseded by #1507, which explicitly credits this PR and its original author and carries the target_refresh work forward with broader governed implementation/documentation and mounted target-handshake coverage. Thank you for the original proposal.

Please keep further target_refresh integration and current-main reconciliation in #1507 rather than maintain two competing implementations. The successor still requires its own revisions and acceptance; closing this PR is not an approval or merge of #1507. Attribution and branch history are preserved.

Disposition is based on source/history comparison, not a new test run.

PR #299, opened 2026-06-16
Add claude-code provider: drive the claude CLI on a Claude subscription

zechenzhangAGI:
## Add `claude-code` provider: drive the `claude` CLI on a Claude subscription

Adds a `claude-code` LLM provider so a LingTai agent can think on a Claude
Pro/Max **subscription** via the official `claude` CLI, instead of an Anthropic
API key. It is the Claude analogue of the existing `codex` provider.

### Design
The `claude` CLI is used in print mode (`claude -p --output-format json`) as a
**stateless reasoning core**. Each turn the adapter serialises the canonical
`ChatInterface` (system prompt + tool schemas + conversation) into one prompt and
asks the CLI to emit exactly one JSON *action* — a tool call or a final answer —
which is parsed back into the kernel's `LLMResponse`. LingTai's own message loop
still executes the tools, so the main agent loop, intrinsics, capabilities and
MCP all stay intact; the CLI's built-in tools are disabled so it acts as a pure
brain.

### Auth / ToS
Auth is owned entirely by the `claude` CLI (its stored OAuth credentials or
`CLAUDE_CODE_OAUTH_TOKEN`). The adapter never reads, stores, or replays a token —
it only strips `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` from the child env so
the subprocess can't fall back to API-key billing. This keeps usage on the
sanctioned Claude Code / subscription channel (per Anthropic's 2026 credit-pool
policy, third-party tools authenticating through the Agent SDK / CLI draw from a
metered subscription credit), rather than the prohibited raw-token "harness
spoofing" path.

### Changes
- `src/lingtai/llm/claude_code/{__init__,adapter,defaults}.py` — adapter + session
- `src/lingtai/llm/_register.py` — register the keyless `claude-code` factory
- `src/lingtai/llm/ANATOMY.md` — document the provider
- `tests/test_claude_code_adapter.py` — 19 CI-safe unit tests (subprocess mocked)
- `tests/integration_test_claude_code.py` — live end-to-end test vs the real CLI

### Testing
- 19/19 unit tests pass (mocked subprocess, no CLI needed).
- 2/2 live integration tests pass against the real `claude` CLI.
- Existing `tests/test_adapter_registry.py` still green (no regression).
- Full-stack: a real `Agent(provider="claude-code", capabilities=["write"])` was
  started, reasoned via the CLI, called its `write` tool, and produced the exact
  file contents — the whole BaseAgent loop on a subscription.

[...]

PR #1545, opened 2026-08-28
Phase B: Driver-backed derived admission

TZZheng:
## Status

**Review candidate after blocker repair — new independent review required; no merge approval.**

- LingTai head: `0b233800bb0dfe4b3ab8f59fcedceccd5925715d`
- Base: `bb8d1136c1e3f4be1a7cc124126f56db681f4bd4` (`main`)
- Puffo target: `695972bbd8d5aeea6657e942d8a3864198465c0c`

The branch was rebased onto current `main` and is merge-clean. Earlier pin-specific
evidence was not reused: the focused and cross-repository gates below were rerun on
this rebased content. This is a request for independent review, not approval to merge.

## Delivered behavior

- A required derived child with no Driver port returns
  `INDETERMINATE / required_derived_launch_admission_port_missing`; it is never
  reported as a nested-policy denial.
- Timeout invalidation is serialized with the authority exchange, so a concurrent
  request cannot consume a timed-out request's late response. The child remains
  fail-closed until its owner restarts or recomposes it.
- Non-indeterminate Driver grants and denials require a non-empty `audit_id`.
- POSIX relay setup releases resources on failure. A rejected authority endpoint
  receives one explicit local disposal attempt; at the pinned Puffo revision,
  peer EOF closes an unclaimed Driver record.
- A truncated SCM_RIGHTS response records and closes every visible descriptor before
  returning an indeterminate transport decision.
- A granted child-endpoint lease is deterministically released on every Avatar or
  Daemon pre-handoff return/exception; a launcher/supervisor that consumes it owns
  the successful handoff path.
- BaseAgent, daemon, and supervisor source-of-truth contracts describe the delivered
  first-hop, execution-child authority, and nested-audit evidence.

## Reproducible review gates

Build the supported interpreter from this exact checkout. Puffo is source-only:
do not inject Puffo runtime site-packages, whose MCP 1.x dependency conflicts with
LingTai's MCP 2.x runtime.

```bash
scripts/setup_driver_supervisor_e2e_env.sh .venv-driver-e2e
git clone https://github.com/puffo-ai/puffo-agent.git /tmp/puffo-agent
git -C /tmp/puffo-agent checkout 695972bbd8d5aeea6657e942d8a3864198465c0c
```

The setup script installs LingTai with `mcp==2.1.1` and
`agent-client-protocol==0.10.1`. The E2E probe independently requires those
exact active-interpreter versions plus `mcp.server.ServerRequestContext` before
product setup, then verifies and prints the actual LingTai and Puffo Driver
module paths so a site-packages Driver cannot silently replace `--puffo-src`.
Under `-I -S`, its controlled bootstrap appends (rather than processes) the
single canonical venv `site-packages` path: no `.pth` file is executed. The
supported MCP/ACP environment used by these gates is therefore required not to
depend on editable-install or other `.pth` path processing.

```bash
.venv-driver-e2e/bin/python -I -S scripts/verify_driver_supervisor_execution_e2e.py \
  --lingtai-src "$PWD/src" --puffo-src /tmp/puffo-agent/src

.venv-driver-e2e/bin/python -I -S scripts/verify_driver_lease_lifecycle.py \
  --lingtai-src "$PWD/src" --puffo-src /tmp/puffo-agent/src
```

This is the real root Agent → `PosixDaemonSupervisorAdapter` → supervisor
entrypoint → execution-child entrypoint → admitted provider-call path, using a
real Puffo Driver. It must print two distinct process IDs, one launch audit, and
the child provider audits. Its business verdict is emitted before best-effort
temporary-directory cleanup.

For the must-red relay control, remove only `authority_fd` from the supervisor
`pass_fds`, rerun the same command with:

```bash
--expect-failure derived_admission_port_unconnected
```

It must accept only that precise diagnostic; restored source must pass again.

Historic behavior gates remain tied to their published evidence pins; they are
not represented here as fresh `63bdfdc1` reruns. The following provenance
delta was freshly checked on the immediately preceding `bc615515` content.
The two subsequent commits narrow the isolated-hop bootstrap and add focused
assertions:

- the real first-hop E2E and the lease-lifecycle oracle both pass against the
  fixed Puffo SHA, printing their actual loaded sources and fixed revision;
- `tests/test_driver_authority_adapter.py`, `tests/test_avatar_launcher.py`,
  `tests/test_layers_avatar.py`, `tests/test_daemon.py`,
  `tests/test_provider_admission.py`, and `tests/test_driver_probe_provenance.py`:
  **273 passed** at `bc615515`;
- at this exact head,
  `test_isolated_supervisor_and_execution_child_keep_startup_flags` and the
  ordinary non-isolated detached-run acceptance test both pass, and an
  inherited fake-LLM environment variable is proven unable to add `tests/`
  without explicit process-local configuration, and the ordinary non-isolated
  detached-run acceptance test passes (**4 passed**),
  and the real isolated two-PID E2E passes with no injected-sitecustomize log;
- a child-only `sitecustomize` wrapper which preloads a forged `lingtai` root
  now fails before product import as `PUFFO_PROVENANCE_INVALID`;
- a timestamp/size-aligned, Git-ignored `puffo_agent/__init__.cpython-313.pyc`
  is rejected before product import by both lifecycle and E2E probes.

The three ownership regressions below were first reproduced against the preceding
head and now pass at this head:

- a `MSG_CTRUNC` reply with eight visible SCM_RIGHTS descriptors leaves no duplicate
  open after the adapter returns `INDETERMINATE`;
- an Avatar duplicate-directory return makes its granted lease unconsumable;
- a Daemon manifest-preparation exception makes its granted lease unconsumable.

The post-rebase change in `f2b113e9` only refreshes the static concurrency inventory
line numbers after current-`main` insertions; it contains no behavior change. The
following `b8aefe9c` change is limited to SCM_RIGHTS/lease cleanup and its regressions;
`f04f01e4` adds the explicit cross-repository lease-lifecycle probe and narrows
the ownership wording. `1bde83c2` waits for that Driver worker to terminate
after observing `CLOSED`, and refreshes the final daemon concurrency inventory.
`a9a91637` makes the oracle bind its result to the documented Puffo revision,
joins denied-claim response IDs to Driver audit records, and makes the first-hop
probe enforce exact MCP/ACP versions plus source-only module provenance. It also
corrects the Contracts: only a child-endpoint *lease* is one-use, and the
relay-loss mutation is a separate temporary validation action rather than a
committed green-only script mode. `d71cff7f` updates the existing avatar-restart
fixture to return the same socket-backed, audit-linked `DriverChildEndpointLease`
shape accepted by production; a bare `object()` is intentionally malformed and
must remain fail-closed. `187c8628` closes the remaining lifecycle-provenance
gap: the lifecycle oracle rejects a dirty Puffo checkout before import and
asserts/prints that its loaded Driver module comes from `--puffo-src`, rather
  than a preloaded or site-packages replacement. `b418d51b` makes those claims
  canonical: both probes require `--puffo-src` to be the checkout's canonical
  `src/`, require a clean fixed revision, and compare the actual loaded Driver
  module with the tracked `HEAD` blob. The E2E helper also emits an interpreter/
  source/version manifest; the probe requires that manifest and verifies the
  runtime `mcp` and `acp` module origins match their pinned distributions. The
  current follow-up additionally rejects preloaded trusted package trees before
  any controlled import; it verifies import spec, loader, package path, and
  loaded code origin for the actual Driver, `mcp`, `mcp.server`,
  `ServerRequestContext`-defining module, and `acp` objects. The regression
  suite includes forged preloaded Driver/MCP/ACP modules whose `__file__` is
  made to look canonical: both probes must fail before product setup.
  The probes also re-exec in a fresh `PYTHONPYCACHEPREFIX` and inspect ignored
  current-interpreter authority-package entry bytecode artifacts. Their recursive marshalled
  code graph must equal a controlled compilation of the tracked `HEAD` blob;
  a timestamp/size-aligned ignored `driver_authority_server.pyc` with a nested
  payload is rejected even when every path and metadata field is canonical.
  The outer probe never adopts a caller-supplied `PYTHONPYCACHEPREFIX`: it
  creates a fresh private prefix and replaces that value for its controlled
  child process. The loaded Driver cache path must remain under that private
  prefix. `-B`/`dont_write_bytecode` are not used as cache-read controls.
  `63bdfdc1` also makes both probes reject the same preloaded trusted roots
  (`mcp`, `acp`, `puffo_agent`, and `lingtai`) before controlled import, so a
  child-only wrapper cannot select LingTai code between the outer and child
  guards. `9b298921` begins the two-level startup boundary; `034513a3` requires
  both verifier levels to use `-I -S` and restores only the canonical
  venv site-packages path derived from `sys.executable`. Its controlled child
  also uses `-X pycache_prefix=<private>` after inherited `PYTHON*` variables
  are scrubbed. This is a supported-interpreter startup contract,
  not a claim that a Python script launched inside an already compromised
  interpreter can bootstrap trust for itself. `bc615515` propagates that
  `-I -S` bootstrap through the real supervisor and execution-child process
  hops whenever the verifier parent is isolated; ordinary production parents
  retain their existing direct `python -m` path. `24978f34` invokes those
  isolated modules with `run_name='__main__', alter_sys=True`, matching
  `python -m` entrypoint semantics rather than silently importing a module
  without running its main guard. `d75c2764` removes the inherited-environment
  selection of the test `sys.path` entry: only the E2E probe can configure the
  canonical source tree's `tests/` root in-process, and the bootstrap carries
  that already-validated value explicitly across subsequent detached hops.
  `76993bce` makes that bootstrap evidence executable: commands captured from
  supervisor, execution-child, and resume-owner launches each run a harmless
  test module whose only result is guarded by `__name__ == '__main__'`.
  The daemon-supervisor Contract now documents both startup branches and all
  three propagated hops. `0b233800` updates the kernel Port and POSIX
  supervisor-entrypoint module docstrings to the same direct-`-m` /
  isolated-bootstrap split.

The BaseAgent, daemon, and supervisor Contracts now list the first-hop setup/
probe as owner evidence; daemon also lists the dispatch probe and treats all
three as review triggers. Root-level navigation is therefore no longer the
only link governing the evidence their normative text cites.

## Scope and non-goals

The first-hop E2E and the real dispatch/admission seam are intentionally separate
instruments; neither replaces the other. Current v1 has no response-to-request
correlation. Atomic socket invalidation blocks the reproduced stale-frame bypass;
unique request IDs with echoed response validation are post-#304 v2 hardening.

Out of scope: #299 identity/workdir binding; provider CAS/exactly-once; dynamic
revoke; global generic-blacklist removal; `get_adapter()` capability-carrier
hardening; and async/cross-process audit propagation.


TZZheng:
Superseded by the reviewable Phase B stack: #1555 → #1572. The final stack head #1572 is tree-equivalent to this PR head; please review and merge the stack in order.

PR #1613, opened 2026-09-02
fix(nudge): gate kernel version probes by UTC day and installed version

### LINKED ISSUE #1577: [nudge] kernel_version persists but never reads last_remote_check_date
## What's wrong

`kernel_version` persists `last_remote_check_date` at five sites, with comments saying a failed remote check should not be retried again that day, but the field is never read. The only active throttle is an in-memory timestamp, so it resets on restart and permits repeated remote probes throughout the same day.

Sibling Nudge producers read their equivalent persisted date gate, which makes the missing read in this producer look accidental rather than intentional.

## Where

- Component: kernel-version Nudge remote-check cadence
- Current-main commit reviewed: `cf0e3b6bd498dcb0afdc8b1f94d5c59dd14fad99`
- Current implementation: https://github.com/Lingtai-AI/lingtai-kernel/blob/cf0e3b6bd498dcb0afdc8b1f94d5c59dd14fad99/src/lingtai/kernel/nudge/kernel_version.py#L223-L317
- Fast in-memory gate: https://github.com/Lingtai-AI/lingtai-kernel/blob/cf0e3b6bd498dcb0afdc8b1f94d5c59dd14fad99/src/lingtai/kernel/nudge/kernel_version.py#L156
- Tests: https://github.com/Lingtai-AI/lingtai-kernel/blob/cf0e3b6bd498dcb0afdc8b1f94d5c59dd14fad99/tests/test_kernel_version_nudge.py

## Reproduction

From the repository root:

```bash
rg -n "last_remote_check_date" src/lingtai
```

At the reviewed commit, all five hits are writes in `kernel_version.py`; there is no corresponding `get()` or comparison that gates a same-day retry.

Compare the sibling producers:

```bash
rg -n "last_check_date" src/lingtai/kernel/nudge
```

`folder_size.py` and `event_journal_count.py` both read their persisted date before repeating work.

One deployment with failing release-mirror access emitted repeated `kernel_version_update_check_error` events throughout the same day despite its persisted state already containing that day's `last_remote_check_date`. The exact retry volume is network/workload dependent; the deterministic defect is the missing read.

## What you expected

After a remote-manifest failure is recorded for the current UTC day, the probe should honor the persisted date gate described by the source comment and avoid repeated same-day work.

## What actually happened

The date is persisted but ignored. Only the in-memory short interval limits retries.

## Severity

minor

The version check still functions, but unreachable mirrors can cause repeated network requests and repetitive diagnostic events.

## Suggested fix (optional)

Load the persisted producer state before the remote probe and return when `last_remote_check_date` already equals today, alongside the existing short in-memory throttle. Add a test that seeds today's persisted date and asserts a second probe is skipped.

runyuan-wang:
## Summary

- Check persisted `last_remote_check_date` and `checked_installed_version` before starting a kernel-version remote fetch.
- Suppress a repeat fetch only for the same installed version on the same UTC day.
- Fail open for missing or malformed state, and re-arm on a version change or the next UTC day.

Closes #1577.

## Validation

- `python -m pytest -q tests/test_kernel_version_nudge.py` — **35 passed** on the final worktree.
- Coverage includes success/error/timeout persistence, UTC-day and version transitions, call counts, and existing single-flight behavior.
- `git diff --check` — PASS.

## Scope

This does not change global Nudge repeat/dismiss policy, state concurrency, or unrelated Nudge behavior.

Base reviewed: `38222152313823e2fa03dab75bf9f2db9f865592`.

### DIFF
--- a/src/lingtai/kernel/nudge/kernel_version.py
+++ b/src/lingtai/kernel/nudge/kernel_version.py
@@ -233,6 +233,8 @@ def check(agent) -> None:
     # each tick either starts one or consumes a finished one.
     pending = _fetch_slot(agent)
     if pending is None:
+        if not _remote_check_due(kernel_state, info.installed_version, today):
+            return
         _start_fetch(agent)
         return
     if not pending.done.is_set():
@@ -437,8 +439,18 @@ def _module_from_source_checkout(module_file: str) -> bool:
 
 
 def _remote_check_due(kernel_state: dict[str, Any], installed_version: str, today: str) -> bool:
-    """Bounded probe gate; repeat/dismiss semantics are global, never daily."""
-    return True
+    """Return whether this installed version has not been checked on this UTC day.
+
+    State is deliberately fail-open for malformed or missing values: a new
+    process must still get one observation. A completed probe (including a
+    failure) records both fields, so retries are suppressed until either the
+    UTC day or installed version changes. Nudge dismissal/repeat remains a
+    separate global policy concern.
+    """
+    return not (
+        kernel_state.get("last_remote_check_date") == today
+        and kernel_state.get("checked_installed_version") == installed_version
+    )
 
 
 def _fetch_latest_version() -> _ReleaseObservation:
--- a/tests/test_kernel_version_nudge.py
+++ b/tests/test_kernel_version_nudge.py
@@ -298,7 +298,7 @@ def test_match_after_refresh_keeps_independent_remote_finding_during_outage(tmp_
     assert _entries(tmp_path)[0]["source"] == kv._GITHUB_SOURCE
 
 
-def test_remote_update_check_uses_bounded_probe_not_daily_product_cadence(tmp_path, monkeypatch):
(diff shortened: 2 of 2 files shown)


huangzesen:
Closing without merge because the proposed daily producer gate reinstates policy that was deliberately replaced, rather than repairing an accidental omission.

Merged #966 moved cadence/visibility policy to shared global Nudge controls. Its commit https://github.com/Lingtai-AI/lingtai-kernel/commit/92eeaaffa575935ff3414856f313ddf4378e3c85 explicitly removed the same day/version guard, documented bounded probe cost, and changed the regression to permit a same-day second fetch. This PR reverses that intentional behavior.

If remote-probe frequency should change again, #1577 can be reconsidered as an explicit policy decision with the global Nudge contract, implementation comments and regressions updated together. The presence of persisted date/version fields alone does not establish that the old gate should return. Issue #1577 is left unchanged.

Disposition is based on exact source/history comparison; no new test run is claimed.

PR #319, opened 2026-06-17
feat: add minimal mcp manager actions

huangzesen:
## Summary
- turn the `mcp` intrinsic from a manual/show surface into a minimal manager with only `list`, `add`, and `remove`
- make `add` register and activate an MCP in `init.json`, and `remove` deregister/deactivate it while stripping same-name `addons`
- keep refresh ownership in `system(action="refresh")`; mutating results return `needs_refresh: true` and include the refresh reminder
- update MCP manual/anatomy plus tests, with redaction and bad-`init.json` safety coverage

## Validation
- `PYTHONPATH="$PWD/src" python -m pytest -q tests/test_mcp_manager.py tests/test_mcp_capability.py tests/test_mcp_closed_resource_restart.py tests/test_mcp_inbox.py tests/test_mcp_licc_client.py tests/test_sdk_namespace_mcp.py tests/test_sdk_namespace_fileio.py` → 148 passed
- `git diff --check`
- `PYTHONPATH="$PWD/src" python -m pytest -q` → 2314 passed, 3 skipped

## Notes
- No `show`, `diagnose`, `validate`, `enable`, `disable`, or `apply_refresh` actions remain.
- Complex inspection/troubleshooting stays in MCP manual + bash/file workflows.
- The runtime loader, MCP transport, registry catalog rendering, LICC inbox, and curated MCP servers are intentionally left in place for later split decisions.

PR #1624, opened 2026-09-04
feat(puffo): admit fixed core MCP in v1

TZZheng:
## Summary
- add `puffo-v1`, reusing the operator-provisioned runtime id while admitting only Puffo Core stdio MCP
- require a successful Driver-authority handoff before v1 starts
- enforce the fixed service name/module and required local-service token while preserving opaque, evolving Puffo environment values; preserve v0 empty-only behavior

## Verification
- `python -m pytest -q -x tests/test_acp_stdio.py tests/test_puffo_v0_profile.py tests/test_architecture_documents.py` (106 passed)
- `python scripts/check_docs_governance.py --check`
- `git diff --check`

`ruff` is not installed in the repository virtualenv.


hellohanchen (commented):
## Pass 1/10 — `v0 / v1 / generic` MCP projection matrix + Driver Authority

Scope: **`35cf91f8c6f36eca565cba69a5bce119f2987d2e`** (base `main`, MERGEABLE), reviewed jointly with puffo-agent #319 @ `e746bb4842f1bfe8ddf9a31e1c8c2b7afbde7d0f`. Both heads re-resolved from the API at the start and again immediately before submitting this review.

Review only — no branch changes, no merge.

**Verdict for this pass: no merge blocker on this PR.** Two non-blocking findings (one here, one filed cross-repo on #319) and several negative results worth recording.

---

### What I verified, and how

Every claim below is an executed experiment against this exact SHA, not a reading of the source. Probe script drove `AcpStdioServer._new_session` and `handle_acp_command` directly.

| # | Experiment | Result |
|---|---|---|
| P3a | `puffo-v0` + a Puffo MCP server in `session/new` | rejected — `mcpServers must be an empty array for this profile` |
| P3a′ | `puffo-v0` + `[]` | accepted, **`mounted: None`** — strictly empty ✅ |
| P3b | `puffo-v1` + the canonical Puffo server | accepted, exactly one `StdioMCPServerConfig(name='puffo', args=('-m','puffo_agent.mcp.puffo_core_server'))` ✅ |
| P3c | `generic` + `{name:'anything', command:'/bin/echo', args:('hi',)}` | accepted and mounted verbatim — unconstrained ✅ |
| P2b | `puffo-v1` with `mcpServers` **absent** | rejected — `mcpServers must be an array` (clean, not a crash) ✅ |

The three positive legs of the matrix hold exactly as described in the PR body.

### Guard counterfactuals (mutation)

Baseline `131 passed`, tree verified clean between every run, each mutant applied alone with its occurrence count asserted `== 1` so the edit provably landed. Verdicts read from the **set of failing test names**, never the exit code.

| Mutant | Verdict | Killing test |
|---|---|---|
| G1 delete the `puffo-v1` CLI Driver-authority requirement | **KILLED** | `test_puffo_v1_cli_requires_an_authenticated_driver_authority` |
| G2 `allow_session_mcp` takes precedence over the validator | **KILLED** | `test_profile_session_rejects_remote_workspace_and_mcp_inputs` |
| G3 `puffo-v0` no longer rejects a non-empty `mcpServers` | **KILLED** | `test_profile_session_rejects_remote_workspace_and_mcp_inputs` |
| G4 v1 validator drops the exactly-one check | **KILLED** | `test_puffo_v1_session_rejects_any_other_mcp_shape[servers1…]` |
| G5 v1 validator drops the fixed-**name** check | **KILLED** | `[servers2…]` |
| G6 v1 validator drops the fixed-**args** check | **KILLED** | `[servers3…]` |
| G7 v1 validator drops the required-token check | **KILLED** | `[servers4/5/6…]` |

7/7 killed, each by a test that names the guard it pins rather than killing it incidentally. This is a genuinely well-pinned set of guards — better than most of what I review.

---

### N1 (non-blocking, Medium) — `validate_puffo_v1_mcp_servers` pins `name` and `args` but never `command`

`src/lingtai/adapters/acp/puffo_v1.py` constrains four things: `len(configs) != 1`, `config.name != "puffo"`, `config.args != ("-m", "puffo_agent.mcp.puffo_core_server")`, and a non-empty `PUFFO_LOCAL_SERVICE_TOKEN`. `command` is unconstrained — it falls through to `_stdio_mcp_configs`' only requirement, that it be a non-empty absolute path. There is no existence check and no allowlist.

**Executed (P2):**

```
P2 ADMITTED arbitrary executable ->
  StdioMCPServerConfig(name='puffo',
                       command='/tmp/definitely-not-python',
                       args=('-m','puffo_agent.mcp.puffo_core_server'),
                       env=(('PUFFO_LOCAL_SERVICE_TOKEN','t'),))
```

The pinned `-m puffo_agent.mcp.puffo_core_server` is only meaningful if the executable is a Python interpreter. Against any other binary it is two opaque argv strings, and the module pin buys nothing.

This matters because of what the module's own docstring claims: that *the fixed name and module arguments prevent this profile from becoming arbitrary MCP ingress*. As written, they do not — a session client that can name a path can run it, with the Puffo service token in its environment. The name/args pins stop a *misconfigured* client, not a chosen one.

Non-blocking because the ACP session client here is the Puffo agent itself, so the practical exposure depends on that trust boundary rather than on this file. But the docstring overstates what the validator enforces, and G5/G6 kill their mutants while nothing at all pins `command` — so the gap is invisible to the suite. Pinning `command` to the resolved interpreter (or an allowlist) would make the stated property true.

### N2 (negative result, recorded) — no CI on this head

`statusCheckRollup` is `null` at `35cf91f8` — absent, not pending. There is no automated evidence for this SHA; everything above is local execution in a clean venv against this exact commit.

### Negative results (checked, nothing found)

- The `generic` profile is **not** subject to the Driver Authority requirement — confirmed by construction, and correctly so.
- Absent / malformed `mcpServers` produces a typed `_RpcError`, not an unhandled exception.
- The `_session_mcp_validator` → `_allow_session_mcp` → generic precedence chain in `server.py::_new_session` is ordered correctly, and G2 proves the ordering itself is pinned.

---

**Cumulative passes: 1/10.** The pass's other finding is filed on puffo-agent#319 (a `puffo-v0` Driver-authority claim in a docstring that this PR's actual behaviour contradicts); it is a defect there, not here.

hellohanchen (commented):
## Pass 2/10 — ACP → LingTai field fidelity and env preservation

Scope: **`35cf91f8c6f36eca565cba69a5bce119f2987d2e`**, reviewed jointly with puffo-ai/puffo-agent#319 @ `e746bb4842f1bfe8ddf9a31e1c8c2b7afbde7d0f`. Both heads re-resolved at the start of the pass and again immediately before submitting; unchanged.

Review only — no branch changes, no merge.

**Verdict for this pass: no merge blocker.** Fidelity through `_stdio_mcp_configs` and `validate_puffo_v1_mcp_servers` is byte-exact on every sample I constructed. Three findings, all non-blocking; two of them concern pre-existing lines that this PR newly makes load-bearing for `puffo-v1`.

---

### Fidelity, executed

Real round trips from Puffo's `McpServerSpec` through the ACP serializer, a genuine `json.dumps`/`json.loads` hop, and into your validators on this head.

| Property | Result |
|---|---|
| `name` / `command` / `args` identical | ✅ |
| env order preserved | ✅ |
| env byte-exact for a hostile sample | ✅ |
| unknown env names preserved string→string, nothing dropped | ✅ |
| `args` order and duplicates preserved | ✅ |
| no secret value logged | ✅ (canary token, 0 bytes of log output) |
| required token: absent → early reject | ✅ |
| required token: empty string → early reject | ✅ |
| `generic` passthrough, `v0` strictly empty | ✅ |

Hostile sample: Unicode values and names, leading/trailing-space names, `=` and newline in names, `SPECIAL!@#$%^&*()`, values containing `"` `'` `\` backtick `$` `;` `|` tab and embedded newlines, an emoji, an empty-string value. `args` sample: `("-v", "-v", "--flag=1", "--flag=1", "", " ", "中文", "-v")`. Everything survives byte-exact, in order.

Your strict field check is also safe against the ACP serializer, which I verified rather than assumed: the client serializes `by_alias=True, exclude_none=True, exclude_defaults=True`, so the wire object is exactly `{args, command, env, name}` — no `type`, no `_meta`. Note that `exclude_defaults` would have dropped an empty `args`/`env` (which `_stdio_mcp_configs` rejects outright); it does not, only because both fields are **required** in `McpServerStdio` and so have no default to match. That compatibility rests on an upstream schema decision, not on either side's code.

### Mutation

Baseline `102 passed` (`tests/test_puffo_v0_profile.py`, `tests/test_acp_stdio.py`). Each mutant alone, occurrence count asserted `== 1`, verdicts from the set of failing test names, tree restored and verified clean between runs.

| Mutant | Verdict |
|---|---|
| Q7 v1 accepts an empty required token (`not token_value` → `token_value is None`) | KILLED — `test_puffo_v1_session_rejects_any_other_mcp_shape[servers6…]` |
| **Q4 env pairs reordered by name** (`tuple(env_pairs)` → `tuple(sorted(env_pairs))`) | **SURVIVED** |
| **Q5 env values whitespace-stripped** (`(key, val)` → `(key, val.strip())`) | **SURVIVED** |
| **Q6 the duplicate-env-name check deleted** | **SURVIVED** |

A method note, because it changed three of these verdicts: my first run named a test file that does not exist in this repo, and pytest exited **4** — a usage error, not a test failure. Every mutant "survived" against a suite that never ran. The numbers above are from the re-run against `tests/test_acp_stdio.py` with a confirmed `102 passed` baseline. I mention it because the failure mode is invisible if you read the exit code instead of the summary line.

### N1 (Low, non-blocking) — three env-fidelity properties that `puffo-v1` now depends on are unpinned

Q4, Q5 and Q6 are all lines in `_stdio_mcp_configs`, which predates this PR — I checked the diff rather than assuming. But `validate_puffo_v1_mcp_servers` calls `_stdio_mcp_configs` and then adds its pins on top, so as of this PR the `puffo-v1` contract's env guarantees — order preserved, values untouched, names unique — rest entirely on three lines that no test observes:

- **Q4:** the projected env can be silently reordered. Order is semantically irrelevant to a POSIX environment, so this is genuinely Low, but it is also one of the properties the profile is supposed to guarantee.
- **Q5:** env **values** can be silently trimmed. This one has teeth: a credential or a path with significant leading/trailing whitespace would be corrupted in a way no test notices and no log would show (correctly, since values are not logged).
- **Q6:** the duplicate-env-name rejection can be deleted outright. With it gone, `env_pairs` carries both entries and `dict(config.env)` at the spawn site keeps the last — a last-wins override of any name, including the required token, with nothing rejecting it.

The `puffo-v1` shape tests already parametrise over rejected shapes; extending them with a duplicate-env-name case and an order/whitespace assertion would close all three.

### N2 (Low, non-blocking) — env **names** are validated as "non-empty string" only, and one class fails after admission

Neither side constrains env names beyond non-empty-string. `session_mcp.py:112` passes `env=dict(config.env)` straight to `MCPClient`, which hands it to `StdioServerParameters`, so admitted names reach the child process verbatim. Executed against a real subprocess:

```
  equals    ("A=B")   -> ValueError: illegal environment variable name
  newline   ("A\nB")  -> spawned; name reaches the child verbatim
  leading   (" A")    -> spawned; verbatim
  trailing  ("A ")    -> spawned; verbatim
  unicode   ("中文")   -> spawned; verbatim
  special   ("A!@#$%")-> spawned; verbatim
```

The pass-through cases are arguably correct — "preserve all other names and values" is what the docstring promises. The `=` case is the one worth changing: CPython raises `ValueError: illegal environment variable name` at spawn, which is *after* `session/new` has validated and accepted the request. `mount_session_mcp_stdio` closes the clients it opened and re-raises, so this is a late, badly-attributed failure rather than a leak, but rejecting a name containing `=` alongside the existing non-empty check would move it to admission time where the other shape errors live, and would cost one condition.

### N3 (Low, non-blocking, forward-looking) — the v1 validator is wired only into `_new_session`

`session_mcp_validator` is consulted at `server.py:562` inside `_new_session` and nowhere else. Today that is complete, because this adapter has no `session/load`: `initialize` advertises `"agentCapabilities": {}` and the dispatch handles only `initialize`, `session/new`, `session/prompt`, `session/cancel`. Puffo's side (#319) reads that capability and raises `ACP agent does not support session/load` when it is absent, so the resume leg is unreachable for all three profiles — I confirmed this rather than assuming, since covering `session/load` was part of this pass's brief.

The note is for whenever `session/load` is added: it takes `mcpServers` in the ACP protocol, and unless it routes through the same three-way precedence chain, `puffo-v1` will have a fully constrained `session/new` and an unconstrained `session/load`. Cheap to prevent now by factoring the precedence into one helper; expensive to notice later.

### N4 (informational) — the child's environment is replaced, not merged

`env=dict(config.env)` gives the MCP subprocess **only** the projected variables — no `PATH`, no `HOME`, no locale. That is a defensible, arguably good default, and it works today because `command` is required to be an absolute path so no `PATH` lookup is needed. Recording it because it is load-bearing for the absolute-path requirement rather than independent of it, and because it interacts with the `command` pinning gap I filed in pass 1: an admitted arbitrary executable also runs with an attacker-chosen environment and nothing else.

### Negative results (checked, nothing found)

- No secret value logged anywhere on this path; a canary token produced zero log output through validation.
- The required-token check is genuinely early — it runs in `_new_session` before any session state is created — and rejects both absent and empty-string tokens (Q7 confirms it is pinned). A whitespace-only token (`" "`) is accepted; flagging only so the intended definition of "non-empty" is a deliberate choice rather than an accident.
- `_meta` handling is correct on both sides: you accept it and ignore it, and the ACP serializer never emits it unset.
- The duplicate **server name** check (as opposed to env name) is present and correct; only the env-name one is unpinned.

---

**Cumulative passes: 2/10.**

hellohanchen (commented):
## Pass 4/10 — `acp` subparser argv semantics vs Puffo's profile classifier

Scope: **`35cf91f8c6f36eca565cba69a5bce119f2987d2e`**, paired with puffo-ai/puffo-agent#319 @ `e746bb4842f1bfe8ddf9a31e1c8c2b7afbde7d0f`. Both heads re-resolved from the API at the start of the pass and again immediately before submitting; unchanged.

Review only — no branch changes, no merge.

**Verdict for this pass: no merge blocker.** One finding, and it is cheaper to fix on this side than on Puffo's.

---

### N1 (Medium, non-blocking) — `allow_abbrev` makes `--profile` reachable by spellings Puffo cannot classify

I captured the genuine parser your `cli.main()` builds (by intercepting `parse_args`) rather than reimplementing it, and ran 31 argv samples through both it and Puffo's `_lingtai_constrained_profile`.

```
captured real parser: lingtai-agent | allow_abbrev: True
```

`add_acp_parser` takes the argparse default, and `--profile` is the only long option on the `acp` subparser beginning with `p`, so every unambiguous prefix resolves to it:

```
argv                    this parser's effective profile    Puffo's classifier
--prof puffo-v0         puffo-v0                           <generic>
--pro=puffo-v1          puffo-v1                           <generic>
--p puffo-v0            puffo-v0                           <generic>
--profil puffo-v1       puffo-v1                           <generic>
```

Puffo matches only the exact `--profile X` / `--profile=X` spellings. When it reads generic, three things it owes a constrained profile do not happen: the Driver authority FD is not injected, the "constrained LingTai ACP requires the POSIX local spawn path" refusal does not fire, and the empty-MCP projection for `puffo-v0` is not applied.

The dominant outcome is fail-closed on your side, which is why this is not a blocker: `puffo-v0` rejects the non-empty `mcpServers` it then receives, and the `puffo-v1` CLI exits without an authenticated Driver authority. But there is a corner that comes up: Puffo builds its MCP server only when its Puffo-Core environment is populated, and with that empty the projection is `()` for every profile — so `--prof puffo-v0` is accepted here, starts as a genuinely constrained profile, and was prepared by Puffo as a generic one. Pass 1 established that `puffo-v0` starts without a Driver handshake and fails closed only at first use, so that launch does come up.

It also breaks the last-occurrence rule both sides intend to share: `--profile puffo-v1 --prof puffo-v0` resolves to `puffo-v0` here and `puffo-v1` there.

**Suggested fix, here rather than there.** `allow_abbrev=False` on the `acp` subparser is one argument and removes the entire class at the wire. The alternative is prefix-aware matching in Puffo's classifier, which is strictly worse: it would have to track this parser's full option table to know which prefixes are unambiguous, and that coupling silently rots whenever an option is added here. Narrowing the accepted spellings keeps the contract on the side that defines it.

For a profile whose whole purpose is to be constrained, refusing abbreviations also seems like the behaviour you want independently of Puffo.

### Negative results (checked, nothing found)

All of the following were run through the real parser and compared against Puffo's classifier; every one agreed.

- Both exact spellings equivalent; repeated and mixed-spelling occurrences resolve to the last effective value.
- Rejected cleanly, with Puffo agreeing on "not constrained": missing value, `--profile=` empty, unknown value `puffo-v2`, look-alike `--profilex`, value starting with `-`, a value that swallows the next flag, leading-whitespace value, and a value carrying a zero-width Unicode character.
- `--runtime-id` / `--agent-dir`: both together rejected, neither rejected, `--agent-dir` + `--profile` accepted and agreed, repeated `--runtime-id` agreed, `--profile` before `--runtime-id` agreed. No ordering or repetition combination produced a disagreement about the effective profile.
- Over-constraining cases where Puffo says `puffo-v0` and this parser rejects (bare `--` before `--profile`, `acp` twice, extra positional, unknown option, both/neither source): all are the safe direction — Puffo prepares a constrained launch that then dies at argument parsing.
- **Every rejection is a `SystemExit` raised by `parse_args` itself**, so it precedes `handle_acp_command` entirely — `resolve_runtime`, `authority_adapter_from_environment()` and `run_acp` are all downstream. No session, service, or subprocess starts on any rejected argv. That was an explicit thing I set out to falsify and could not.
- `'acp'` appearing as a flag value (`--runtime-id acp --profile puffo-v0`) does not confuse either side.

---

**Cumulative passes: 4/10.**

hellohanchen (commented):
## Pass 5/10 — Driver Authority FD / handshake / ownership lifecycle

Scope: **`35cf91f8c6f36eca565cba69a5bce119f2987d2e`**, paired with puffo-ai/puffo-agent#319 @ `e746bb4842f1bfe8ddf9a31e1c8c2b7afbde7d0f`. Both heads re-resolved from the API at the start of the pass and again immediately before submitting; unchanged.

Risk surface for this pass only: env-locator consumption, handshake typing, claim-once, role checks, and teardown. Not the profile/MCP projection matrix (pass 1), field fidelity (pass 2), sealed-plan aliasing (pass 3), or argv classification (pass 4).

Review only — no branch changes, no merge.

**Verdict for this pass: no merge blocker.** The consuming side fails closed on every hostile input I could construct, and the properties that matter most are mutation-pinned. Two guards on this side are load-bearing and unobserved; the first has a nastier failure mode than its own docstring implies.

---

### N1 (Medium, non-blocking) — the one-time `pop` is unpinned, and a stale locator closes an unrelated descriptor

`adapters/acp/driver_authority.py:533` is `os.environ.pop(DRIVER_AUTHORITY_FD_ENV, None)`, and the function's docstring makes this its central promise: *"It is removed before Agent construction so a child process cannot rediscover authority from inherited configuration."*

Changing that `pop` to `get` leaves the suite fully green (102 passed). No test observes the removal.

What a surviving locator actually costs is not what the docstring suggests. The descriptor is **not** inherited by descendants — both ends of the pair are `set_inheritable(False)` and only Puffo's `pass_fds` carries the endpoint into the direct child — so a grandchild that re-reads a stale locator finds a number naming something else entirely. `from_inherited_fd` correctly fails closed on it, but its rejected path deliberately closes the fd (`:275-294`). Measured:

```
adapter: UnavailableDriverAuthorityAdapter        (typed fail-closed, correct)
BUT the unrelated descriptor 4 was closed by the rejected path: True
```

So the hazard of a leaked locator is not authority rediscovery — that direction is genuinely safe — it is that any descendant calling `authority_adapter_from_environment()` closes whatever descriptor now holds that number, in a process that never had anything to do with the authority. The `os.close(fd)` is right for a real inherited locator; the `pop` is the only thing keeping it from firing on a stale one. A guard with that failure mode should not be invisible to the suite.

Suggested pin: assert `DRIVER_AUTHORITY_FD_ENV not in os.environ` after a successful call **and** after each rejection class.

### N2 (Low, non-blocking) — the root-role requirement is unpinned

A derived-role endpoint offered to the ACP profile is correctly refused — I verified this with a real derived endpoint, which produced `UnavailableDriverAuthorityAdapter`. But deleting the three lines at `:549-552`

```python
if authority.identity.role != "root":
    authority.close()
    raise DriverAuthorityTransportError("ACP profile requires a root authority endpoint")
```

leaves the suite green, after which the profile accepts a derived endpoint as its authority. The negative result is real; it is guarded by nothing but those lines.

### N3 (Low, non-blocking) — cleanup paths on the Puffo side are unpinned

Filed in full on #319: the parent's `endpoint.close()` and the spawn-failure `authority.close()` both survive mutation, though the shipped code measurably does the right thing (child exit ends the serve thread; five failed spawns leak 0 fds and 0 threads).

---

### Real experiments

Real `socketpair`s, real subprocesses, real `/proc/<pid>/fd` inspection. Nothing mocked.

| # | Experiment | Result |
|---|---|---|
| P1 | root handshake end to end | `DriverAuthorityClient`, `role=root`, `launch_id` matches; env popped ✅ |
| P2 | second hello on a **dup** of the same FD | `RejectedDriverAuthorityAdapter`, reason `endpoint_already_claimed` ✅ |
| P4 | 13 hostile FD locator classes | all typed fail-closed, all popped, zero tracebacks ✅ |
| P5 | derived-role endpoint | refused ✅ (but see N2) |
| P6 | peer closes immediately / half frame during hello | typed, no hang ✅ |
| P7 | `close()` racing a claim paused inside the state lock | completed cleanly ✅ |

**P4 in detail** — absent, empty string, non-numeric, negative, `3.5`, `True`, a 20-digit integer, an out-of-range fd number, a closed fd, a regular file, a live pipe (non-socket), whitespace-padded, and Arabic-Indic digits. All thirteen returned `UnavailableDriverAuthorityAdapter`, all thirteen popped the variable, none raised. Notably none of them falls back to a generic adapter — the `except (TypeError, ValueError, OverflowError, DriverAuthorityTransportError)` arm and the typed wrapping inside `from_inherited_fd` cover the whole space I could reach.

**A correction on my own method.** My first concurrency probe tried to race two hellos against one endpoint and reported "no single winner". That was a flaw in the probe, not a finding: one root endpoint has one serve loop, so two hellos on dups of the same FD are serialised by construction — which is what P2 measures. The genuine interleaving case is covered by this repo's own `test_endpoint_claim_transition_cannot_be_interleaved` via the `claim_probe` hook, and T5 below confirms that test is load-bearing. I re-ran the race as P7 instead.

### Mutation

Baseline `102 passed`. Each mutant alone, occurrence count asserted `== 1`, verdicts from the set of failing test names, tree restored and verified clean between runs.

| Mutant | Verdict |
|---|---|
| T3 `puffo-v1` accepts an unauthenticated authority | KILLED — `test_puffo_v1_cli_requires_an_authenticated_driver_authority` |
| T5 the endpoint can be claimed more than once | KILLED — `test_endpoint_claim_is_single_use_and_reuse_is_audited`, `test_endpoint_claim_transition_cannot_be_interleaved` |
| **T1 read the FD env without consuming it** | **SURVIVED** → N1 |
| **T2 accept a non-root (derived) endpoint** | **SURVIVED** → N2 |

The Puffo-side mutants (caller-FD stripping, `pass_fds` exactness, factory refusal, teardown, inheritability) are reported on #319.

### Negative results (checked, nothing found)

- `puffo-v1` exits **before** Agent construction without a real `DriverAuthorityClient` (`cli_acp.py:251-256`), and T3 pins it.
- `puffo-v0` with a missing or broken authority yields only `UnavailableDriverAuthorityAdapter` / `RejectedDriverAuthorityAdapter`; the `else authority` fallback at `:261` keeps that same non-client object as the derived-launch port, so derived launches stay indeterminate/denied. No path converts an unauthenticated adapter into a permissive one, and there is no generic fallback for a constrained profile.
- Rejection results preserve their `state`, `reason_code`, and `audit_id` through `DriverAuthorityHelloRejected` into `RejectedDriverAuthorityAdapter` — the reason/source/audit semantics survive the boundary.
- No malformed locator class reaches a traceback or an untyped exception (P4).
- Half-frame and abrupt disconnect during hello are typed with no hang (P6).
- Claim-once holds under a dup of the same FD (P2), and the locked transition is covered by the repo's own interleaving test (T5).

---

**Cumulative passes: 5/10.**

hellohanchen (commented):
## Pass 6/10 — Puffo runtime registry: identity binding, atomicity, revocation, TOCTOU

Scope: **`35cf91f8c6f36eca565cba69a5bce119f2987d2e`**, paired with puffo-ai/puffo-agent#319 @ `e746bb4842f1bfe8ddf9a31e1c8c2b7afbde7d0f`. Both heads re-resolved from the API at the start of the pass and again immediately before submitting; unchanged. Merge-base `8857a0d31fe8eb6f610489c02b229cca37b85177`.

Risk surface for this pass only: `puffo_v0.py` registry — id/schema validation, filesystem hardening, directory binding, resolve/provision/revoke atomicity and linearization, and the resolve-to-start window. Not the authority socket/FD lifecycle (pass 5), argv classification (pass 4), or MCP projection (passes 1–2).

Review only — no branch changes, no merge.

**Verdict: one merge blocker**, found while establishing a mutation baseline rather than by reviewing the registry. Plus four non-blocking findings. The registry's own core properties — linearization, atomic writes, revocation monotonicity, schema strictness — held against every experiment I ran.

---

## B1 — Blocker: a test is red at this head, green at the merge-base, and no CI has run

```
FAILED tests/test_provider_admission.py::test_provider_dispatch_concurrency_inventory_is_explicit
```

- At merge-base `8857a0d3`: **1 passed**.
- At head `35cf91f8`: **1 failed**.
- `statusCheckRollup` at this head is **`null`** — not pending, absent. No automated run has ever evaluated this commit.

**Cause.** The test pins every thread-spawning call site by literal line number. This PR adds six lines to `src/lingtai/adapters/acp/server.py` (the `session_mcp_validator` parameter, its assignment, and the validator branch in `_new_session`), which shifts all three `Thread(` sites in that file:

| expected (`test_provider_admission.py:343-345`) | actual at head |
|---|---|
| `server.py`, 277 | 279 |
| `server.py`, 304 | 306 |
| `server.py`, 726 | 730 |

**Severity, stated precisely.** I checked whether this is a genuinely unaudited concurrency site or a stale fixture: the three sites are the same three constructs, and all three remain in `outside_root_provider_dispatch`. So the fix is a line-number update, not a design question. It is a blocker because the concurrency-audit gate itself is currently failing at the head, not because of a behavioural defect.

Worth saying plainly: this is the second consecutive pass where a defect on this PR was invisible to CI because CI has not run. A red test that no gate evaluates is indistinguishable from a green one.

---

## N1 (Medium, non-blocking) — "replacement at the same path" is not rejected when the inode is recycled

`CONTRACT.md:227-229` states: *"Resolve rejects a symlink retarget, canonical-path drift, **or replacement at the same path**."*

The binding is `(device, inode, owner, group)` from `_directory_binding`. Inode **numbers** are recycled by the filesystem, so deleting a bound directory and recreating one at the same path very often reproduces the same number — and the binding then matches a physically different directory. Five trials, ext4:

```
trial 0: inode 32506148 -> 32506148  reused=True   resolve=OK
trial 1: inode 32506156 -> 32506156  reused=True   resolve=OK
trial 2: inode 32506164 -> 32506164  reused=True   resolve=OK
trial 3: inode 32506172 -> 32506172  reused=True   resolve=OK
trial 4: inode 32506180 -> 32506180  reused=True   resolve=OK
accepted a physically different directory at the same path in 5/5 trials
```

With substituted content, the opaque runtime id resolves to it:

```
same-path substitute inode 32506188 -> 32506188 ; resolve=OK
resolved agent_dir now serves: attacker content | init.json = {"pwned":true}
```

**Why the suite does not see this.** The pinning test is named for exactly this property — `test_runtime_binding_rejects_replaced_directory_at_the_same_canonical_path` — but it replaces via `rename`:

```python
agent_dir.rename(tmp_path / "original-agent")
...
agent_dir.mkdir()
```

Keeping the original directory alive means its inode *cannot* be reused, so the test can only ever exercise the case where the number already differs. It passes for a reason narrower than its name. The rename case is genuinely rejected (I confirmed it as a control), and mutation M2 shows the inode dimension is load-bearing for that case — but the delete-and-recreate case is uncovered and unrejected.

The actor here needs write access to the agent directory, which is a strictly weaker capability than editing the `0600` registry. Either narrow the contract sentence, or bind to something that is not recycled — a provision-time nonce file inside the directory would be the cheap version.

## N2 (Medium, non-blocking) — the symlink-retarget rejection rests on one unpinned line

Symlink retarget at a bound path is correctly rejected today:

```
SHIPPED   symlink retarget at the bound path -> REJECT  'agent_dir binding no longer matches its canonical path'
```

M3 removes the two-line canonical-path equality check in `_bound_directory` (`if str(resolved) != path_value:`) with the full suite green, and the same scenario is then **accepted**, resolving to a different directory than the registered one:

```
M3 mutant symlink retarget -> OK  agent_dir=PosixPath('.../real')   # registered path was .../a
```

This is the first clause of the contract sentence quoted in N1. Notably it is the canonical-path check doing the work, not the `lstat`: M4 (swapping `lstat` for `stat` in `_directory_binding`) leaves the same scenario rejected, so those two guards are redundant with each other and only one of them is decisive. Pin the decisive one.

## N3 (Low, non-blocking) — a crash between the tombstone log and the registry write wedges provisioning permanently

`provision_runtime` calls `_initialize_revocation_log` (which creates the log `O_EXCL`) and then `_write_registry`. A crash in between leaves the log present and the registry absent. Every later provision — for **any** runtime id — then fails at the `O_EXCL` create:

```
child rc: -9 (SIGKILL between the two writes)
registry: False  tombstone: True
any later provision -> puffo-v0 registry initialization found an unexpected revocation log
```

The state is correctly *not* treated as a valid registry, which is what the contract requires. The gap is recovery: the error names no remedy, and the file to delete is a dotfile the operator has never been told exists. `_write_registry` also does not fsync the parent directory after `os.replace`, so the same shape is reachable from a power loss rather than a signal. Both failure directions are fail-closed; this is an operability finding, not a security one.

## N4 (Low, non-blocking) — the registry parent is never checked for being a symlink, and only the leaf is hardened

`_secure_registry_directory` uses `mkdir(parents=True, mode=0o700, exist_ok=True)` followed by `chmod`. Both follow symlinks, and there is no `lstat`/`O_NOFOLLOW` check on the directory — unlike the registry, lock, and tombstone files, which reject symlinks, FIFOs, and directories (I confirmed all three). A link planted at the registry directory before first use survives hardening and silently redirects the whole control plane:

```
provision -> OK
registry physically landed in the attacker dir: True
planted symlink survived hardening: True
attacker dir mode after chmod-through-link: 0o700
```

The `mode=0o700` argument applies only to the final component. Intermediate directories created by `parents=True` take the process umask, so the `~/.lingtai` level came out **`0o775`** on this host (umask `002`) while the leaf was `0o700`. `_secure_registry_directory`'s docstring — *"Create and harden the registry parent independently of umask"* — is accurate for the leaf only.

**Honest scoping:** confidentiality still holds, because the leaf is `0700` and the files are `0600`. Planting the link requires local write access under `~/.lingtai`. That is a weaker capability than "able to edit the registry", which `CONTRACT.md:274` explicitly places outside the threat model, so this sits just inside the stated boundary rather than clearly outside it. An `lstat` on the parent is a two-line fix; hardening intermediates is the other half.

## N5 (informational) — registry and tombstone reads are unbounded

`_read_registry` and `_read_revoked_runtime_ids` both `read_text()` the whole file with no size cap, and requirement 1 asks for bounded behaviour on large input. Measured: a 4.9 MB tombstone log (200k entries) still resolves in 0.30 s. The only supported writer is the local control plane, so I am recording this rather than filing it.

---

## Real experiments

Real temporary filesystems, real `rename`/`unlink`/`symlink` races, real threads, a real `SIGKILL`ed child. Nothing mocked except the two deliberate crash injections.

**Identity and schema (requirement 1).** `_valid_runtime_id` accepts `a`, `A9`, `a-b_c1`, and a 128-char id; rejects empty, 129 chars, leading `-`/`_`, `a/b`, `../etc`, `a b`, `a.b`, `a\n`, NUL, full-width `ＡＢ`, and the non-`str` types `int`/`None`/`bool`/`bytes`/`dict`. Registry-level: extra field ❌, stale digest ❌, truncated JSON ❌, `version` as bool ❌, and a binding member as `True` ❌ — the explicit `isinstance(item, bool)` guard in `_parse_binding` is doing real work here, since `True == 1` would otherwise pass an `int` check.

**Filesystem hardening (requirement 2).** Leaf directory `0o700`; registry, tombstone log, and lock all `0o600`. Registry as a symlink ❌, as a FIFO ❌, as a directory ❌; tombstone log as a FIFO ❌. See N4 for the directory gap.

**Directory binding (requirement 3).** Rename-swap at the same path ❌ *"binding no longer matches its provisioned identity"*; symlink at the bound path ❌ *"…its canonical path"*; regular file at the bound path ❌; `init.json` removed ❌. Delete-and-recreate — see N1.

**The two-resolve window (requirement 4).** A revoke landing between `handle_acp_command`'s resolve and `run_acp`'s re-resolve is refused at the second resolve, before `build_agent`:

```
second resolve -> REJECT | runtime registry entry is inactive or does not match puffo-v0
```

so the equality compare is never even reached. After the second resolve the window is genuinely open — `PuffoV0Runtime.agent_dir` is a `Path`, i.e. a name, not a pinned descriptor, and a swap at that moment redirects it. There is no `openat`/fd pinning anywhere in this module. That matches the contract's explicit same-OS-principal exclusion, and I am recording it as the boundary rather than as a finding.

**Atomicity and durability (requirement 5).** A resolve spinning while 39 provisions ran produced **zero** spurious failures — `os.replace` gives readers whole-file snapshots. Repeated failed provisions leave no stray `.tmp` files. See N3 for the crash midpoint.

**Linearization (requirement 6).**

```
8x concurrent provision, same id      -> {'OK': 1, 'REJECT': 7}; entries: ['same']
12x concurrent provision, distinct ids -> ok=12 entries=12  lost-update=False
5336 resolves spanning a revoke        -> 993 OK then 4343 REJECT; monotonic after first reject = True
```

No lost update, no mixed-field read, and revocation is monotonic from the reader's point of view.

**Revocation terminality (requirement 7).** Resolve after revoke ❌; reprovision of a revoked id ❌ *"runtime_id is revoked and cannot be provisioned again"*; double revoke is idempotent and writes one tombstone; hand-flipping `status` back to `active` **with a correctly recomputed digest** is still ❌ because the tombstone is independent; deleting the tombstone log is ❌ rather than treated as empty history.

**On `entry_digest`.** It is an unkeyed SHA-256, so a writer who can edit the registry can forge a consistent entry — I confirmed a forged workspace redirect with a recomputed digest resolves OK. `CONTRACT.md:233` already scopes the digest to the entry rather than to tamper-resistance, and `:274` excludes that principal, so this is a confirmation of the stated model, not a finding.

## Mutation

Baseline **1 failed, 302 passed** across `test_puffo_v0_profile.py`, `test_cli.py`, `test_system_runtime_policy.py`, `test_provider_admission.py`, `test_acp_stdio.py`, `test_correlated_turns.py`. (That one baseline failure is B1; every verdict below is taken from *new* failing test names relative to it, so it does not launder any result.) Each mutant applied alone, occurrence count asserted `== 1`, tree restored and `git status --porcelain` verified empty between runs.

| Mutant | Verdict | Consequence |
|---|---|---|
| M1 drop the `entry_digest` check | KILLED — `test_registry_rejects_tampering_and_revoked_runtime` | |
| M2 drop `inode` from the binding | KILLED — `test_runtime_binding_rejects_replaced_directory_at_the_same_canonical_path` | but see N1 |
| M7 non-atomic registry write | KILLED — `test_registry_directory_and_files_are_owner_only_even_with_a_permissive_umask` | |
| M9 revoke persists no tombstone | KILLED — `test_revocation_tombstone_denies_a_stale_active_registry_snapshot` | |
| M10 revoke does not clear `status` | KILLED — `test_registry_mutations_are_linearized_so_revoke_cannot_be_resurrected` | |
| M11 no cross-process mutation lock | KILLED — `test_registry_mutation_lock_blocks_a_second_process`, `…linearized_so_revoke_cannot_be_resurrected` | |
| M15 resolve ignores the revocation log | KILLED — `test_missing_required_tombstone_log_fails_closed`, `…stale_active_registry_snapshot` | |
| **M3 drop the canonical-path equality check** | **SURVIVED** | **consequential** — symlink retarget accepted → N2 |
| M4 `stat` instead of `lstat` in the binding | SURVIVED | not consequential — masked by M3's check |
| M5 resolve only once (no `run_acp` re-resolution) | SURVIVED | requirement-4 guard entirely unpinned |
| M6 drop the second-resolve equality compare | SURVIVED | same |
| M8 skip the registry file-type check | SURVIVED | not consequential — a FIFO registry still rejects, with different wording |
| M12 provision ignores revocation history | SURVIVED | not consequential — masked by the surviving `"runtime_id is already provisioned"` entry |
| M13 resolve accepts a non-active entry | SURVIVED | low — redundant with the tombstone log, which is the durable check |
| M14 no non-POSIX fail-closed | SURVIVED | structurally untestable on Linux without patching `os.name`; recording, not filing |
| M16 drop `owner`/`group` from the binding | SURVIVED | not consequential — `(device, inode)` already identifies the object |

**M5 and M6 deserve a note.** The composition root's re-resolve-and-compare is the entire mechanism for requirement 4, and *both* halves can be deleted with the suite green. My experiment shows the mechanism works; nothing in the suite would notice if it stopped. Given the second resolve is the only thing standing between the CLI's first resolve and `build_agent`, one test asserting a revoke between the two refuses startup would be cheap and would pin both.

I also want to be explicit that six of the ten survivors above are survivors I checked and then argued *against* filing. An unpinned line is only worth reporting once you have shown what breaks without it.

## Negative results (checked, nothing found)

- No id syntax, JSON shape, type-confusion, or bool-as-int case reached an unhandled exception; every rejection was a bounded `PuffoV0RegistryError`.
- No concurrent schedule produced a lost update, a resurrected revoked entry, a mixed-field entry, or a torn read.
- No failure path left a stray temporary file or a stale lock.
- Revocation is terminal against every route I could construct, including a full-registry rollback with a valid recomputed digest.
- Active bindings are one-to-one: a second active runtime cannot claim the same `agent_dir` or `workspace`.
- The second resolve closes the revoke-during-startup window before Agent construction.

---

**Cumulative passes: 6/10.**

